Issue #21, opened 2017-05-31
Updates to feeShare, blockRewardCut, and pricePerSegment in transcoder() should be pending


dob:
We need to differentiate between pending state and active state in transcoder() settings. One way to handle this is that whatever the transcoder sets when they call this function is considered pending. When `initializeRound()` is called, `setCurrentActiveTranscoders()` can copy the pending parameters into the active parameters for the active transcoders inside the loop.

I'm not certain whether the active states should live inside the Transcoder objects themselves, or whether the active states should be copied into some separate ActiveTranscoders structure. I could go either way.

yondonfu:
Closed by #29

Issue #20, opened 2017-05-31
Enhance transcoder() function to include job parameters.


dob:
Enhance the `transcoder()` function to include `_feeShare` and `_pricePerSegment`.

In the future we will want to use gas accounting rather than a flat `_pricePerSegment`, but for now this is a reasonable starting point.

`_feeShare` should be given as a percent. For example, if they pass it in as `5`, then they will share 5% of the fees with their delegators.

`_pricePerSegment` can be given in LPT base units (we need a name for that since 1 LPT == 10^18 base units).

yondonfu:
Regarding a name for LPT base units - is that something that should be defined at the token/protocol contract level? Since the token contract already defines `decimals`, perhaps a name for LPT base units can be something that is defined by the community at a later point in time? Maybe there will end up being different thresholds as well similar to Ether (i.e. wei, finney, etc.)

dob:
Correct, it's only convention and isn't strictly defined within the
protocol. It'll just make it easier for us to pronounce: 10,000 dovens vs
10 * 10^-15 LPT.

yondonfu:
Closed by #29

PR #29, opened 2017-06-01
Transcoder job params + pending transcoder information

yondonfu:
Addresses #20 and #21 

- Introduce variables representing pending values for `blockRewardCut`, `feeShare` and `pricePerSegment`
- Pending transcoder information values are set as current transcoder information values when `setCurrentActiveTranscoders()` is called

### DIFF
--- a/contracts/LivepeerProtocol.sol
+++ b/contracts/LivepeerProtocol.sol
@@ -504,16 +504,19 @@ contract LivepeerProtocol {
             }
             // Copy node
             currentActiveTranscoders[i] = transcoderPools.activeTranscoders.nodes[i];
+
+            address currentActiveTranscoder = currentActiveTranscoders[i].id;
+
             // Set address of node to be present in current active transcoder set
-            isCurrentActiveTranscoder[currentActiveTranscoders[i].id] = true;
+            isCurrentActiveTranscoder[currentActiveTranscoder] = true;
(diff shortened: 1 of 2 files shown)


contracts/LivepeerProtocol.sol
```diff
@@ -498,6 +508,12 @@ contract LivepeerProtocol {
             isCurrentActiveTranscoder[currentActiveTranscoders[i].id] = true;
             // Set index position of node in current active transcoder set
             currentActiveTranscoderPositions[currentActiveTranscoders[i].id] = i;
+            // Set pending blockRewardCut as actual value
+            transcoders[currentActiveTranscoders[i].id].blockRewardCut = transcoders[currentActiveTranscoders[i].id].pendingBlockRewardCut;
+            // Set pending feeShare as actual value
+            transcoders[currentActiveTranscoders[i].id].feeShare = transcoders[currentActiveTranscoders[i].id].pendingFeeShare;
+            // Set pending pricePerSegment as actual value
+            transcoders[currentActiveTranscoders[i].id].pricePerSegment = transcoders[currentActiveTranscoders[i].id].pendingPricePerSegment;
```
dob: There's a lot of reptition in these 3 assignments of indexing into the transcoders array. Should we pull it out into a temporary transcoder variable, and then write it back? Could be more efficient with gas and easier to read.

Issue #22, opened 2017-05-31
Job() request transaction


dob:
The `job()` function is how broadcasters request transcoding jobs on chain. The inputs are:

* streamID
* transcoding options
* max price/segment

`streamID` lets the transcoding node know which stream to request on the network

transcoding option semantics are TBD. Let's assume it's a binary string of fixed length for now, and we can interpret it as 0x1, 0x2, etc, without use. 

The function should consider all transcoders who are currently available and are charging a price/segment less than or equal to the max price/segment provided, and choose one (hopefully randomly). The function should populate the metadata associated with a `Job` and store this so that the contract knows what jobs are open, who they are assigned to, and what they're status is in terms of claiming the work and verification. This likely requires a new `Job` struct. Jobs can also be assigned a unique ID so that they can be retrieved from the contract by ID.

yondonfu:
Some thoughts on considering transcoders that are currently available and charging a price/segment less than or equal to the max price/segment.

In order to avoid the use of additional loops, what if we keep track of an `availableTranscoders` array for a `Job` struct. There would be some timedelta within which current active transcoders can call `transcoderAvailability()` for a particular job which would add the transcoder to `availableTranscoders`, but only if the `transcoders[msg.sender].pricePerSegment <= job.maxPricePerSegment`. After the timedelta, the broadcaster can call `startJob(jobId)` and the function will pseudorandomly select a transcoder from `availableTranscoders` for the job. If a transcoder is unresponsive, a broadcaster can call `startJob(jobId)` again to select a different transcoder from `availableTranscoders`.

dob:
Right now I like that in our protocol a broadcaster only needs one on-chain call in the whole workflow (I think). Adding a second `startJob()` call would double the fees and add another confirmation period before they could start broadcasting. It's also nice that transcoders can just declare their availability once, rather than having to offer availability for each additional job (many of which they won't receive).

Are you trying to avoid looping over the availableTranscoders to compare their prices on every job? I wonder if it's more efficient to randomly select, and then check for availability and price? If we assume that transcoders are largely available and looking for work, then in the worst case this becomes the same as a loop, but in the avg case it avoids doing a loop and finds a hit in some much more efficient frequency.

yondonfu:
Closed by #30

PR #30, opened 2017-06-12
job() function and Job struct

yondonfu:
Addresses #22 

- `job()` function to create a new `Job` struct
- `electCurrentActiveTranscoder(_maxPricePerSegment)` is used to pseudorandomly select a current active transcoder with an acceptable price per segment
- `transcoderOptions` is currently a `bytes32` with encoding semantics to be determined at a later point in time

### DIFF
--- a/contracts/LivepeerProtocol.sol
+++ b/contracts/LivepeerProtocol.sol
@@ -122,7 +122,7 @@ contract LivepeerProtocol {
     mapping (address => mapping (uint256 => uint256[2])) public rewardMultiplierPerTranscoderAndRound;
 
     // The various states a job can be in
-    enum JobStatus { Inactive, Active, Ended, ClaimsProvided, Verified }
+    enum JobStatus { Inactive, Unclaimed, Claimed, Verified }
 
     // Represents a transcoding job
     struct Job {
@@ -550,7 +550,7 @@ contract LivepeerProtocol {
         // Check if there is an elected current active transcoder
         if (electedTranscoder == address(0)) throw;
 
-        jobs.push(Job(jobs.length - 1, _streamId, _transcodingOptions, _maxPricePerSegment, msg.sender, electedTranscoder, JobStatus.Active));
+        jobs.push(Job(jobs.length - 1, _streamId, _transcodingOptions, _maxPricePerSegment, msg.sender, electedTranscoder, JobStatus.Unclaimed));
 
         return true;
     }
--- a/contracts/LivepeerProtocol.sol
+++ b/contracts/LivepeerProtocol.sol
@@ -122,17 +122,17 @@ contract LivepeerProtocol {
     mapping (address => mapping (uint256 => uint256[2])) public rewardMultiplierPerTranscoderAndRound;
 
     // The various states a job can be in
-    enum JobStatus { Inactive, Unclaimed, Claimed, Verified }
+    enum JobStatus { Inactive, Active }
 
     // Represents a transcoding job
     struct Job {
         uint256 jobId;                     // Unique identifer for job
-        uint256 streamId;                  // Unique identifier for stream
+        uint256 streamId;                  // Unique identifier for stream. TODO: change to more semantically proper type when we settle on a streamId representation in the system
         bytes32 transcodingOptions;        // Options used for transcoding
         uint256 maxPricePerSegment;        // Max price (in LPT base units) per segment of a stream
         address broadcasterAddress;        // Address of broadcaster that requestes a transcoding job
         address transcoderAddress;         // Address of transcoder selected for the job
-        JobStatus status;                  // State that job is in
+        uint256 endBlock;                  // Block at which the job is ended and considered inactive
     }
 
     Job[] jobs;
@@ -175,6 +175,7 @@ contract LivepeerProtocol {
 
         // Set block time to 1 second for testing purposes
         blockTime = 1;
+
 
         // Round length of ~1 day assuming ~17 second block time on main net
         // Current value is for testing purposes
@@ -550,19 +551,49 @@ contract LivepeerProtocol {
         // Check if there is an elected current active transcoder
         if (electedTranscoder == address(0)) throw;
 
-        jobs.push(Job(jobs.length - 1, _streamId, _transcodingOptions, _maxPricePerSegment, msg.sender, electedTranscoder, JobStatus.Unclaimed));
+        jobs.push(Job(jobs.length - 1, _streamId, _transcodingOptions, _maxPricePerSegment, msg.sender, electedTranscoder, 0));
 
         return true;
+    }
+
+    /*
+     * Computes the status of a job
+     * @param _jobId Job identifier
+     */
+    function jobStatus(uint256 _jobId) constant returns (JobStatus) {
+        if (jobs[_jobId].endBlock > 0 && jobs[_jobId].endBlock <= block.number) {
+            // A job is inactive if its end block is set and the current block is greater than or equal to the job's end block
+            return JobStatus.Inactive;
+        } else {
+            // A job is active if the current block is less than the job's termination block
+            return JobStatus.Active;
+        }
     }
 
     /*
      * Return a job
      * @param _jobId Job identifier
      */
-    function getJob(uint256 _jobId) constant returns (uint256, uint256, bytes32, uint256, address, address, JobStatus) {
+    function getJob(uint256 _jobId) constant returns (uint256, uint256, bytes32, uint256, address, address, uint256) {
         Job job = jobs[_jobId];
 
-        return (job.jobId, job.streamId, job.transcodingOptions, job.maxPricePerSegment, job.broadcasterAddress, job.transcoderAddress, job.status);
+        return (job.jobId, job.streamId, job.transcodingOptions, job.maxPricePerSegment, job.broadcasterAddress, job.transcoderAddress, job.endBlock);
+    }
+
+    /*
+     * End a job. Can be called by either a broadcaster or transcoder of a job
+     */
+    function endJob(uint256 _jobId) returns (bool) {
+        // Check if job exists
+        if (_jobId < 0 || _jobId >= jobs.length) throw;
+        // Check if job already has an end block
+        if (jobs[_jobId].endBlock > 0) throw;
+        // Check if called by either broadcaster or transcoder
+        if (msg.sender != jobs[_jobId].broadcasterAddress && msg.sender != jobs[_jobId].transcoderAddress) throw;
+
+        // Set set end block for job to be in 10 blocks
+        // Note: this is a placeholder value and subject to change
+        jobs[_jobId].endBlock = block.number + 10;
     }
 
     /*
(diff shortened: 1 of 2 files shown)


contracts/LivepeerProtocol.sol
```diff
@@ -121,6 +121,22 @@ contract LivepeerProtocol {
     // rewardMultiplier[1] -> transcoder's cumulative stake for round
     mapping (address => mapping (uint256 => uint256[2])) public rewardMultiplierPerTranscoderAndRound;
 
+    // The various states a job can be in
+    enum JobStatus { Inactive, Unclaimed, Claimed, Verified }
```
yondonfu: The values for this enum are the states that a job can be in. A job is `Unclaimed` after a broadcaster submits a transcoding job by calling `job()`. A job is `Claimed` after a transcoder calls `endJob()` and submits on-chain the merkle root of the transcode claims for a range of segments. A job is `Verified` after a transcoder successfully calls `verify()` for all segments that are eligible for verification (determined after `endJob()`).

Not sure if we need an `Inactive` state. We might use it if we introduce a function to "close" a job if a transcoder is not being responsive or if a broadcaster does not want to live stream after a period of time during which no transcoder claims work for the job. Or we can just leave jobs that fall under these circumstances as `Unclaimed` .

dob: There's a state that I'm not sure how to handle...

I think a transcoder should be able to call `endJob()` for any range of segments, which means that they may make **many** calls to `endJob()` for a particular job. For example...if they never received segment 9, they may call `endJob(1, 8)` and then `endJob(10, 99)`. 

Perhaps it's possible that a job never truly ends, and it's only the claims that are closed out, based on the time limit from when a claim is submitted until it is verified.

contracts/LivepeerProtocol.sol
```diff
@@ -121,6 +121,22 @@ contract LivepeerProtocol {
...
+
+    // Represents a transcoding job
+    struct Job {
+        uint256 jobId;                     // Unique identifer for job
+        uint256 streamId;                  // Unique identifier for stream
```
dob: Let's remember to change `streamID` from `uint256` to it's more semantically proper type when we settle on how to represent `streamID` in the system. For now the concatenation of a node address + unique id in the same space might not be great since it's so long, but even that long I don't think it's a uint256.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +539,49 @@ contract LivepeerProtocol {
...
+     * Pseudorandomly elect a currently active transcoder that charges a price per segment less than or equal to the max price per segment for a job
+     * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
+     */
+    function electCurrentActiveTranscoder(uint256 _maxPricePerSegment) constant returns (address) {
+        // Compute a random index to select a transcoder
+        uint256 randIdx = uint256(block.blockhash(block.number.sub(1))) % currentActiveTranscoders.length;
+
+        for (uint256 i = 0; i < currentActiveTranscoders.length; i++) {
+            // Find randomly selected transcoder and check if it has an acceptable price per segment
+            if (i == randIdx && transcoders[currentActiveTranscoders[i].id].pricePerSegment <= _maxPricePerSegment) {
```
dob: Won't this only check the one transcoder of index `randIdx`? In this case we don't need to loop over all of the transcoders.

Instead I think the loop should be generating a new `randIdx`, right?

contracts/LivepeerProtocol.sol
```diff
@@ -121,6 +121,22 @@ contract LivepeerProtocol {
     // rewardMultiplier[1] -> transcoder's cumulative stake for round
     mapping (address => mapping (uint256 => uint256[2])) public rewardMultiplierPerTranscoderAndRound;
 
+    // The various states a job can be in
+    enum JobStatus { Inactive, Active }
```
yondonfu: As mentioned in Doug's previous comment, a transcoder might submit transcode claims for multiple ranges of segments. As a result, I don't think we can determine on-chain whether all the work for a job has been claimed without additional information. Furthermore, since multiple submissions of transcode claims will require multiple calls to `verify()`, I also don't think we can determine on-chain whether all the claims for a job have been verified without additional information.

It seems that as of right now we can only accurately say that a job is one of two states: `Inactive` or `Active`. A job would become `Active` when a broadcaster submits a job using `job()`. A job would become `Inactive` after either the broadcaster or transcoder for a job calls `endJob()` and a certain number of blocks has passed.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +540,87 @@ contract LivepeerProtocol {
...
+        // Check if job exists
+        if (_jobId < 0 || _jobId >= jobs.length) throw;
+        // Check if job already has an end block
+        if (jobs[_jobId].endBlock > 0) throw;
+        // Check if called by either broadcaster or transcoder
+        if (msg.sender != jobs[_jobId].broadcasterAddress && msg.sender != jobs[_jobId].transcoderAddress) throw;
+
+        // Set set end block for job to be in 10 blocks
+        // Note: this is a placeholder value and subject to change
+        jobs[_jobId].endBlock = block.number + 10;
```
yondonfu: There could be a scenario in which a broadcaster has streamed all segments to a transcoder and then calls `endJob()`, but the transcoder has not had a chance to submit transcode claims for all the segments yet. The `endBlock` field for a job gives a transcoder a grace period during which it can finish submitting transcode claims.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +540,87 @@ contract LivepeerProtocol {
...
+    function getJob(uint256 _jobId) constant returns (uint256, uint256, bytes32, uint256, address, address, uint256) {
+        Job job = jobs[_jobId];
+
+        return (job.jobId, job.streamId, job.transcodingOptions, job.maxPricePerSegment, job.broadcasterAddress, job.transcoderAddress, job.endBlock);
+    }
+
+    /*
+     * End a job. Can be called by either a broadcaster or transcoder of a job
+     */
+    function endJob(uint256 _jobId) returns (bool) {
```
yondonfu: In one of the issues `endJob()` is described as the function called by transcoders to submit transcode claims for a range of segments. This name does not seem too semantically accurate and I think `endJob()` would make more sense as a name for a function used by a broadcaster or transcoder to set a job as inactive. Perhaps the function for submitting transcode claims for a range of segments can instead be called `claimWorkForJob()`.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +540,87 @@ contract LivepeerProtocol {
...
+        // Check if job exists
+        if (_jobId < 0 || _jobId >= jobs.length) throw;
+        // Check if job already has an end block
+        if (jobs[_jobId].endBlock > 0) throw;
+        // Check if called by either broadcaster or transcoder
+        if (msg.sender != jobs[_jobId].broadcasterAddress && msg.sender != jobs[_jobId].transcoderAddress) throw;
```
yondonfu: A broadcaster would call `endJob()` when it is done streaming all segments, setting the job as inactive after a grace period and preventing a transcoder from trying to submit additional transcode claims.

A transcoder would call `endJob()` if a broadcaster does not stream segments to it. The transcoder can just dismiss the job.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +539,87 @@ contract LivepeerProtocol {
...
+                numAvailableTranscoders++;
+            }
+        }
+
+        if (numAvailableTranscoders == 0) {
+            // There is no currently available transcoder that charges a price per segment less than or equal to the max price per segment for a job
+            return address(0);
+        } else {
+            // Pseudorandomly select an available transcoder that charges an acceptable price per segment
+            return availableTranscoders[uint256(block.blockhash(block.number.sub(1))) % numAvailableTranscoders];
```
yondonfu: We loop through all `currentActiveTranscoders` to find the set of transcoders charging an acceptable price per segment. Then we generate a pseudorandom index to select a transcoder from the set of available transcoders.

I'm wary of pseudorandomly selecting a transcoder and _then_ checking if the transcoder is charging an acceptable price per segment because using modulo to generate a pseudorandom index in a range might bias smaller indices and there could be a case where transcoders with an acceptable price per segment are associated with higher indices leading to long loops. While the best case of pseudorandomly selecting a transcoder immediately that charges an acceptable price per segment is nice, I'm not sure if it's worth dealing with the worst case.

contracts/LivepeerProtocol.sol
```diff
@@ -523,11 +539,87 @@ contract LivepeerProtocol {
...
+            return JobStatus.Active;
+        }
+    }
+
+    /*
+     * Return a job
+     * @param _jobId Job identifier
+     */
+    function getJob(uint256 _jobId) constant returns (uint256, uint256, bytes32, uint256, address, address, uint256) {
+        Job job = jobs[_jobId];
```
dob: should we throw if the jobID doesn't exist? It might be better to fail fast than to return an empty job that needs to be checked.

contracts/LivepeerProtocol.sol
dob: This reminds me...we haven't added any Events yet. But we should define ones for all the common actions that may need to be observed. This way a broadcaster could subscribe to an EndJob event in the client and know that they have to reopen the job for a new transcoder.

contracts/LivepeerProtocol.sol
dob: I know this is just a constant, but let's use a protocol parameter, and set the default to something longer than 10, which is only a couple of minutes. Let's set it to something more lenient for now at least like 1000.

contracts/LivepeerProtocol.sol
dob: This is fine for now. In reality I think we want to select in proportion to stake as well. I have seen techniques for generating a random # and making selection according to stake. Let's use this for now, but track as a separate issue.

contracts/LivepeerProtocol.sol
yondonfu: I don't think we need to throw here because currently when `endJob()` is called, a job becomes inactive, but is not deleted from the `jobs` array. I lean toward keeping it this way because we retain a history of jobs, but curious to hear what you think. As a result, this function would never return an empty job and might also be used to inspect information about an inactive job.

I also realized that `jobs[_jobId]` will throw if `_jobId >= jobs.length` (handled by Solidity) and if `_jobId < 0` (since `_jobId` is a `uint256` any negative values will be rejected), so we don't need an additional check + throw for these cases.

contracts/LivepeerProtocol.sol
```diff
@@ -196,6 +215,9 @@ contract LivepeerProtocol {
         // Fail no more than 1% of the time
         verificationFailureThreshold = 1;
 
+        // A job becomes inactive 100 blocks after endJob() is called
+        jobEndingPeriod = 100;
```
yondonfu: This is a test value rather than an actual default value. Although in reality, the default value would be larger, our tests use a helper function to fast forward a certain number of blocks by repeatedly using TestRPC's `evm_mine` request. A large number of `evm_mine` requests slows down our tests a lot, so I think it makes sense to use smaller values for time related variables just for testing purposes. I can file an issue as a reminder to come back and replace these test values with realistic default values.

Issue #24, opened 2017-05-31
Verify() and proof of transcoding


dob:
The `verify()` transaction is what a transcoder calls when he provides proof that he transcoded a particular segment. The inputs are:

* `streamID` (or jobID)
* `segmentSequenceNumber`
* hash of transcoded output for that segment
* signed transcode claim (which includes signed segment data from original broadcaster)
* merkle proof that this transcoded segment is included in the merkle root from the preceeding `endJob()` call.

The function should then:

1. Verify that this segment for this job id is in fact eligible for verification (can be deterministically known based upon the endJob() call).
2. Verify that the original segment was signed by the broadcaster.
3. Verify that the merkle proof is valid and this segment was claimed in endJob() before the transcoder knew this would be challenged.
4. Invoke the verification process. To do so it will also need a swarm hashes of the location of the original and transcoded segments. This should be able to be determined from the signed transcode claim from the broadcaster.

This is non trivial and requires some on chain merkle proof verification and signature verification. We can track those as separate issues, and they would be good opportunities to contribute back to a common library like OpenZeppelin.

Issue #23, opened 2017-05-31
EndJob() and transcoding claim


dob:
The `endJob()` transaction allows the transcoder to claim which segments they transcoded for a particular stream. The params are:

* `streamID` (or maybe jobID is better)
* `startSequenceNumber`
* `endSequenceNumber` - these two params provide the range of segments they're claiming
* merkle root of the segments they are claiming.

When this segment range is claimed, it should kick off a timer window for verification to be completed. When the block is mined, it should be deterministic which segments will be challenged (if any), based on the new block hash, verification rate, and merkle root (or other appropriate sources of unpredictability). The challenged segments don't necessarily need to be stored in state, as long as they can be calculated deterministically after the fact, though if easier to store them in state then that is ok.

The verification will kick off in a call to `Verify()` to be tracked in a separate issue.

yondonfu:
Closed by #35

PR #35, opened 2017-06-20
claimWork and verify without transcoding verification

yondonfu:
Addresses #23 and #24

- Using pasted in `ECVerify.sol` and `MerkleProof.sol` libraries. The functionality of both these libraries are both being added to OpenZeppelin - once the relevant PRs are merged in we can remove these pasted in libraries and directly use the libraries in OpenZeppelin.
- Pulled transcode job logic into a separate `TranscodeJobs.sol` library. This was needed to deploy the `LivepeerProtocol.sol` contract without exceeding gas limits. Should also be a +1 for code modularity


contracts/LivepeerProtocol.sol
```diff
@@ -665,7 +682,7 @@ contract LivepeerProtocol {
      * for its time window
      * @param _transcoder Address of transcoder
      */
-    function validRewardTimeWindow(address _transcoder) internal returns (bool) {
+    function validRewardTimeWindow(address _transcoder) internal constant returns (bool) {
```
yondonfu: No state changes in this function so should be marked constant

test/LivepeerProtocol.js
```diff
@@ -1192,4 +1197,501 @@ contract('LivepeerProtocol', function(accounts) {
...
+                await instance.verify(0, 2, "0x1", "0x2", "0x3", "0x4", {from: accounts[3]});
+            } catch (err) {
+                threw = true;
+            }
+
+            assert.isOk(threw, "verify did not throw when sender is not the assigned transcoder for the job");
+        });
+
+        it("should fail if the segment is not eligible for verification", async function() {
+
```
yondonfu: This is left blank for now because still figuring out the best way to test segment verification eligibility

test/LivepeerProtocol.js
```diff
@@ -1192,4 +1197,501 @@ contract('LivepeerProtocol', function(accounts) {
...
+            await instance.job(streamId, "0x1", 200, {from: accounts[2]});
+
+            // Segment data hashes
+            const d0 = Buffer.from("80084bf2fba02475726feb2cab2d8215eab14bc6bdd8bfb2c8151257032ecd8b", "hex");
+            const d1 = Buffer.from("b039179a8a4ce2c252aa6f2f25798251c19b75fc1508d9d511a191e0487d64a7", "hex");
+            const d2 = Buffer.from("263ab762270d3b73d3e2cddf9acc893bb6bd41110347e5d5e4bd1d3c128ea90a", "hex");
+            const d3 = Buffer.from("4ce8765e720c576f6f5a34ca380b3de5f0912e6e3cc5355542c363891e54594b", "hex");
+
+            // Segment hashes (streamId, segmentSequenceNumber, dataHash)
+            const s0 = abi.soliditySHA3(["uint256", "uint256", "bytes"], [streamId, 0, d0]);
```
yondonfu: Solidity's built-in `sha3` function accepts an arbitrary of number of arguments and tightly packs them together meaning they are concatenated without padding. The `soliditySHA3` function from `ethereumjs-abi` tightly packs arguments before hashing thereby following the behavior of Solidity's `sha3` in Javascript.

contracts/TranscodeJobs.sol
```diff
@@ -0,0 +1,141 @@
...
+        return true;
+    }
+
+    function claimWork(Jobs storage self, uint256 _jobId, uint256 _startSegmentSequenceNumber, uint256 _endSegmentSequenceNumber, bytes32 _transcodeClaimsRoot, uint256 _verificationPeriod) returns (bool) {
+        // Check if job is active
+        if (jobStatus(self, _jobId) != JobStatus.Active) throw;
+        // Check if sender is assigned transcoder
+        if (self.jobs[_jobId].transcoderAddress != msg.sender) throw;
+        // Check if previous verification period over
+        if (block.number < self.jobs[_jobId].endVerificationBlock) throw;
```
yondonfu: For now we force transcoders to call verify a call to `claimWork()` before they can make a subsequent call to `claimWork()`

contracts/TranscodeJobs.sol
```diff
@@ -0,0 +1,141 @@
...
+        self.jobs[_jobId].endBlock = block.number + _jobEndingPeriod;
+
+        return true;
+    }
+
+    function claimWork(Jobs storage self, uint256 _jobId, uint256 _startSegmentSequenceNumber, uint256 _endSegmentSequenceNumber, bytes32 _transcodeClaimsRoot, uint256 _verificationPeriod) returns (bool) {
+        // Check if job is active
+        if (jobStatus(self, _jobId) != JobStatus.Active) throw;
+        // Check if sender is assigned transcoder
+        if (self.jobs[_jobId].transcoderAddress != msg.sender) throw;
```
yondonfu: Only the assigned transcoder for a job can submit transcode claims for a job. Therefore, the transcode claims used to form the submitted Merkle root do not need to be signed by the transcoder

contracts/TranscodeJobs.sol
```diff
@@ -0,0 +1,152 @@
...
+        return true;
+    }
+
+    /*
+     * Computes whether a segment is eligible for verification based on the last call to claimWork()
+     * @param _jobId Job identifier
+     * @param _segmentSequenceNumber Sequence number of segment in stream
+     * @param _verificationRate Rate at which a particular segment should be verified
+     */
+    function shouldVerifySegment(uint256 _segmentSequenceNumber,
```
yondonfu: This helper function determines if a segment should be verified or not. Rather than access state (i.e. a `Jobs` struct), all dependencies are passed in as parameters in order to make this function easier to test.

test/LivepeerProtocol.js
```diff
@@ -1192,4 +1197,497 @@ contract('LivepeerProtocol', function(accounts) {
...
+            const root = merkleTree.getHexRoot();
+
+            // Account 1 claims work
+            await instance.claimWork(0, 0, 3, root, {from: accounts[1]});
+
+            // Get Merkle proof
+            const proof = merkleTree.getHexProof(tClaim0);
+
+            // Account 1 calls verify
+            await instance.verify(0, 0, utils.bufferToHex(d0), utils.bufferToHex(tD0), utils.bufferToHex(bSig0), proof, {from: accounts[1]});
```
dob: Should check that this does not throw.

test/LivepeerProtocol.js
```diff
@@ -1192,4 +1197,497 @@ contract('LivepeerProtocol', function(accounts) {
...
+            await instance.claimWork(0, 0, 3, root, {from: accounts[1]});
+
+            // Get Merkle proof
+            const proof = merkleTree.getHexProof(tClaim2);
+
+            let threw = false;
+
+            try {
+                // Account 1 calls verify
+                await instance.verify(0, 2, utils.bufferToHex(d2), utils.bufferToHex(tD2), utils.bufferToHex(bSig3), proof, {from: accounts[1]});
```
dob: Fails because of bSig3 when it should be bSig2.

test/TranscodeJobs.js
```diff
@@ -0,0 +1,79 @@
...
+            const block = await web3.eth.getBlock(blockNum);
+            const blockHash = block.hash;
+            const verificationRate = 4;
+
+            // Find an eligible segment
+            let hash = abi.soliditySHA3(["uint256", "bytes", "uint256"], [blockNum, Buffer.from(blockHash.slice(2), "hex"), segNum]).toString("hex");
+            let hashNum = new BigNumber(hash, 16);
+            let res = hashNum.mod(verificationRate);
+
+            while (res != 0) {
```
dob: Let's add a comment about why we're iterating here to find a suitable candidate segment number.

test/TranscodeJobs.js
```diff
@@ -0,0 +1,79 @@
...
+            const block = await web3.eth.getBlock(blockNum);
+            const blockHash = block.hash;
+            const verificationRate = 4;
+
+            // Find an ineligble segment
+            let hash = abi.soliditySHA3(["uint256", "bytes", "uint256"], [blockNum, Buffer.from(blockHash.slice(2), "hex"), segNum]).toString("hex");
+            let hashNum = new BigNumber(hash, 16);
+            let res = hashNum.mod(verificationRate);
+
+            while (res == 0) {
```
dob: Same comment as above about why we're looping here.

yondonfu:
In response to the comment on marking library functions in `TranscodeJobs.sol` by @dob: if an internal library function is called by a contract, the Solidity compiler will actually pull in the internal library function code and inline it into the caller contract function. This is in contrast to when a public library function (public by default) is called by a contract, the Solidity compiler will use `delegatecall` to execute the library function code in the context of the current contract without pulling in additional code.

If many internal library functions are called by a contract, the bytecode size of the contract will increase along with the gas costs of deploying the contract. I tested this out locally and marking all the relevant functions in `TranscodeJobs.sol` as internal led to a failed deployment of the `LivepeerProtocol.sol` contract due to insufficient gas.

PR #36, opened 2017-07-05
Change stream id type

yondonfu:
Changes the streamId type from `uint256` to `string` since the LP node represents a streamId as a string.

Also included some small changes for developing the Go client

### DIFF
--- a/contracts/LivepeerProtocol.sol
+++ b/contracts/LivepeerProtocol.sol
@@ -152,6 +152,9 @@ contract LivepeerProtocol {
         _;
     }
 
+    // Events
+    event NewJob(address indexed transcoder, address indexed broadcaster);
+
     // Initialize protocol
     function LivepeerProtocol(uint64 _n, uint256 _roundLength, uint256 _cyclesPerRound) {
         // Deploy new token contract
@@ -379,12 +382,6 @@ contract LivepeerProtocol {
      * Distribute the token rewards to transcoder and delegates.
      */
     function reward() currentRoundInitialized returns (bool) {
-        // Check if in current round active transcoder set
-        if (!isCurrentActiveTranscoder[msg.sender]) throw;
-
-        // Check if already called for current cycle
-        if (transcoders[msg.sender].rewardRound == currentRound() && transcoders[msg.sender].rewardCycle == cycleNum()) throw;
-
         // Check if in a valid transcoder reward time window
         if (!validRewardTimeWindow(msg.sender)) throw;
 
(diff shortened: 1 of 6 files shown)


contracts/LivepeerProtocol.sol
```diff
@@ -683,9 +680,11 @@ contract LivepeerProtocol {
      * for its time window
      * @param _transcoder Address of transcoder
      */
-    function validRewardTimeWindow(address _transcoder) internal constant returns (bool) {
+    function validRewardTimeWindow(address _transcoder) constant returns (bool) {
```
yondonfu: This function is used by the client to poll for when it can call `reward()` so it is now a public function

contracts/TranscodeJobs.sol
```diff
@@ -63,6 +66,8 @@ library TranscodeJobs {
         self.jobs[self.numJobs].transcoderAddress = _electedTranscoder;
         self.numJobs++;
 
+        NewJob(_electedTranscoder, msg.sender);
```
yondonfu: We can't listen for library events, but as a workaround if the library emits an event that is also defined in the contract that uses the library, you can listen for the event via the contract and you will be notified when the library emits the event

parseArtifacts.js
```diff
@@ -0,0 +1,53 @@
+const fs = require("fs");
+const path = require("path");
+const mkdirp = require("mkdirp");
+
+const parseArtifact = (inFile, outDir, outFile, type = "abi") => {
```
yondonfu: For dev purposes, currently using Truffle to compile contracts and then using this script to parse out the ABI and bytecode for contracts. Contract ABI and bytecode is then used to generate Go bindings for the Go client whenever a contract is updated

contracts/TranscodeJobs.sol
dob: Interesting, so if the event is defined in both the main contract and the imported library, then are they considered to be the same if they have the same signature? Or does subscribing to one not necessarily subscribe you to the other?

yondonfu: Yeah so since an imported library can be considered an implicit base contract, if the library emits an event it will show up in the main contract's event log. ATM, the main contract's ABI will not reflect the events of the main contract's imported libraries so clients will not be able to decode events emitted from libraries. However, if you define an event `A` in the main contract with the same signature as the event `A` defined in the library, then the main contract ABI will reflect that event `A` and clients will be able to decode that event `A`.

Issue #41, opened 2017-07-14
Should Return JobID and StreamID in NewJob event


ericxtang:
The jobID never gets passed back to the broadcaster, and it needs it to be able to call EndJob() in case the transcoder isn't being responsive.  It needs the StreamID to figure out which JobID to choose when monitoring the event.

yondonfu:
Closed by #42

Issue #39, opened 2017-07-13
Return StreamID in JobDetails


ericxtang:
Transcoder needs the broadcast streamID to do the transcoding.

yondonfu:
Closed by #42. StreamID can be retrieved using `jobsManager.getJobStreamId(jobId)`.

Issue #38, opened 2017-07-07
TranscodeOptions in Job() should be 128 bytes long


ericxtang:
This is to account for an array of options.

Related to https://github.com/livepeer/golp/issues/21

yondonfu:
Closed by #42. `transcodeOptions` is now a dynamic string

Issue #34, opened 2017-06-19
Upgrade to solidity 0.4.13


dob:
Update the pragma statements in each of the solidity files to force use of the latest compiler version to get optimizations and bug fixes.

dob:
Latest version is now 0.4.13.

PR #42, opened 2017-07-19
Initial refactor for modularization

yondonfu:
The beginnings of a refactor to modularize protocol smart contracts.  The refactor also allowed for some changes (since deployment gas cost is now less of a concern) to address #41 #39 #38 #34 

The refactor required rewriting the single test for the original `LivepeerProtocol` contract so I broke up the test suite into test suites for `JobsManager`, `BondingManager`, `RoundsManager`, `LivepeerProtocol` and `LivepeerProtocolIntegration` which contains two integration tests: one for a reward flow and another for a job-claim-verify flow. 

As is the tests are incomplete and there might be a few things still missing that were lost during the refactor, but opening this PR now to get some eyes early on so we can decide what we think about the new contract structure.


contracts/ContractRegistry.sol
```diff
@@ -0,0 +1,19 @@
+pragma solidity ^0.4.11;
+
+contract ContractRegistry {
```
dob: I'd add a comment specifying what this is used for, and the intended usage of the key => address mapping. For example, it's recommend that the key be the keccak hash of the name of the target contract.

contracts/bonding/IBondingManager.sol
```diff
@@ -0,0 +1,23 @@
...
+    // External functions
+    function transcoder(uint8 _blockRewardCut, uint8 _feeShare, uint256 _pricePerSegment) external returns (bool);
+    function resignAsTranscoder() external returns (bool);
+    function bond(uint _amount, address _to) external returns (bool);
+    function unbond() external returns (bool);
+    function withdraw() external returns (bool);
+    function reward() external returns (bool);
+
+    // Public functions
+    function setActiveTranscoders() public returns (bool);
```
dob: Hmm, what's unique about `setActiveTranscoders()` that would make it the only publicly scoped state modifying transaction across all three interfaces? Is it the only one that needs to be invoked by another manager?

contracts/LivepeerProtocol.sol
```diff
@@ -1,756 +1,28 @@
...
-                    rewards = rewards.add(delegatorShare);
-                }
-            }
-        }
+    function updateController(bytes32 _key, address _controller) onlyOwner public returns (bool) {
+        // Check if key is in registry
+        if (!registryContains(_key)) throw;
 
-        return rewards;
+        return registryGet(_key).call(bytes4(keccak256("setController(address)")), _controller);
```
dob: As this line is a little hairy, I think it at least warrants a comment of why we're using the `.call()` indirection through the registry.

contracts/bonding/IBondingManager.sol
yondonfu: I left it as public because I wasn't sure if it would ever be called internally by `BondingManager` if we change how round progression and active transcoder set selection in the future. But I do think it makes more sense for it to be marked `external` for now since it is only called by `RoundManager` externally and never internally. We can come back in the future and change it if neccessary.

package.json
```diff
@@ -7,7 +7,7 @@
     "test": "test"
   },
   "scripts": {
-    "test": "docker run -d -p 8545:8545 --name testrpc wbuchwalter/testrpc && truffle test",
+    "test": "docker run -d -p 8545:8545 --name testrpc yondon/testrpc && truffle test",
```
yondonfu: Switched to using our own Docker image for TestRPC that is more up to date

contracts/jobs/libraries/TranscodeJobs.sol
```diff
@@ -245,6 +248,18 @@ library TranscodeJobs {
     }
 
     /*
+     * @dev Compute the personal segment hash of a segment. Hashes the concatentation of the personal hash prefix and the segment hash
+     * @param _streamId Stream identifier
+     * @param _segmentSequenceNumber Segment sequence number in stream
+     * @param _dataHash Segment data hash
+     */
+    function personalSegmentHash(string _streamId, uint256 _segmentSequenceNumber, bytes32 _dataHash) internal constant returns (bytes32) {
```
yondonfu: As of TestRPC v4.0.0 the `eth_sign` RPC call (which `web3.eth.sign` uses) will accept an arbitrary message that it concatenates with a pre-defined prefix and then hashes the result. This "personal hash message" is then used in combination with a private key to generate a signature. https://github.com/ethereumjs/testrpc/pull/244 We replicate this behavior here, by generating the segment hash and then hashing the concatenation of the pre-defined prefix `personalHashPrefix` and the segment hash.

Issue #27, opened 2017-05-31
Verification interface


dob:
The plan is to eventually use Truebit for scalable verification of selected segments. However prior to Truebit's implementation, we should create a common interface that wraps various potential backends for verification. In the short term options for implementations include Oracle, 3x transcoder verification, and even the trivial "true" verifier that just stubs a result for testing. The interface should probably specify:

* The swarm hash of the verification code to run
* The swarm hash of the input data
* The function address to invoke with the output
* The interface to providing the output (is it also a Swarm hash?)

The interface is also likely `payable`, as we'd have to pay to use Truebit or an Oracle to do verification.

yondonfu:
Closed by #44

PR #44, opened 2017-07-24
Verification interface with IdentityVerifier

yondonfu:
Addresses #27 

Things to note:

- We will need to keep track of how many failed verifications a transcoder has. One option is to update the state of a transcoder in `BondingManager` from `JobsManager`. Another option is to keep track of failed verifications separately in a `VerificationManager` which might handle state for verification and slashing. I lean toward the latter at the moment since it isolates verification logic, but will explore more as we implement a non-trivial `Verifier` contract (i.e. Oraclize) that will actually return `false` for failed verifications.

### DIFF
--- a/test/jobs/JobsManager.js
+++ b/test/jobs/JobsManager.js
@@ -287,6 +287,15 @@ contract("JobsManager", accounts => {
             await expectThrow(jobsManager.verify(jobId, segmentSequenceNumber, utils.bufferToHex(d0), utils.bufferToHex(tD0), utils.bufferToHex(bSig0), proof, {from: accounts[0]}))
         })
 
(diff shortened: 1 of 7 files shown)


test/jobs/JobsManager.js
```diff
@@ -284,6 +287,15 @@ contract("JobsManager", accounts => {
             await expectThrow(jobsManager.verify(jobId, segmentSequenceNumber, utils.bufferToHex(d0), utils.bufferToHex(tD0), utils.bufferToHex(bSig0), proof, {from: accounts[0]}))
         })
 
+        it("should throw if segment not signed by broadcaster", async () => {
```
yondonfu: A test for `verify` in which the broadcaster signature is incorrect was previously missing, so added it in here

PR #47, opened 2017-07-31
Oraclize verification

yondonfu:
`OraclizeVerifier` implementation backed by Oraclize's off-chain computation service.

Had to make some Solidity type changes to get this to work. Details in in-line review comments.

Tests for `OraclizeVerifier` are run separately for now since Oraclize can take awhile to return a result.


.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  "extends": ["google"],
+  "extends": ["./node_modules/eslint-config-google/index.js"],
```
yondonfu: Using local eslint-config-google

contracts/jobs/libraries/TranscodeJobs.sol
```diff
@@ -179,13 +179,14 @@ library TranscodeJobs {
...
+    function validateTranscodeClaim(Jobs storage self,
                                      uint256 _jobId,
                                      uint256 _segmentSequenceNumber,
-                                     bytes32 _dataHash, bytes32 _transcodedDataHash,
+                                     string _dataHash,
+                                     string _transcodedDataHash,
                                      bytes _broadcasterSig,
                                      bytes _proof,
-                                     uint64 _verificationRate) constant returns (bool) {
+                                     uint64 _verificationRate) internal constant returns (bool) {
```
yondonfu: Due to the type change of `_dataHash` and `_transcodedDataHash` from `bytes32` to `string` I ran into a `Stack too deep, try removing local variables` compiler error. Local variables including function parameters and return parameters are placed on the stack and the EVM can only reach a specific depth for the stack - the current compiler defined max is 16. I need to investigate this further, but an initial guess as to why the compiler complained when this library function was public is that since library functions are invoked using `delegatecall`, all the variables are copied which perhaps leads to them being placed on the stack. When the library function is internal, library code is copied into the caller function and variables are referenced rather than copied.

contracts/verification/OraclizeVerifier.sol
```diff
@@ -0,0 +1,87 @@
...
+     * @param _jobId Job identifier
+     * @param _segmentSequenceNumber Segment being verified for job
+     * @param _code Content-addressed storage hash of binary to execute off-chain
+     * @param _dataHash Content-addressed storage hash of input data of segment
+     * @param _transcodedDataHash Content-addressed storage hash of transcoded input data of segment
+     * @param _callbackContract Address of Verifiable contract to call back
+     */
+    function verify(uint256 _jobId, uint256 _segmentSequenceNumber, string _code, string _dataHash, string _transcodedDataHash, address _callbackContract) payable sufficientOraclizeFunds external returns (bool) {
+        // Create Oraclize query
+        bytes32 queryId = oraclize_query("computation", [_code, _dataHash, _transcodedDataHash], 3000000);
```
yondonfu: `oraclize_query` can accept an array of arguments, but they have to either be of type `string` or `bytes`. We are using IPFS to store segment data and our Oraclize computation archive - since IPFS hashes are longer than 32 bytes, `_code`, `_dataHash` and `_transcodedDataHash` could not remain `bytes32` (note: it is possible to chop off the multihash identifier from the hex encoded bytes of an IPFS hash since IPFS only uses sha2-256 right now, but this isn't great for future proofing).

verification_computation_archive/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return Promise.all(results)
+}
+
+const verifyTranscoding = (dataHash, transcodedDataHash) => {
+    console.log("Verifying transcoding")
+    console.log("Segment data hosted on IPFS at: " + dataHash)
+    console.log("Transcoded data hash: " + transcodedDataHash)
+    return transcode(dataHash, "seg.ts", "out.ts").then(hash => {
+        return hash == transcodedDataHash
```
yondonfu: Compute the keccak256 hash of the transcoded segment data and compare it with the keccak256 hash passed in as a query argument

contracts/verification/OraclizeVerifier.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    // Check if sender is Oraclize
+    modifier onlyOraclize() {
+        if (msg.sender != oraclize_cbAddress()) throw;
+        _;
+    }
+
+    // Check if sufficient funds for Oraclize computation
+    modifier sufficientOraclizeFunds() {
+        if (oraclize_getPrice("computation") > this.balance) throw;
```
dob: Think we should check against this.balance or msg.amount? I'm wondering if there's a scenario where one transaction could use funds from another previous transaction without paying if this contract continued to accumulate balance somehow.

dob:
Two to-do's after an in person code walkthrough:

1. Remove the binaries from the repo if possible and replace with a README and a script that says how to pull them down and modify if necessary.
2. Don't pass the transcodedDataHash into oraclize. Instead let oraclize provide their computed transcodedDataHash to the callback function, and we can do the comparison there. One of the downsides is that this means we need to store the string in the contract for comparison later. But it feels more trustless if oraclize doesn't know the output that they are expected to compute.

PR #49, opened 2017-08-08
Updating to zeppelin 1.2.0

leos:
`truffle install zeppelin@1.2.0`


yondonfu:
Ah so ended up migrating to installing OpenZeppelin via NPM in my latest PR because NPM is more up to date with OpenZeppelin's latest releases and wanted to use one of the newer library contracts. This updated EthPM installation shouldn't be needed anymore. Sorry this wasn't communicated earlier!

leos:
Sounds great! I think NPM is a better idea than EthPM.

Issue #37, opened 2017-07-06
Replace deprecated throw with require(), assert()


dob:
`throw;` is now deprecated in the latest versions of solidity, and should instead be replaced by `require()` for checks on externally provided input data, and `assert()` for checks on internal conditions that should never achieve a certain state. See: 

https://solidity.readthedocs.io/en/develop/control-structures.html#error-handling-assert-require-revert-and-exceptions

Also related to #34, upgrading solidity.

yondonfu:
Done

Issue #33, opened 2017-06-14
Replace test values with realistic default values for time related variables


yondonfu:
Some of our protocol variables such as `roundLength`, `unbondingPeriod` and `jobEndingPeriod` are time related and have values denominated in blocks. In our tests, we use a helper function to fast forward a certain number of blocks by repeatedly using TestRPC's `evm_mine` request to mine a block. A large number of `evm_mine` requests slows down our tests a lot, so currently we use smaller test values for these protocol variables.

We should replace these smaller test values with realistic default values at a later point in time.

yondonfu:
This can be closed - we pass in these values in the constructor of contracts now

PR #50, opened 2017-08-10
Transcode job payments

### LINKED ISSUE #34: Upgrade to solidity 0.4.13

### LINKED ISSUE #43: Swap copy-and-pasted ECRecovery library for OpenZeppelin version
Our codebase currently uses a copy-and-pasted version of the ECRecovery library used for recovering a signer's address from an elliptic curve signature over a message. OpenZeppelin's latest release includes a refined version of that library. We should remove our copy-and-pasted library and instead use OpenZeppelin's library: https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/contracts/ECRecovery.sol

yondonfu:
Broadcasters submit deposits to the JobsManager contract to pay for job fees. Once a transcoder is elected for a job, the transcoder can submit claims via `claimWork`. Each claim kicks off a verification period followed by a slashing period. During the verification period, the transcoder is required to verify segments pseudorandomly selected by the protocol. If the transcoder fails verification, it is slashed. During the slashing period, a finder can prove to the JobsManager contract that a transcoder missed a verification for a segment in order to slash the transcoder. Anytime that a transcoder is slashed, all fees for a claim are returned from job escrow to the broadcaster and the transcoder is penalized.

Note: delegator slashing when a transcoder is slashed is not yet implemented

Using OpenZeppelin's ECRecovery library instead of our own copy and pasted version - resolves #43 
Switched to using npm from ethpm for installing OpenZeppelin contracts

Upgraded to Solidity 0.4.13 and using require, assert, revert instead of throw as throw is now deprecated - resolves #34 and #37 

Initial pass at #33 but is currently untested


contracts/bonding/BondingManager.sol
```diff
@@ -498,41 +529,133 @@ contract BondingManager is IBondingManager, Controllable {
         return transcoderPools.transcoderStake(_transcoder);
     }
 
+    function transcoderRates(address _transcoder) public constant returns (uint256, uint256, uint256, uint256, uint256, uint256) {
```
dob: Let's add a comment for documentation generation purposes about what people can expect the 6 return values to be.

contracts/bonding/BondingManager.sol
```diff
@@ -498,41 +529,133 @@ contract BondingManager is IBondingManager, Controllable {
...
+            return DelegatorStatus.Inactive;
+        }
+    }
 
-        return delegators[_delegator].bondedAmount.add(delegatorRewards(_delegator));
+    /*
+     * @dev Return number of minted tokens for a reward call
+     */
+    function mintedTokensPerReward() public constant returns (uint256) {
+        return initialTokenSupply.mul(initialYearlyInflation).div(100).div(roundsManager().rewardCallsPerYear());
```
dob: Is this assuming 1 reward call per round max? I'm trying to remember if that's our constraint around how block rewards get distributed? And then subsequent calls will throw?

contracts/bonding/BondingManager.sol
```diff
@@ -498,41 +529,133 @@ contract BondingManager is IBondingManager, Controllable {
...
 
-        // Check if delegator bonded to a transcoder
-        if (delegators[_delegator].transcoderAddress != address(0)) {
+        if (del.transcoderAddress != address(0)) {
             // Iterate from round that delegator last transitioned states to current round
-            // If the delegator is bonded to a transcoder, it has been bonded to the transcoder since lastStateTransitionRound
-            for (uint256 i = delegators[_delegator].lastStateTransitionRound; i <= roundsManager().currentRound(); i++) {
-                uint256[2] rewardMultiplier = rewardMultiplierPerTranscoderAndRound[delegators[_delegator].transcoderAddress][i];
+            // If the delegator is bonded to a transcoder, it has been bonded to the transcoder since lastStakeUpdateRound
+            for (uint256 i = del.lastStakeUpdateRound; i <= roundsManager().currentRound(); i++) {
```
dob: Note for the future that we'll probably want to use the loop limit pattern where this executes the min of # of rounds since last state transition, and some safe loop param like 100. And then it will need to be called multiple times to get current. I'll file an issue.

contracts/bonding/BondingManager.sol
```diff
@@ -498,41 +529,133 @@ contract BondingManager is IBondingManager, Controllable {
...
+        } else {
+            return rewardPool.rewards.mul(del.bondedAmount).div(rewardPool.transcoderTotalStake);
+        }
+    }
+
+    /*
+     * @dev Computes delegator's share of fee pool for a round
+     */
+    function delegatorFeePoolShare(Delegator storage del, uint256 _round) internal constant returns (uint256) {
+        ClaimFees[] storage feePool = transcoders[del.transcoderAddress].tokenPoolsPerRound[_round].feePool;
```
dob: Let's do a quick in person walkthrough of the feePool and tokenPoolsPerRound structures.

contracts/ContractRegistry.sol
```diff
@@ -1,22 +1,19 @@
...
         return true;
     }
 
-    function registryGet(bytes32 _key) internal constant returns (address) {
-        return registry[_key];
-    }
-
-    function registryContains(bytes32 _key) internal constant returns (bool) {
-        return registry[_key] != address(0x0);
+    function updateManagerRegistry(bytes32 _key, address _registry) onlyOwner whenPaused public returns (bool) {
```
yondonfu: Yeah all state mutating functions would need `whenNotPaused` modifiers to ensure that they can't be called during a pause. I wonder if there is a benefit in implementing those modifiers such that they will check if `LivepeerProtocol` (the main entry point into the system) is paused - if `LivepeerProtocol` is paused then all manager contracts are paused as well.

It might look like this:
```
modifier whenRegistryNotPaused() {
    require(!ContractRegistry(registry).paused);
    _;
}
```

This might make it easier to perform a system wide pause. Thoughts?

contracts/bonding/BondingManager.sol
```diff
@@ -28,33 +29,52 @@ contract BondingManager is IBondingManager, Controllable {
...
+        uint256 delegateBlock;             // The block the delegator bonds to a transcoder
         uint256 withdrawRound;             // The round at which a delegator can withdraw
-        uint256 lastStateTransitionRound;  // The last round the delegator transitioned states
+        uint256 lastStakeUpdateRound;      // The last round the delegator transitioned states
         bool initialized;                  // Is this delegator initialized
     }
 
     // The various states a delegator can be in
-    enum DelegatorStatus { Inactive, Pending, Bonded, Unbonding }
+    enum DelegatorStatus { Inactive, Pending, Bonded, Unbonding, Unbonded }
```
yondonfu: Hm I don't think there is a difference at the moment, so `Inactive` can probably just be scrapped and delegators would always default to the `Unbonded` state

contracts/bonding/BondingManager.sol
```diff
@@ -71,22 +91,23 @@ contract BondingManager is IBondingManager, Controllable {
...
+        require(msg.sender == address(jobsManager()));
         _;
     }
 
     /*
      * @dev BondingManager constructor. Sets a pre-existing address for the LivepeerToken contract
      * @param _token LivepeerToken contract address
      */
-    function BondingManager(address _token, uint256 _numActiveTranscoders) {
+    function BondingManager(address _registry, address _token, uint256 _numActiveTranscoders) Manager(_registry) {
```
yondonfu: It is the constructor for the base contract (http://solidity.readthedocs.io/en/develop/contracts.html#arguments-for-base-constructors). So in this case since `BondingManager` inherits from `Manager` it needs to invoke the constructor for `Manager`

contracts/bonding/BondingManager.sol
```diff
@@ -224,32 +219,38 @@ contract BondingManager is IBondingManager, Controllable {
...
-     * @dev Unbond your current stake. You will enter the unbonding phase for
-     * the unbondingPeriod.
+    /*
+     * @dev Unbond delegator's current stake. Delegator enters unbonding state
      */
     function unbond() external returns (bool) {
-        // Check if current round is initialized
-        if (!roundsManager().currentRoundInitialized()) throw;
-        // Check if this is an initialized delegator
-        if (delegators[msg.sender].initialized == false) throw;
```
yondonfu: If the sender passes this check `require(delegatorStatus(msg.sender) == DelegatorStatus.Bonded);` `delegator[msg.sender].initialized` cannot be `false` because the only times it is `false` is if the delegator is not registered (i.e. never bonded to a transcoder, or has unbonded and withdrawn) and if that is the case, `delegator[msg.sender].startRound` would be 0 so the sender wouldn't pass that first check

contracts/bonding/BondingManager.sol
```diff
@@ -351,141 +303,220 @@ contract BondingManager is IBondingManager, Controllable {
...
+        onlyJobsManager
+        returns (bool)
+    {
+        Transcoder storage t = transcoders[_transcoder];
+
+        // Transcoder must be valid
+        require(t.initialized);
+
+        uint256 currentRound = roundsManager().currentRound();
+        uint256 delegatorsFeeShare = _fees.mul(t.rates[1]).div(100);
```
yondonfu: I think we can use a struct here and reference each rate by field name i.e. `rates.blockRewardCut`, `rates.feeShare`, etc.

contracts/bonding/BondingManager.sol
yondonfu: I believe we said each active transcoder can call `reward` once per round. `roundsManager().rewardCallsPerYear()` just calculates the number of reward calls per year given the current round length and the protocol defined number of active transcoders

contracts/ContractRegistry.sol
dob: Yah I like that idea. Or `whenProtocolNotPaused()`.

contracts/jobs/JobsManager.sol
```diff
@@ -26,40 +37,122 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
     }
 
     /*
-     * @dev Returns BondingManager
+     * @dev Deposit funds for jobs
+     * @param _amount Amount to deposit
      */
-    function bondingManager() internal constant returns (IBondingManager) {
-        LivepeerProtocol protocol = LivepeerProtocol(controller);
+    function deposit(uint256 _amount) external returns (bool) {
```
dob: Any advantage to using `deposit` and `withdraw` transactions instead of just using the ERC20 `approve` method? Whatever the `allowance` is for the protocol could be considered the deposited balance at the moment. Have to dig into ERC20 to make sure `allowance` updates correctly if their balance changes though.

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+    function claimWork(uint256 _jobId, uint256[2] _segmentRange, bytes32 _claimRoot) jobExists(_jobId) external returns (bool) {
+        Job storage job = jobs[_jobId];
+
+        // Job must be active
+        require(jobStatus(_jobId) == JobStatus.Active);
+        // Sender must be elected transcoder
+        require(job.transcoderAddress == msg.sender);
+
+        // Move fees from broadcaster deposit to escrow
+        uint256 fees = _segmentRange[1].sub(_segmentRange[0]).add(1).mul(job.maxPricePerSegment);
```
dob: let's add a `require(_segmentRange[1] > _segmentRange[0])`

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
      */
-    function getJobDetails(uint256 _jobId) public constant returns (uint256, uint256, address, address, uint256) {
-        return jobs.getJobDetails(_jobId);
+    function verify(
+        uint256 _jobId,
+        uint256 _claimId,
+        uint256 _segmentNumber,
+        string _dataHash,
+        string _transcodedDataHash,
+        bytes _broadcasterSig,
```
dob: I think `bytes` is an alias for `string` right? We can probably specify the length of these hashes/signatures.

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+     * @param _segmentNumber Segment being verified for job
+     * @param _result Boolean result of whether verification succeeded or not
      */
-    function getJobStreamId(uint256 _jobId) public constant returns (string) {
-        return jobs.jobs[_jobId].streamId;
+    function receiveVerification(uint256 _jobId, uint256 _claimId, uint256 _segmentNumber, bool _result) onlyVerifier external returns (bool) {
+        if (!_result) {
+            refundBroadcaster(_jobId, _claimId);
+            // Protocol slashes transcoder for failing verification (no finder)
+            bondingManager().slashTranscoder(jobs[_jobId].transcoderAddress, address(0), failedVerificationSlashAmount, 0);
```
dob: just to clarify, this is burning the tokens, right?

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+     * @dev Slash transcoder for missing verification
      * @param _jobId Job identifier
+     * @param _claimId Claim identifier
+     * @param _segmentNumber Segment that was not verified
      */
-    function getJobTranscodeClaimsDetails(uint256 _jobId) public constant returns (uint256, uint256, uint256, uint256, bytes32) {
-        return jobs.getJobTranscodeClaimsDetails(_jobId);
+    function missedVerificationSlash(uint256 _jobId, uint256 _claimId, uint256 _segmentNumber) external returns (bool) {
+        Job storage job = jobs[_jobId];
+        Claim storage claim = job.claims[_claimId];
```
dob: I believe this is an index into an array, and therefore will throw if the claim doesn't exist. Just sanity checking that since we're not explicitly checking this returns a valid claim (if it were a mapping I think we'd have to check.)

dob: For example in `distributeFees` I think we check that claimID is < then claims.length

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+        // Claim must not be complete
+        require(claim.status == ClaimStatus.Pending);
+        // Slashing period must be over for claim
+        require(claim.endSlashingBlock < block.number);
+
+        uint256 fees = claim.segmentRange[1].sub(claim.segmentRange[0]).add(1).mul(job.maxPricePerSegment);
+        // Deduct fees from escrow
+        job.escrow = job.escrow.sub(fees);
+        // Add fees to transcoder's fee pool
+        bondingManager().updateTranscoderFeePool(msg.sender, fees, claim.claimBlock, claim.transcoderTotalStake);
```
dob: Do we need to assign the transcoder their cut of the fees here, or is that taken care of somewhere else?

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+        uint256 _jobId,
+        uint256 _claimId,
+        uint256 _segmentNumber,
+        string _dataHash,
+        string _transcodedDataHash,
+        bytes _broadcasterSig,
+        bytes _proof
+    )
+        internal
+        returns (bool)
```
dob: should this be a constant function?

contracts/jobs/IJobsManager.sol
```diff
@@ -6,8 +6,4 @@ pragma solidity ^0.4.11;
  */
 contract IJobsManager {
     // External functions
-    function job(string _streamId, string _transcodingOptions, uint256 _maxPricePerSegment) external returns (bool);
-    function endJob(uint256 _jobId) external returns (bool);
-    function claimWork(uint256 _jobId, uint256 _startSegmentSequenceNumber, uint256 _endSegmentSequenceNumber, bytes32 _transcodeClaimsRoot) external returns (bool);
-    function verify(uint256 _jobId, uint256 _segmentSequenceNumber, string _dataHash, string _transcodedDataHash, bytes _broadcasterSig, bytes _proof) payable external returns (bool);
```
yondonfu: None of the manager contracts currently invoke of the functions in `JobsManager` so the abstract contract `IJobsManager` doesn't need to declare any functions.

contracts/jobs/JobsManager.sol
```diff
@@ -1,19 +1,30 @@
...
+    using SafeMath for uint256;
+
+    // Token address
+    LivepeerToken public token;
 
     // Verifier address
     Verifier public verifier;
 
+    // Content-addressed storage hash of verification code
+    string public verificationCodeHash;
```
yondonfu: This should be removed from `JobsManager` because `verificationCodeHash` is now set in `OraclizeVerifier`. This was prompted because I was running into a `Stack too deep` compiler error for `JobsManager.verify()` which passed in `verificationCodeHash` as a parameter to `verifier.verify()` (https://github.com/livepeer/protocol/blob/71698abc2887a596c63745d02851513ec26bd62e/contracts/jobs/JobsManager.sol#L280). The compiler error was solved by removing `verificationCodeHash` as a parameter to `verifier.verify()` and instead having it be set in `OraclizeVerifier` which is currently the only contract that uses it

contracts/jobs/JobsManager.sol
yondonfu: I think we talked about this on Gitter. A broadcaster could approve `JobsManager` as a spender for some amount of tokens and then transfer the tokens without updating the allowance for `JobsManager` such that the allowance could be more than the broadcaster's actual balance.

contracts/jobs/JobsManager.sol
yondonfu: They are both dynamically sized arrays, but `string` is `UTF-8` encoded. We use `string` right now for Oraclize compatibility. But yeah in this case the length of `_dataHash` (IPFS hash) and `_transcodedDataHash` (32 byte hex encoded hash) and `_broadcasterSig` are all known.

contracts/jobs/JobsManager.sol
yondonfu: The tokens are currently just left in the BondingManager contract and not associated with anyone. We could send them to a burn address as well if we want to actually burn them

contracts/jobs/JobsManager.sol
yondonfu: `updateTranscoderFeePool` handles it

contracts/jobs/JobsManager.sol
yondonfu: Ah so there needs to be explicit check for `_jobId` since `jobs` is a mapping. The explicit check for `_claimId` isn't necessary since `job.claims` is an array.

contracts/jobs/JobsManager.sol
```diff
@@ -70,94 +163,300 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+     * @dev End a job. Can be called by either a broadcaster or transcoder of a job
+     * @param _jobId Job identifier
+     */
+    function endJob(uint256 _jobId) external returns (bool) {
+        Job storage job = jobs[_jobId];
+
+        // Job must not already have an end block
+        require(job.endBlock == 0);
+        // Sender must be the job's broadcaster or elected transcoder
+        require(job.broadcasterAddress == msg.sender || jobs[_jobId].transcoderAddress == msg.sender);
```
yondonfu: `jobs[_jobId].transcoderAddress == msg.sender` can be simplified to `job.transcoderAddress == msg.sender`

contracts/jobs/JobsManager.sol
yondonfu: Yeah I agree that the price should be `transcoder.rates.pricePerSegment` instead of `job.maxPricePerSegment` the rationale being that a transcoder can set the lowest price per segment it is willing to accept and if a job's max price per segment is higher than that, the transcoder would still be happy to accept the job at the price it originally set.

contracts/bonding/BondingManager.sol
```diff
@@ -28,33 +29,54 @@ contract BondingManager is IBondingManager, Controllable {
...
-        uint256 pricePerSegment;        // Lowest price transcoder is willing to accept for a job. Denominated in LPT base units
-        uint8 pendingBlockRewardCut;    // Pending value for blockRewardCut to be set at the beginning of a new round
-        uint8 pendingFeeShare;          // Pending value for feeShare to be set at the beginning of a new round
-        uint256 pendingPricePerSegment; // Pending value for pricePerSegment to be set at the beginning of a new round
-        bool active;                    // Is this transcoder active. Also will be false if uninitialized
+        address transcoderAddress;                           // The address of this transcoder
+        uint256 bondedAmount;                                // The amount they have bonded themselves
+        uint256 withdrawRound;                               // The round at which delegators to this transcoder can withdraw if this transcoder resigns
+        uint256 lastRewardRound;                             // Last round that the transcoder called reward
+        uint8 blockRewardCut;                                // % of block reward cut paid to transcoder by a delegator
```
yondonfu: Ended up switching back to listing `blockRewardCut`, `feeShare` and `pricePerSegment` individually. Originally wanted to switch to using `TranscodeRates` struct instead of an array to represent these fields so that we could reference each field using the struct i.e. `transcoders[msg.sender].rates.blockRewardCut`. However, Solidity does not allow the generation of public getters for structs that have struct fields. We can consider writing custom getters for the transcoder mapping if we really want to include structs a fields, but for now listing `blockRewardCut`, `feeShare` and `pricePerSegment` as separate fields allows us to keep the automatically generated getter

contracts/bonding/BondingManager.sol
```diff
@@ -350,142 +309,231 @@ contract BondingManager is IBondingManager, Controllable {
...
+        returns (bool)
+    {
+        // Transcoder must be valid
+        require(transcoderStatus(_transcoder) == TranscoderStatus.Registered);
+
+        Transcoder storage t = transcoders[_transcoder];
+
+        uint256 penalty = t.bondedAmount.mul(_slashAmount).div(100);
+
+        if (penalty > t.bondedAmount) {
```
yondonfu: Emptying out a transcoder's bond if the penalty is greater than its current bond amount

contracts/bonding/BondingManager.sol
```diff
@@ -350,142 +309,231 @@ contract BondingManager is IBondingManager, Controllable {
...
+
+        return true;
+    }
+
+    /*
+     * @dev Pseudorandomly elect a currently active transcoder that charges a price per segment less than or equal to the max price per segment for a job
+     * Returns address of elected active transcoder and its price per segment
+     * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
+     */
+    function electActiveTranscoder(uint256 _maxPricePerSegment) external constant returns (address, uint256) {
```
yondonfu: This function also returns an elected transcoder's price per segment so that the jobs manager can use that value to calculate fees for a job

contracts/jobs/JobsManager.sol
```diff
@@ -68,96 +154,329 @@ contract JobsManager is IJobsManager, Verifiable, Controllable {
...
+        // Slashing period must be over for claim
+        require(claim.endSlashingBlock < block.number);
+
+        uint256 fees = claim.segmentRange[1].sub(claim.segmentRange[0]).add(1).mul(job.pricePerSegment);
+        // Deduct fees from escrow
+        job.escrow = job.escrow.sub(fees);
+        // Add fees to transcoder's fee pool
+        bondingManager().updateTranscoderFeePool(msg.sender, fees, claim.claimBlock, claim.transcoderTotalStake);
+        // Send fees to bonding manager
+        token.transfer(address(bondingManager()), fees);
```
yondonfu: This was missing before. Now this function actually sends tokens deposited to the JobsManager to the BondingManager so the BondingManager can send tokens to delegators and transcoders when they withdraw

contracts/bonding/BondingManager.sol
dob: Let's have a future discussion about charging the transcoder fee or the max price per segment submitted by the broadcaster.

PR #59, opened 2017-08-25
Configurable deployment scripts

yondonfu:
Updated deployment scripts in `migrations` so that they can be configurable based on what network we are deploying to. The scripts use a config file to specify the protocol params used in our contract constructors (params are now set by constructors rather than hardcoded in the contract). Currently there is only one config file `migrations.config.js`, but we could create a `dev.config.js` for TestRPC, `testnet.config.js` for our private test net and `rinkeby.config.js` for the rinkeby test net.

Transcoders are also now rewarded relative to their proportional stake. For every round, there is a token reward associated with it calculated by `roundReward = secondsInYear / blockTime / roundLength`. When an active transcoder calls reward, their reward is calculated by `transcoderReward = (roundReward * transcoderActiveStake) / totalActiveTranscoderStake`.


yondonfu:
Closing in favor of #60

PR #60, opened 2017-08-28
Faucet + configurable deployment scripts

yondonfu:
Copied from #59: Updated deployment scripts in migrations so that they can be configurable based on what network we are deploying to. The scripts use a config file to specify the protocol params used in our contract constructors (params are now set by constructors rather than hardcoded in the contract). Currently there is only one config file migrations.config.js, but we could create a dev.config.js for TestRPC, testnet.config.js for our private test net and rinkeby.config.js for the rinkeby test net.

Transcoders are also now rewarded relative to their proportional stake. For every round, there is a token reward associated with it calculated by roundReward = secondsInYear / blockTime / roundLength. When an active transcoder calls reward, their reward is calculated by transcoderReward = (roundReward * transcoderActiveStake) / totalActiveTranscoderStake.

New to this PR: faucet contract with rate limit and whitelist. Faucet contract can be deployed on our private test network with configurable values for the amount minted to the faucet, amount to be sent to a requestor, the amount of time a requestor must wait between requests and a list of addresses to be whitelisted. Addresses first part of https://github.com/livepeer/go-livepeer/issues/60 - next will be a user interface for the faucet


yondonfu:
Yep, will follow up with a PR in go-livepeer

PR #63, opened 2017-08-30
Keep track of all registered transcoders

yondonfu:
Added an array to keep track of addresses of all registered transcoders so a client can iterate through the array to retrieve addresses and then use the addresses to get transcoder stats from the `transcoders` map.

Also, defaulting to deploying with the `IdentityVerifier` for all networks for the time being until `OraclizeVerifier` + IPFS node integration is ready.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -83,9 +83,6 @@ contract BondingManager is IBondingManager, Manager {
     mapping (address => Delegator) public delegators;
     mapping (address => Transcoder) public transcoders;
 
-    // Keep track of all transcoder addresses
-    address[] public registeredTranscoders;
-
     // Active and candidate transcoder pools
     TranscoderPools.TranscoderPools transcoderPools;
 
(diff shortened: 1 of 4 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -152,6 +155,8 @@ contract BondingManager is IBondingManager, Manager {
         t.pendingFeeShare = _feeShare;
         t.pendingPricePerSegment = _pricePerSegment;
 
+        registeredTranscoders.push(msg.sender);
```
dob: Probably should add a check so that a transcoder doesn't get added if it already exists.

yondonfu:
Updated with getters for retrieving the current size of the candidate transcoder pool and reserve transcoder pool in addition to getters for retrieving transcoder addresses from their respective pools by index/position

contracts/bonding/BondingManager.sol
```diff
@@ -276,11 +280,11 @@ contract BondingManager is IBondingManager, Manager {
         // Current round must be initialized
         require(roundsManager().currentRoundInitialized());
 
-        if (transcoderStatus(msg.sender) == TranscoderStatus.Unbonding) {
+        if (transcoderStatus(msg.sender) == TranscoderStatus.Unbonded) {
```
dob: Good catch. Was this a bug before?

yondonfu: Yeah...caught while running the node. Will need to write more thorough test suites

PR #66, opened 2017-09-05
Transcoder earns share of reward as a delegator. Fixes #64

### LINKED ISSUE #64: Transcoder with 0 delegators block reward share and fee share
Right now, a `transcoder` that has 0 delegators and blockRewardCut < 100 and feeShare > 0 will take its block reward and fee shares for `currentRound`, and leave the rest for its delegators in `transcoders[transcoder].tokenPoolsPerRound[currenRound]`. However, since `transcoder` has no delegators, no one will ever be able to claim these tokens.

In this case, I think the transcoder should just take all the fees and rewards. If its delegator count > 0, then the transcoder will only claim its own share of the block reward and fee share.

yondonfu:
When a transcoder calls reward, it should receive not only the block reward cut, but also its sahre of the delegator rewards if it is bonded to itself. Calculate transcoder reward share as (mintedTokens * blockRewardCut / 100) + delegatorRewardShare * transcoderBond / transcoderTotalStake where delegatorRewardShare = mintedTokens * (100 - blockRewardCut) / 100.

Opening this PR with develop as the base. I was thinking we use master as the branch that reflects the current deployment of the contracts on the test net, and use develop as the latest version of the contracts that might be used in a future deployment.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -360,11 +360,16 @@ contract BondingManager is IBondingManager, Manager {
         /* // Mint token reward and allocate to this protocol contract */
         token.mint(this, mintedTokens);
 
-        // Compute transcoder share of minted tokens
-        uint256 transcoderRewardShare = mintedTokens.mul(t.blockRewardCut).div(100);
+        // Compute reward share for delegators
+        uint256 delegatorsRewardShare = mintedTokens.mul(uint256(100).sub(t.blockRewardCut)).div(100);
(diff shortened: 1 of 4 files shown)


dob:
Independent comment before I get into the review: Let's talk about the branching strategy. I agree that having a known version that's pristine is a good idea, so whether that's master, or just a tag on master, we should know what commit is deployed. In the past we've used a "staging" branch to merge into before finally merging to master to deploy, but I think Github may give us some benefits if we're merging into master continuously. Let's discuss.

yondonfu:
Closed in favor of #70

PR #70, opened 2017-09-07
Separating transcoder candidacy from delegation

yondonfu:
A spike to see how the code looks when transcoder candidacy is separated from delegation - opening PR to get some eyes on the structure, not meant to be considered for a merge yet. There are no updated tests yet.

- `bond` can be called on anyone
- If a user bonds to himself and calls `transcoder`, we try to add the user to the transcoder pools with his delegated stake. If a user is _not_ bonded to himself and calls `transcoder`, we try to add the user to the transcoder pools with 0 stake
- If a user resigns as a transcoder, he enters the `Resigned` state and the `delegatorWithdrawRound` field is set. When `delegatorWithdrawRound` is now or in the past, the user enters the `NotRegistered` state. A user can only re-register as a transcoder if he is in the `NotRegistered` state. This is to prevent a user from resigning and immediately re-registering
- All bonded stake is tracked in the `bondedAmount` field of a `Delegator` struct
- If a user is not a bonded delegator and `reward` or `updateTranscoderFeePool` is called, it becomes a bonded delegator and the appropriate values are set
- Note that the changes in this spike allow for a user to be a registered transcoder and have bonded stake delegated to someone else

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -29,9 +29,7 @@ contract BondingManager is IBondingManager, Manager {
 
     // Represents a transcoder's current state
     struct Transcoder {
-        address transcoderAddress;                           // The address of this transcoder
-        uint256 bondedAmount;                                // The amount they have bonded themselves
-        uint256 withdrawRound;                               // The round at which delegators to this transcoder can withdraw if this transcoder resigns
+        uint256 delegatorWithdrawRound;                      // The round at which delegators to this transcoder can withdraw if this transcoder resigns
         uint256 lastRewardRound;                             // Last round that the transcoder called reward
         uint8 blockRewardCut;                                // % of block reward cut paid to transcoder by a delegator
         uint8 feeShare;                                      // % of fees paid to delegators by transcoder
@@ -43,7 +41,7 @@ contract BondingManager is IBondingManager, Manager {
     }
 
     // The various states a transcoder can be in
-    enum TranscoderStatus { NotRegistered, Registered, Unbonding, Unbonded }
+    enum TranscoderStatus { NotRegistered, Registered, Resigned }
 
     // Represents rewards and fees to be distributed to delegators
     struct TokenPools {
@@ -66,9 +64,8 @@ contract BondingManager is IBondingManager, Manager {
 
     // Represents a delegator's current state
     struct Delegator {
-        address delegatorAddress;          // The ethereum address of this delegator
-        uint256 bondedAmount;              // The amount they have bonded
-        address transcoderAddress;         // The ethereum address of the transcoder they are delgating towards
+        uint256 bondedAmount;              // The amount of bonded tokens
+        address delegationTarget;          // The address delegated to
         uint256 startRound;                // The round the delegator transitions to bonded phase
         uint256 delegateBlock;             // The block the delegator bonds to a transcoder
         uint256 withdrawRound;             // The round at which a delegator can withdraw
@@ -76,7 +73,7 @@ contract BondingManager is IBondingManager, Manager {
     }
 
     // The various states a delegator can be in
-    enum DelegatorStatus { NotRegistered, Pending, Bonded, Unbonding, Unbonded }
+    enum DelegatorStatus { Pending, Bonded, Unbonding, Unbonded }
 
     // Keep track of the known transcoders and delegators
     // Note: Casper style implementation would have us using arrays and bitmaps to index these
@@ -107,6 +104,26 @@ contract BondingManager is IBondingManager, Manager {
         _;
     }
 
+    // Check if current round is initialized
+    modifier currentRoundInitialized() {
+        require(roundsManager().currentRoundInitialized());
+        _;
+    }
+
+    // Update delegator stake with rewards/fees from past rounds
+    modifier updateDelegatorStake() {
+        Delegator storage del = delegators[msg.sender];
+
+        if (delegatorStatus(msg.sender) == DelegatorStatus.Bonded && transcoderStatus(del.delegationTarget) == TranscoderStatus.Registered) {
+            // Update stake if delegator is bonded to a registered transcoder
+            del.bondedAmount = del.bondedAmount.add(delegatorTokenPoolsShare(del));
+        }
+
+        del.lastStakeUpdateRound = roundsManager().currentRound();
+
+        _;
+    }
+
     /*
      * @dev BondingManager constructor. Sets a pre-existing address for the LivepeerToken contract
      * @param _token LivepeerToken contract address
@@ -116,17 +133,16 @@ contract BondingManager is IBondingManager, Manager {
         address _token,
         uint256 _numActiveTranscoders,
         uint64 _unbondingPeriod
-    ) Manager(_registry) {
+    )
+        Manager(_registry)
+    {
         // Set LivepeerToken address
         token = LivepeerToken(_token);
-
         // Set unbonding period
         unbondingPeriod = _unbondingPeriod;
-
         // Set up transcoder pools
         transcoderPools.init(_numActiveTranscoders, _numActiveTranscoders);
     }
-
 
     /*
      * @dev The sender is declaring themselves as a candidate for active transcoding.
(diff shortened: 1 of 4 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -137,23 +153,33 @@ contract BondingManager is IBondingManager, Manager {
...
+            t.delegatorWithdrawRound = 0;
+
+            if (delegatorStatus(msg.sender) == DelegatorStatus.Bonded && delegators[msg.sender].delegationTarget == msg.sender) {
+                // Sender is delegated to itself
+                // Add new transcoder with delegated stake to pools. Fails for insufficient delegated stake
+                transcoderPools.addTranscoder(msg.sender, delegators[msg.sender].bondedAmount);
+            } else {
+                // Sender is not delegated to itself
+                // Add new transcoder with zero stake to pools. Fails if pools are full
+                transcoderPools.addTranscoder(msg.sender, 0);
```
dob: Is there any chance that someone else has delegated towards this user, and therefore their total stake would not be == 0?

contracts/bonding/BondingManager.sol
```diff
@@ -184,57 +204,42 @@ contract BondingManager is IBondingManager, Manager {
...
+            uint256 stakeForTranscoder = _amount;
 
-            if (del.transcoderAddress != address(0) && _to != del.transcoderAddress) {
-                // Delegator is moving bond
-                // Set round start if delegator moves bond to another active transcoder
-                del.startRound = roundsManager().currentRound().add(1);
-                // Decrease former transcoder cumulative stake
-                transcoderPools.decreaseTranscoderStake(del.transcoderAddress, del.bondedAmount);
-                // Stake amount includes delegator's total bonded amount since delegator is moving its bond
+            if (del.delegationTarget != address(0) && _to != del.delegationTarget) {
```
ericxtang: Not exactly related to what this PR is about - 

I get what we are trying to do here.  But what are we trying to accomplish by allowing people to "change bond"?  In reality, can't I unbond and then bond to someone else?

contracts/bonding/BondingManager.sol
```diff
@@ -248,50 +253,44 @@ contract BondingManager is IBondingManager, Manager {
...
      * @dev Unbond delegator's current stake. Delegator enters unbonding state
+     * @param _amount Amount of tokens to unbond
      */
-    function unbond() external whenSystemNotPaused returns (bool) {
-        // Current round must be initialized
-        require(roundsManager().currentRoundInitialized());
-
-        Delegator storage del = delegators[msg.sender];
-
+    function unbond(uint256 _amount)
```
dob: Is _amount used in this method? It looks like it unbonds the del.bondedAmount.

contracts/bonding/BondingManager.sol
```diff
@@ -248,50 +253,44 @@ contract BondingManager is IBondingManager, Manager {
...
-        } else if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded){
-            token.transfer(msg.sender, delegators[msg.sender].bondedAmount);
+        token.transfer(msg.sender, delegators[msg.sender].bondedAmount);
 
-            delete delegators[msg.sender];
-        } else {
-            // Sender is neither a transcoder or delegator
-            revert();
-        }
+        delete delegators[msg.sender];
```
dob: Will this be a problem if someone withdraws, but then receives a reward or fee payout after the fact?

contracts/bonding/BondingManager.sol
```diff
@@ -357,11 +354,23 @@ contract BondingManager is IBondingManager, Manager {
...
         // Calculate number of tokens to mint
         uint256 mintedTokens = mintedTokensPerReward(msg.sender);
-        /* // Mint token reward and allocate to this protocol contract */
+        // Mint token reward and allocate to this protocol contract
         token.mint(this, mintedTokens);
 
         // Compute transcoder share of minted tokens
         uint256 transcoderRewardShare = mintedTokens.mul(t.blockRewardCut).div(100);
+
+        delegators[msg.sender].bondedAmount = delegators[msg.sender].bondedAmount.add(transcoderRewardShare);
```
ericxtang: So if I'm a delegator, I'm expected to call reward now?

contracts/bonding/BondingManager.sol
```diff
@@ -408,10 +415,20 @@ contract BondingManager is IBondingManager, Manager {
...
+
+        delegators[msg.sender].bondedAmount = delegators[msg.sender].bondedAmount.add(transcoderFeeShare);
+
+        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded) {
+            // Create new delegator if transcoder is not already a bonded delegator
+            delegators[msg.sender].delegationTarget = msg.sender;
+            delegators[msg.sender].startRound = currentRound;
+            delegators[msg.sender].delegateBlock = block.number;
+            delegators[msg.sender].withdrawRound = 0;
+            delegators[msg.sender].lastStakeUpdateRound = currentRound;
```
dob: I think all this logic might be repeated in `reward()` as well. Maybe a convenience method for having a new delegator be initialized in one place would be good to isolate the logic.

contracts/bonding/BondingManager.sol
```diff
@@ -29,9 +29,7 @@ contract BondingManager is IBondingManager, Manager {
 
     // Represents a transcoder's current state
     struct Transcoder {
-        address transcoderAddress;                           // The address of this transcoder
```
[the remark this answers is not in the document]
yondonfu: Yeah having the address in the struct in addition to using as the key in the mapping didn't seem necessary and removing it also saves some gas since there is one less store op

contracts/bonding/BondingManager.sol
yondonfu: Ah I started thinking about unbonding specific amounts of stake, but didn't end up implementing it for this spike. I'll remove it for now

contracts/bonding/BondingManager.sol
yondonfu: If you are a delegator that unbonds and then withdraws, you shouldn't be eligible for any future rewards or fee payouts since you are no longer bonded to the transcoder. If you are a transcoder and you unbond/withdraw your stake, the next time you receive rewards or fee payouts, a delegator struct corresponding to your address will be populated with the correct amounts - essentially re-instantiating a bonded delegator

contracts/bonding/BondingManager.sol
yondonfu: No, so now if I am a registered transcoder I might have 2 structs on-chain that store data relevant to me associated with my address. The `Transcoder` struct stores transcoding relevant info i.e. blockRewardCut, feeShare. The `Delegator` struct stores bonded amount, withdraw round, etc. When I call reward, the bonded amount field in the `Delegator` struct associated with my address will be updated. Thus, if I want to unbond and withdraw my bonded stake I do so as a delegator. This is to separate transcoder candidacy and delegation - in this case I am both a transcoder and a delegator (likely delegated to myself).

contracts/bonding/BondingManager.sol
yondonfu: Yeah I wasn't sure how to handle this case without including more on-chain state since users currently do not know how much others have delegated to them unless they are already registered transcoders (at that point `transcoderPools.transcoderStake()` can provide their total stake). I'll need to think about this one

contracts/bonding/BondingManager.sol
yondonfu: I think the idea here is to allow users to delegate their bonded tokens to different delegatees without forcing them to first unbond since they will have to wait through the unbonding period before they can re-bond their tokens.

contracts/bonding/BondingManager.sol
yondonfu: One option is to include a `delegatedAmount` field for delegators. Whenever a delegator delegates to a delegatee, the delegatee's `delegatedAmount` field increases by the amount of tokens delegated.

dob:
I think our current approach should be to leave them as `delegators`, and
to encapsulate the code that updates a transcoders bonded stake inside a
method that clearly is named for and comments why the delegator struct is
being updated in order to do this. At first glance I don't think that
renaming `delegators` to `stakeholders` makes much of a difference to the
protocol.

yondonfu:
Ready for another review

- Anyone that participates in the protocol is a delegator and can delegate to any other delegator
- `delegatedAmount` tracks the amount of stake other delegator have delegated to a particular delegator
- A result from the above is that a user can now register as a transcoder with stake already delegated to him
- Fixed failing tests
- Did some refactoring of internal function logic to encapsulate code that updates transcoder stake in clearly named functions

Issue #55, opened 2017-08-14
Transcoder fee vs. broadcaster's max price per segment


yondonfu:
Currently, the price per segment of a job is always the transcoder's fee. As a result, even if the broadcaster's max price per segment is greater than the transcoder fee, the transcoder will still only earn the fee it originally charged.

In this case, the broadcaster benefits because it is always charged the lowest fees, but the transcoder loses out on additional returns if the broadcaster is willing to pay more i.e. (broadcaster's max price per segment > transcoder's fee).

At first glance, I think the current design makes sense because transcoders set their fee based upon what they believe the minimum compensation for their work should be. Thus, as long as the transcoder receives this minimum compensation, it would be happy. Opening this up for future discussion though

dob:
**Pro's for paying MaxPricePerSegment**

* You always pay a predictable fee, and it's not subject to the whim of which transcoder you get assigned to.
* In the broadcaster's mind the tradeoff is availability vs price. This doesn't fit perfectly into our protocol right now because we only assign one transcoder...but when we figure out the way to seamlessly switch between a bunch of candidates, paying a higher price would give you more consistent service.
* If we end up having jobs that contain multiple profiles, then they can be split across transcoders. It's probably simpler mentally for the broadcaster to just have a common price/gas offered that gets paid in total than to have to calculate it separately for every participating transcoder. I'm following the Ethereum model here.
* Similarly, if we ever figured out how to securely let transcoders compete for the job in near real time (by using an offchain scaling solution like Plasma), then I like the Ethereum model where transcoders would want to get better paying jobs before lower paying jobs.

**The Con's Are**

* Higher price is paid than what transcoder was willing to accept. Network becomes more expensive on average for the broadcaster than it needs to be.

yondonfu:
Done. We are going with paying maxPricePerSegment now

PR #75, opened 2017-09-21
A pass at upgradeability

yondonfu:
- `ManagerProxy.sol` contains the logic for forwarding method calls to target contracts such as BondingManager, JobsManager, etc. Target contract methods are invoked using delegatecall thereby allowing the target contract to execute methods and modify state in the context of the proxy contract
- The file `tests/ManagerProxy.js` runs through a simple scenario of setting/getting storage types that are 32 bytes or less and swaps out an old version of a target contract with a new version
- Added `Minter.sol` which owns the LivepeerToken contracts. When bonding or making deposits for jobs, tokens are transferred to this Minter contract. The Minter can mint new tokens if it is called by the BondingManager. The Minter can also transfer tokens in its possession to a recipient if called by the BondingManager or JobsManager
- Register all contracts with the controller so that a user can always get the most up to date version of a contract
- `Minter.sol` and `LivepeerToken.sol` are not proxy target contracts and are not meant to be upgradeable
- Cleaned up the test suite a little bit so we actually have unit tests that rely on mocked contracts. I commented out the integration test for now because I'm still trying to figure out the best strategy for that - ideally we can run our integration tests on any chain/client and not rely solely on TestRPC features (i.e. mine a block per tx, ability to fast forward time/blocks)

I'm going to leave some clarifying comments, but there is a lot in this one so it might be easier to review by reading over it once and then we can sit down and run through it


contracts/ManagerProxy.sol
```diff
@@ -0,0 +1,52 @@
+pragma solidity ^0.4.13;
+
+import "./ManagerProxyTarget.sol";
+
+
+contract ManagerProxy is ManagerProxyTarget {
```
yondonfu: The proxy contract inherits from the ManagerProxyTarget base contract to ensure that it starts off with the same storage layout as the target contract

contracts/bonding/BondingManager.sol
```diff
@@ -127,20 +116,10 @@ contract BondingManager is IBondingManager, Manager {
...
-        uint256 _numActiveTranscoders,
-        uint64 _unbondingPeriod
-    )
-        Manager(_registry)
-    {
-        // Set LivepeerToken address
-        token = LivepeerToken(_token);
+    function BondingManager(address _controller) Manager(_controller) {}
+
+    function initialize(uint64 _unbondingPeriod, uint256 _numActiveTranscoders) external beforeInitialization returns (bool) {
```
yondonfu: When we instantiate a target contract using its constructor we update the storage of the target contract. However, we want to update the storage of the proxy contract. So, the logic of the constructor is now moved into this `initialize` function that the proxy contract can call and execute it its own context to modify the storage of the proxy contract.

contracts/bonding/BondingManager.sol
```diff
@@ -127,20 +116,10 @@ contract BondingManager is IBondingManager, Manager {
...
+        finishInitialization();
```
yondonfu: It's important to call this to set the `initialized` flag (inherited from `Initializable`) to `true` so that the `initialize` method cannot be called more than once

contracts/bonding/BondingManager.sol
```diff
@@ -441,7 +430,7 @@ contract BondingManager is IBondingManager, Manager {
      * Returns address of elected active transcoder and its price per segment
      * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
      */
-    function electActiveTranscoder(uint256 _maxPricePerSegment) external constant returns (address, uint256) {
+    function electActiveTranscoder(uint256 _maxPricePerSegment) external constant returns (address) {
```
yondonfu: I changed this to just return the address of the elected transcoder. Originally it returned the address of the elected transcoder and the transcoder's price. It was returning the price before because the default behavior was to use the transcoder's price which might be lower than the max price per segment set by a broadcaster when calling `job` from the JobsManager.

There were some points made here https://github.com/livepeer/protocol/issues/55 that I thought made sense. Defaulting to the max price per segment set by a broadcaster would mean we do not need to return the transcoder's price here. This also helps with using a proxy contract to return values when using the BondingManager as a target contract - the proxy contract cannot return tuples right now. Since we are now just returning a single address the proxy contract can return that value just fine

contracts/bonding/BondingManager.sol
```diff
@@ -566,12 +555,68 @@ contract BondingManager is IBondingManager, Manager {
...
 
-    /*
-     * @dev Return number of minted tokens for a reward call
-     */
-    function mintedTokensPerReward(address _transcoder) public constant returns (uint256) {
-        uint256 transcoderActiveStake = activeTranscoders[activeTranscoderPositions[_transcoder]].key;
-        return initialTokenSupply.mul(initialYearlyInflation).div(100).div(roundsManager().roundsPerYear()).mul(transcoderActiveStake).div(totalActiveTranscoderStake);
+    // Transcoder getters
+
+    function getTranscoderDelegatorWithdrawRound(address _transcoder) public constant returns (uint256) {
```
yondonfu: We need these getters to individually retrieve values for a transcoder or delegator struct (the proxy contract cannot return tuples at the moment)

contracts/bonding/BondingManager.sol
```diff
@@ -613,6 +658,7 @@ contract BondingManager is IBondingManager, Manager {
      */
     function increaseTranscoderStake(address _transcoder, uint256 _totalAmount, uint256 _transcoderShare, uint256 _round) internal returns (bool) {
         delegators[_transcoder].bondedAmount = delegators[_transcoder].bondedAmount.add(_transcoderShare);
+        delegators[_transcoder].delegatedAmount = delegators[_transcoder].delegatedAmount.add(_totalAmount);
```
yondonfu: This was not updated properly before. When we increase a transcoder's stake we should not only increase the transcoder's stake in the transcoder pools, but also update the total amount delegated to it

contracts/jobs/JobsManager.sol
```diff
@@ -136,21 +128,22 @@ contract JobsManager is IJobsManager, Verifiable, Manager {
...
-        token.transferFrom(msg.sender, this, _amount);
+        // Transfer tokens for deposit to Minter. Sender needs to approve amount first
+        livepeerToken().transferFrom(msg.sender, minter(), _amount);
 
         return true;
     }
 
     /*
      * @dev Withdraw deposited funds
+     * FIXME: Attacker can withdraw funds before transcoder has a chance to claim them
```
yondonfu: Going to need logic to make sure this method is not used to withdraw funds before a transcoder has a chance to claim them. Will track with an issue

contracts/jobs/JobsManager.sol
```diff
@@ -175,11 +169,10 @@ contract JobsManager is IJobsManager, Verifiable, Manager {
         job.streamId = _streamId;
         job.transcodingOptions = _transcodingOptions;
         job.maxPricePerSegment = _maxPricePerSegment;
-        job.pricePerSegment = pricePerSegment;
         job.broadcasterAddress = msg.sender;
         job.transcoderAddress = electedTranscoder;
 
-        NewJob(electedTranscoder, msg.sender, numJobs);
+        NewJob(electedTranscoder, msg.sender, numJobs, _streamId, _transcodingOptions);
```
yondonfu: We currently cannot return streamId or transcodingOptions when using a proxy contract because the delegatecall mechanism cannot return dynamic length types (i.e. strings). Instead we can include these fields in events and a node could retrieve the values from the event

contracts/ManagerProxy.sol
```diff
@@ -0,0 +1,52 @@
...
+    function ManagerProxy(address _controller, bytes32 _targetContractId) Manager(_controller) {
+        targetContractId = _targetContractId;
+    }
+
+    // Based on https://github.com/AugurProject/augur-core/blob/develop/src/libraries/Delegator.sol
+    function() public payable {
```
ericxtang: Dose this make every call payable?

contracts/bonding/BondingManager.sol
yondonfu: Not sure, I'll make a note to test this out. Since constant getters are using the eth_call RPC request for a node maybe we can batch together the RPC requests to make it faster

contracts/ManagerProxy.sol
yondonfu: Discussed offline and the fallback function needs to be payable for the case when a user is calling the `verify()` function in JobsManager which is payable. If we really wanted to prevent anyone from accidentally sending ether to this contract except for when `verify()` is called, we could add a check for the method signature and if it is not the signature for `verify()` and `msg.value > 0` we throw

contracts/bonding/BondingManager.sol
```diff
@@ -363,10 +352,8 @@ contract BondingManager is IBondingManager, Manager {
         // Set last round that transcoder called reward
         transcoders[msg.sender].lastRewardRound = currentRound;
 
-        // Calculate number of tokens to mint
-        uint256 mintedTokens = mintedTokensPerReward(msg.sender);
-        // Mint token reward and allocate to this protocol contract
-        token.mint(this, mintedTokens);
+        // Mint token reward
+        uint256 mintedTokens = minter().mint(activeTranscoders[activeTranscoderPositions[msg.sender]].key, totalActiveTranscoderStake);
```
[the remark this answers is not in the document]
yondonfu: With the addition of the Minter contract I thought it might make sense to move the logic for calculating how many new tokens to mint into the Minter contract. Now, the BondingManager tells the Minter what the current transcoder's stake and the total active transcoder stake is. The Minter then uses that information to decide how many tokens to mint.

yondonfu:
Agreed on how to look at the upgradeability related additions in this PR. I think it makes sense to have a working version of this approach in this codebase for now and press on with other tasks knowing that we have _a_ basic upgradeability solution to work with. And perhaps we will get more information in time (i.e. feedback, Byzantium on the mainnet, Solidity development) that changes how we look at this solution (for better or for worse)

PR #79, opened 2017-09-28
Verifier interface changes + LivepeerVerifier

yondonfu:
- Added the `dataStorageHash` parameter to the `verify` function. This parameter should be content addressable storage hash for segment data used during verification (in the case of Oraclize this would be the IPFS hash). Previously, we were treating `dataHash` as the storage hash. Now, `dataHash` is just the keccak256 hash of the segment data.
- `dataHash` and `transcodedDataHash` are now both bytes32 instead of strings
- The OraclizeVerifier now uses a binary helper (http://docs.oraclize.it/#general-concepts-parsing-helpers) that unhexlifies the result of running a Docker application before submitting the result on-chain in a callback
- Added a `getPrice` function to the Verifier interface. The JobsManager checks if a transcoder can pay for the price of verification and forwards the payment to the Verifier contract
- Added a LivepeerVerifier contract that replicates a lot of the logic in OraclizeVerifier, but we can run an application off-chain to watch for request events, run verification computations ourselves and then submit the results back on-chain
- Using the MerkleProof library that is now included in OpenZeppelin


contracts/jobs/JobsManager.sol
```diff
@@ -281,19 +287,51 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
...
+    {
+        IVerifier verifierContract = verifier();
+
+        uint256 price = verifierContract.getPrice();
+
+        // Send payment to verifier if price is greater than zero
+        if (price > 0) {
+            return verifierContract.verify(_jobId, _claimId, _segmentNumber, jobs[_jobId].transcodingOptions, _dataStorageHash, _transcodedDataHash);
+        } else {
+            return verifierContract.verify.value(price)(_jobId, _claimId, _segmentNumber, jobs[_jobId].transcodingOptions, _dataStorageHash, _transcodedDataHash);
```
yondonfu: Since this function is only called by `verify`, which will ensure that a transcoder provided enough payment to cover the price of verification, we can forward that payment to the verifier

contracts/verification/OraclizeVerifier.sol
```diff
@@ -4,22 +4,26 @@ import "../Manager.sol";
...
 
 
 /*
  * @title Verifier contract that uses Oraclize for off-chain computation
  */
 contract OraclizeVerifier is Manager, usingOraclize, IVerifier {
+    using SafeMath for uint256;
+
     string public verificationCodeHash;
+    uint256 public gasPrice;
```
yondonfu: Since the price of using Oraclize includes the price of the callback, we can set the gas price and gas limit to be used for the callback. Oraclize does not provide refunds if the gas provided is not all used, so it is important to accurately estimate the cost of the callback...We can probably do this and then set the gasLimit and gasPrice to be used in the constructor, but perhaps we want to introduce setters to change the values used over time - less so for gasLimit, more so for gasPrice.

ericxtang:
Did in-person code review. 🚢

PR #82, opened 2017-09-29
Verifier checks commit hash defined as h(dataHash,transcodedDataHash)

yondonfu:
Previously, a verifier contract did not check if the dataHash submitted by a transcoder matches the data located at the dataStorageHash submitted. Now the off-chain computation (Oraclize, an off-chain solver, etc.) will download `data` using the dataStorageHash, and submit back on-chain `keccak256(keccak256(data),transcodedDataHash)`. The verifier contract will compare this result with `keccak256(dataHash,transcodedDataHash)` where `dataHash` and `transcodedDataHash` are submitted by the transcoder, which is defined as the commit hash. If the transcoder provides a `dataStorageHash` that does not contain the data that matches `dataHash`, then this check will fail.

### DIFF
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -253,8 +253,8 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
      * @dev Submit transcode receipt and invoke transcoding verification
      * @param _jobId Job identifier
      * @param _segmentNumber Segment sequence number in stream
-     * @param _dataHash Content-addressed storage hash of segment data
-     * @param _transcodedDataHash Hash of transcoded segment data
+     * @param _dataStorageHash Content-addressed storage hash of segment data
+     * @param _dataHashes Hash of segment data and hash of transcoded segment data
      * @param _broadcasterSig Broadcaster's signature over segment hash
      * @param _proof Merkle proof for transcode receipt
      */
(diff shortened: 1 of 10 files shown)


contracts/jobs/JobsManager.sol
```diff
@@ -263,8 +263,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         uint256 _claimId,
         uint256 _segmentNumber,
         string _dataStorageHash,
-        bytes32 _dataHash,
-        bytes32 _transcodedDataHash,
+        bytes32[2] _dataHashes,
```
yondonfu: Grouped `dataHash` and `transcodeDataHash` into a single fixed array parameter to get around stack depth too deep issues...

contracts/jobs/JobsManager.sol
```diff
@@ -325,9 +324,9 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
 
         // Send payment to verifier if price is greater than zero
         if (price > 0) {
-            return verifierContract.verify.value(price)(_jobId, _claimId, _segmentNumber, jobs[_jobId].transcodingOptions, _dataStorageHash, _transcodedDataHash);
+            return verifierContract.verify.value(price)(_jobId, _claimId, _segmentNumber, jobs[_jobId].transcodingOptions, _dataStorageHash, _dataHashes);
```
dob: What is this `value(price)` syntax? I assume it just passes the value into the receiving function?

yondonfu: This forwards funds to the verifier contract to be used for verification. `verify` will check that the caller provides enough funds to forward

yondonfu:
Callback invocation does not necessarily have to reference the previously called verification since `receiveVerification` in the JobsManager accepts a jobId and claimId which it uses to refund broadcasters and slash transcoders for the specific jobId and claimId

PR #84, opened 2017-10-02
Calculate fees based on number of video profiles

### LINKED ISSUE #78: Broadcaster pays per rendition
A transcodingOptions string might define multiple renditions and a broadcaster should pay more for a job that has more renditions since a transcoder is currently responsible for all of those renditions. The fees associated with a transcoder claim becomes `# of segments claimed * # of renditions * pricePerSegment`

yondonfu:
- Video profile identifiers calculated as `bytes4(keccak256(PROFILE_NAME))`
- We keep `transcodingOptions` as a string so the identifiers can be easily parsed by the Docker application that Oraclize passes the string to as a parameter
- Since we keep `transcodingOptions` as a utf8 encoded string, we actually use 8 bytes to store the utf8 encoded representation of the first 4 bytes of the keccak hash of a video profile name (4 bytes  * 2 characters per byte)

Fixes #78

### DIFF
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -223,7 +223,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         require(_segmentRange[1] >= _segmentRange[0]);
 
         // Move fees from broadcaster deposit to escrow
-        uint256 fees = _segmentRange[1].sub(_segmentRange[0]).add(1).mul(job.maxPricePerSegment);
+        uint256 fees = JobLib.calcFees(_segmentRange[1].sub(_segmentRange[0]).add(1), job.transcodingOptions, job.maxPricePerSegment);
         broadcasterDeposits[job.broadcasterAddress] = broadcasterDeposits[job.broadcasterAddress].sub(fees);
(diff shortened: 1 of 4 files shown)


contracts/jobs/JobsManager.sol
```diff
@@ -432,7 +432,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         // Slashing period must be over for claim
         require(claim.endSlashingBlock < block.number);
 
-        uint256 fees = claim.segmentRange[1].sub(claim.segmentRange[0]).add(1).mul(job.maxPricePerSegment);
+        uint256 fees = JobLib.calcFees(claim.segmentRange[1].sub(claim.segmentRange[0]).add(1), job.transcodingOptions, job.maxPricePerSegment);
```
dob: Do you think `calcFees` should be invoked in the `job()` transaction and just saved on the job itself, rather than calculated in three separate places?

dob: essentially calculating fee/segment for the job? On the other hand I could see why this wouldn't work because you don't know which renditions will end up being claimed, and only if the claim contains all renditions would the pre-calculated fees make sense.

PR #86, opened 2017-10-12
Unbonded fee payouts + delayed computation of transcoder bonded/unbonded amounts

yondonfu:
This is an implementation of a solution to our current gas limit vulnerabilities in looping through rounds to update a delegator's stake and looping through claims within a round to update a delegator's stake with fees that the delegator is eligible for.

This PR includes:
- Fees are paid out unbonded. Delegators claim their fee shares from fee pools during the rounds since their lastClaimTokenPoolsShareRound
- Transcoders claim their blockRewardCut and feeShare when they claim their reward and fee shares from token pools during the rounds since their lastClaimTokenPoolsShareRound as a delegator
- Transcoders calculate unclaimable fees and rewards when adding rewards and fees to token pools. The unclaimable fees and rewards go to a redistribution pool managed by the Minter
- When initializeRound is called in RoundsManager it sets the mintable tokens and redistributable tokens for a round in the Minter

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.4.13;
 import "../ManagerProxyTarget.sol";
 import "./IBondingManager.sol";
 import "./libraries/TranscoderPools.sol";
+import "./libraries/TokenPools.sol";
 import "../token/ILivepeerToken.sol";
 import "../token/IMinter.sol";
 import "../rounds/IRoundsManager.sol";
@@ -14,6 +15,7 @@ import "zeppelin-solidity/contracts/math/SafeMath.sol";
 contract BondingManager is ManagerProxyTarget, IBondingManager {
     using SafeMath for uint256;
     using TranscoderPools for TranscoderPools.TranscoderPools;
+    using TokenPools for TokenPools.Data;
 
     // Time between unbonding and possible withdrawl in rounds
     uint64 public unbondingPeriod;
@@ -28,19 +30,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         uint8 pendingBlockRewardCut;                         // Pending block reward cut for next round if the transcoder is active
         uint8 pendingFeeShare;                               // Pending fee share for next round if the transcoder is active
         uint256 pendingPricePerSegment;                      // Pending price per segment for next round if the transcoder is active
-        mapping (uint256 => TokenPools) tokenPoolsPerRound;  // Mapping of round => token pools for the round
+        mapping (uint256 => TokenPools.Data) tokenPoolsPerRound;  // Mapping of round => token pools for the round
     }
 
     // The various states a transcoder can be in
     enum TranscoderStatus { NotRegistered, Registered, Resigned }
-
-    // Represents rewards and fees to be distributed to delegators
-    struct TokenPools {
-        uint256 rewardPool;      // Reward tokens in the pool
-        uint256 feePool;         // Fee tokens in the pool. stakeRemaining / totalStake = % of claimable fees in the pool
-        uint256 totalStake;      // Transcoder's total stake during the pool's round
-        uint256 stakeRemaining;  // Stake that has not been used to claim fees from the fee pool
-    }
 
     // Represents a delegator's current state
     struct Delegator {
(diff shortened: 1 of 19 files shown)


contracts/token/Minter.sol
```diff
@@ -34,15 +46,25 @@ contract Minter is Manager, IMinter {
...
+        // Compute fraction of redistributable tokens to include in reward
+        uint256 redistributeAmount = currentRedistributableTokens.mul(_fracNum).div(_fracDenom);
+        // Update amount of redistributable tokens for round
+        currentRedistributableTokens = currentRedistributableTokens.sub(redistributeAmount);
+        redistributionPool = redistributionPool.sub(redistributeAmount);
+
+        // Compute and mint fraction of mintable tokens to include in reward
+        uint256 mintAmount = currentMintableTokens.mul(_fracNum).div(_fracDenom);
+        // Update amount of mintable tokens for round
+        currentMintableTokens = currentMintableTokens.sub(mintAmount);
```
dob: if `currentMintableTokens` gets reduced here after we've minted some tokens to one caller of reward, then does that screw up calculations for the next caller? I'm guessing the num and denom are the callers total stake and the total overall stake?

For example, if there are 100 mintable tokens, and two people each have 10% of stake, then you'd want them each to get 10 tokens. By this logic would the second caller get 10% of 90 remaining `currentMintableTokens` and only end up with 9 tokens? It's possible I'm following it wrong.

contracts/bonding/BondingManager.sol
```diff
@@ -28,40 +30,21 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
     struct Delegator {
-        uint256 bondedAmount;              // The amount of bonded tokens
-        address delegateAddress;           // The address delegated to
-        uint256 delegatedAmount;           // The amount of tokens delegated to the delegator
-        uint256 startRound;                // The round the delegator transitions to bonded phase
-        uint256 delegateBlock;             // The block the delegator bonds to somone
-        uint256 withdrawRound;             // The round at which a delegator can withdraw
-        uint256 lastStakeUpdateRound;      // The last round the delegator transitioned states
+        uint256 bondedAmount;                    // The amount of bonded tokens
+        uint256 unbondedAmount;                  // The amount of unbonded tokens
```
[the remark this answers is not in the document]
yondonfu: Yep. Although I think once we allow for partial unbonding, we would move away from a delegator being in a bonded or unbonded state - rather a delegator converts a portion of bondedAmount to unbondedAmount when it calls `unbond(amount)`

contracts/bonding/BondingManager.sol
```diff
@@ -650,149 +746,73 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-                claimBlock: _claimBlock,
-                fees: delegatorsFeeShare,
-                transcoderTotalStake: _transcoderTotalStake
-            })
-        );
+    function updateDelegatorWithTokenPoolsShares(address _delegator, uint256 _endRound) internal returns (bool) {
+        Delegator storage del = delegators[_delegator];
 
-        increaseTranscoderStake(_transcoder, _fees, _fees.sub(delegatorsFeeShare), _round);
+        if (delegatorStatus(_delegator) == DelegatorStatus.Bonded && transcoderStatus(del.delegateAddress) == TranscoderStatus.Registered) {
```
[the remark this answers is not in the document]
yondonfu: Yeah a transcoder is still registered even if it is not in the active transcoder set. I think we need to make sure a delegator can still collect its token pool shares if a transcoder is slashed though...

contracts/token/Minter.sol
yondonfu: Ah good catch...will need to fix this

contracts/bonding/BondingManager.sol
```diff
@@ -352,10 +341,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Set last round that transcoder called reward
         transcoders[msg.sender].lastRewardRound = currentRound;
 
-        // Mint token reward
-        uint256 mintedTokens = minter().mint(activeTranscoders[activeTranscoderPositions[msg.sender]].key, totalActiveTranscoderStake);
+        // Create reward
+        uint256 rewardTokens = minter().createReward(activeTranscoders[activeTranscoderPositions[msg.sender]].key, totalActiveTranscoderStake);
```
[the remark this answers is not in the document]
yondonfu: I'm going to add a clarifying comment for now, but I agree that this `.key` syntax to access the activeTranscoders data structure is confusing and worth changing to make the code clearer (perhaps wrapping the data structures related to the active transcoder set into their own struct with custom getter functions) in a separate PR

contracts/bonding/BondingManager.sol
```diff
@@ -285,14 +265,30 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     /**
      * @dev Withdraws withdrawable funds back to the caller after unbonding period.
      */
-    function withdraw() external afterInitialization whenSystemNotPaused currentRoundInitialized returns (bool) {
-        // Delegator must be unbonded
-        require(delegatorStatus(msg.sender) == DelegatorStatus.Unbonded);
+    function withdraw()
```
ericxtang: So now we are overloading the withdraw() call with 2 meanings right?  If the delegator is still bonded, we just withdraw the unblondedAmount.  If it IS unbonded, we withdraw the bonded amount.

Is there a possibility where we have an unbonded delegator with both bonded AND unbonded amounts?

yondonfu: Yeah...so the delegator would have to call withdraw twice: once to empty out the unbondedAmount and a second time to empty out the bondedAmount. A temporary better way could be: 

- If the delegator does not have an unbonded amount OR it is not unbonded
- If the delegator has an unbonded amount, empty it
- If the delegator is also unbonded, empty its bonded amount as well

ericxtang: That makes sense.

ericxtang:
A lot of details here, I have a feeling I'm not capturing the whole picture, but in the spirit of time, let's 🚢 .  We'll need to reduce the complexity in a future iteration.

dob:
Before the ship, I think there was one unaddressed issue about correctly calculating the payouts. Was that fixed?

yondonfu:
@dob Updated `createReward` in `Minter.sol` to always calculate the minted tokens and redistributed tokens based on the fixed values `currentMintableTokens` and `currentRedistributableTokens` for the round

dob:
Yep, see that now. 👍 🚢

PR #96, opened 2017-10-17
Dockerized Geth and Parity

### LINKED ISSUE #90: Dockerized Geth and Parity
Enable running integration tests using a dockerized Geth PoA node with fast block times and a dockerized Parity running on an instant seal dev chain. This will be especially helpful for testing new Byzantium features since TestRPC does not yet support Byzantium changes

yondonfu:
Docker files for building Geth and Parity docker images. The Geth image sets up a PoA node with 1 second block times. The Parity image sets up a instant seal node. Both nodes are set up with 4 accounts preallocated with ether.

Closes #90


scripts/geth-dev/start.sh
```diff
@@ -0,0 +1,3 @@
+geth --datadir /geth/.lpTest init /geth/genesis.json
```
dob: It wouldn't be crazy to move to a convention of `.livepeer/` as the default datadir and `.livepeer/testnet` as the default testnet data dir. The `.lpTest` feels arbitrary and like at some point we should move to something more professional.

yondonfu:
`scripts/parseArtifacts.js` is used to parse the contract artifacts and create `.abi` and `.bin` files which are used for generating Go contract bindings. `geth-dev` and `parity-dev` contain everything we need to set up Geth and Parity nodes that we can use for integration testing. I think these will be useful for testing out the RETURNDATACOPY and RETURNDATASIZE opcodes in Byzantium since TestRPC is not Byzantium compatible yet, but also for ensuring integration test scenarios run properly across the major client implementations and not just TestRPC. As a part of CI, we could run our unit tests with TestRPC and our integration tests with TestRPC, Parity and Geth.

PR #100, opened 2017-10-24
Pure/view & solc v0.4.17 & truffle v4.0.0-beta

### LINKED ISSUE #89: Pure and view functions
Use the `pure` and `view` function declarations in favor of `constant`

yondonfu:
Switching from constant to pure/view function declarations and to the newest version of truffle that uses a version of Solidity that supports pure/view.

Closes #89

### DIFF
--- a/contracts/jobs/libraries/JobLib.sol
+++ b/contracts/jobs/libraries/JobLib.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.13;
+pragma solidity ^0.4.17;
 
 import "zeppelin-solidity/contracts/ECRecovery.sol";
 import "zeppelin-solidity/contracts/MerkleProof.sol";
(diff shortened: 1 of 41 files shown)


contracts/jobs/libraries/JobLib.sol
```diff
@@ -104,7 +104,7 @@ library JobLib {
      * @param _segmentSequenceNumber Segment sequence number in stream
      * @param _dataHash Content-addrssed storage hash of segment data
      */
-    function personalSegmentHash(string _streamId, uint256 _segmentNumber, bytes32 _dataHash) public constant returns (bytes32) {
+    function personalSegmentHash(string _streamId, uint256 _segmentNumber, bytes32 _dataHash) public pure returns (bytes32) {
```
dob: What makes these two functions candidates for `pure` instead of `view`? This is just a learning moment for those following along.

yondonfu: `pure` functions do not read from or write to contract state. `view` functions can read from, but cannot write to contract state. Right now these declarations just allow the compiler to check if you are accidentally violating the associated restrictions for a `pure`/`view` declaration, but in the future the compiled bytecode for `view` functions might use the new STATICCALL opcode which actually enforces read-only calls during runtime on the EVM level

PR #101, opened 2017-10-28
ManagerProxy returns dynamic length data

### LINKED ISSUE #88: RETURNDATACOPY and RETURNDATASIZE
Use the new RETURNDATACOPY and RETURNDATASIZE opcodes in `ManagerProxy.sol` so proxy contracts can return dynamic data. Once we make this change, we can also reduce the number of getter functions that we define in our contracts - instead of each getter returning a specific field of a struct, we can use a single getter to return a tuple of all the fields of a struct. This will reduce the size/cost of deploying the contracts

yondonfu:
ManagerProxy now uses RETURNDATACOPY and RETURNDATASIZE opcodes to return dynamic length data from a target contract. As a result, we can use getter functions in BondingManager and JobsManager to return tuples rather than define a getter function for each individual field for a transcoder, delegator, job, claim, etc. We can also now directly retrieve streamId and transcodingOptions values directly from the JobsManager without relying on events. Using ethereumjs-testrpc v5.0.1@beta which includes the new opcodes. 

Closes #88

### DIFF
--- a/contracts/ManagerProxy.sol
+++ b/contracts/ManagerProxy.sol
@@ -4,16 +4,12 @@ import "./ManagerProxyTarget.sol";
 
 
 contract ManagerProxy is ManagerProxyTarget {
-    function ManagerProxy(address _controller, bytes32 _targetContractId) Manager(_controller) {
+    function ManagerProxy(address _controller, bytes32 _targetContractId) public Manager(_controller) {
         targetContractId = _targetContractId;
     }
 
     // Based on https://github.com/AugurProject/augur-core/blob/develop/src/libraries/Delegator.sol
     function() public payable {
-        // Set size of method call result to 32 for now until
-        // we can use the RETURNDATACOPY and RETURNDATASIZE opcodes introduced
-        // in the Byzantium hard fork
-        uint32 size = 32;
         address target = controller.getContract(targetContractId);
         // Target contract must be registered
         require(target > 0);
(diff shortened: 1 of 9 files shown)


yondonfu:
1. Yep, can definitely do an in-person walkthrough

2. The functions that return tuples i.e. `getTranscoder`, `getDelegator`, `getJob ` and `getClaim` just return all the fields for their associated struct except for fields that are nested data structures - there are separate getters for the nested data structure fields i.e. `getTranscoderTokenPoolsPerRound`. But I do agree that the long tuple could be better documented. I think named returns are not great for readability in more complex functions, but since these getters would just be setting values for the named return parameters that might be ok

contracts/ManagerProxy.sol
```diff
@@ -4,48 +4,42 @@ import "./ManagerProxyTarget.sol";
...
             calldatacopy(calldataMemoryOffset, 0x0, calldatasize)
+
             // Call method on target contract and store result starting at calldataMemoryOffset
-            let ret := delegatecall(gas, target, calldataMemoryOffset, calldatasize, calldataMemoryOffset, size)
+            let ret := delegatecall(gas, target, calldataMemoryOffset, calldatasize, 0, 0)
+
+            // Load the free memory pointer at 0x40
+            let returndataOffset := mload(0x40)
+            // Update free memory pointer to after memory space we reserve for returndata
+            mstore(0x40, add(returndataOffset, returndatasize))
```
ericxtang: Where is returndatasize declared?

contracts/ManagerProxy.sol
```diff
@@ -4,48 +4,42 @@ import "./ManagerProxyTarget.sol";
 
 
 contract ManagerProxy is ManagerProxyTarget {
-    function ManagerProxy(address _controller, bytes32 _targetContractId) Manager(_controller) {
+    function ManagerProxy(address _controller, bytes32 _targetContractId) public Manager(_controller) {
```
ericxtang: This means Manager can be called by anyone right?

contracts/ManagerProxy.sol
```diff
@@ -4,48 +4,42 @@ import "./ManagerProxyTarget.sol";
...
+
+            // Load the free memory pointer at 0x40
+            let returndataOffset := mload(0x40)
+            // Update free memory pointer to after memory space we reserve for returndata
+            mstore(0x40, add(returndataOffset, returndatasize))
+            // Copy returndata to memory
+            returndatacopy(returndataOffset, 0, returndatasize)
+
             switch ret
             case 0 {
```
ericxtang: Do we need this switch if the 2 cases are the same?

contracts/ManagerProxy.sol
yondonfu: Yeah, but its the constructor so it is only ever called once during contract creation. The constructor is `public` by default anyway - the explicit `public` is added to get rid of the compiler warning

contracts/ManagerProxy.sol
yondonfu: `returndatasize` is the opcode that returns the size of the return data buffer allocated in the stack frame after the `delegatecall`

contracts/ManagerProxy.sol
yondonfu: We're using the `revert` opcode in the first case and the `return` opcode in the second case. Once Solidity supports error messages when using the `require` guard, `revert(returndataOffset, returndatasize)` will cause the function call to fail and return the error message retrieved from the `delegatecall`.

contracts/ManagerProxy.sol
```diff
@@ -4,48 +4,42 @@ import "./ManagerProxyTarget.sol";
...
-            }
-            // Update free memory pointer to after memory space we reserve for calldata and method call results
-            mstore(0x40, add(calldataMemoryOffset, reservedSize))
-
+            // Update free memory pointer to after memory space we reserve for calldata
+            mstore(0x40, add(calldataMemoryOffset, calldatasize))
             // Copy method signature and params of the call to memory
             calldatacopy(calldataMemoryOffset, 0x0, calldatasize)
+
             // Call method on target contract and store result starting at calldataMemoryOffset
```
dob: This comment should be updated as it's no longer accurate

contracts/ManagerProxy.sol
```diff
@@ -4,48 +4,42 @@ import "./ManagerProxyTarget.sol";
...
-        // Set size of method call result to 32 for now until
-        // we can use the RETURNDATACOPY and RETURNDATASIZE opcodes introduced
-        // in the Byzantium hard fork
-        uint32 size = 32;
         address target = controller.getContract(targetContractId);
         // Target contract must be registered
         require(target > 0);
 
         assembly {
             // Load the free memory pointer at 0x40
```
dob: Let's consider adding a variable called `freeMemoryPointer` which equals `0x40` initially, since we keep using it and updating the value stored there. Then we could replace all the `0x40`'s throughout the code.

contracts/ManagerProxy.sol
dob: two nits: use consistent naming for callDataMemoryOffset and returndataOffset, and in one call we use `0x0` and the other just `0`.

yondonfu:
Reviewed the ManagerProxy contract offline and updated it with a few naming/comment fixes.

Issue #103, opened 2017-11-06
Randomness attack vectors


yondonfu:
Currently, when a transcoder calls `claimWork`, the current block hash is used to determine which segments are challenged for verification such that:

`segNo` is challenged for verification if `h(blockNumber, blockHash, segNo) % verificationRate` == 0.

The idea here is that a transcoder will not know what the block hash is when it calls `claimWork` so it will not be able to determine which segments are challenged for verification until it actually calls `claimWork`. However, this property does not hold in the case of a transcoder also being an Ethereum miner because the miner can control the block hash at the time of a call to `claimWork`.

A miner/transcoder could send garbage data to the broadcaster as a part of its transcoded stream, while holding onto all of the broadcasted segment data. When the miner/transcoder calls `claimWork`, it can generate blocks until it finds a block hash that guarantees that certain segments are challenged which can be known ahead of time. The miner also has a bit of flexibility because it can choose when to call `claimWork` and only call it when it has mined a block with a favorable block hash. If the miner knows which segments are challenged, it can selectively transcode the segments that will be challenged. The rest of the transcode receipts will contain the hash of garbage data for the `transcodedDataHash` field. As a result, the transcoder will be able to pass verification for the challenged segments since they were transcoded properly, and will be able to get away with submitting garbage data for non-challenged segments.

The cost of the attack is the lost block reward - the miner runs the chance of losing block reward if it keeps generating new blocks in search of a favorable block hash and an honest miner submits its mined block before the dishonest miner. This attack is only worth it for the miner/transcoder if the value of the LPT fees that can be distributed from a `claimWork` call is > `block reward * # of tries to get a favorable block hash`. This would likely only potentially be the case for very long lived streams with many segments.

This can be mitigated by broadcaster challenges. A transcoder sends signed transcoded segments to the broadcaster off-chain so a broadcaster can observe that the transcoded data it received is different from the transcoded data submitted as a part of a transcode receipt when a transcoder calls `verify` for a challenged segment. Since the transcoded segment is signed, the broadcaster can prove to the contract that the transcoder sent it garbage data, but submitted valid transcoded data for verification. A successful proof could slash the transcoder.

A transcoder can send garbage data to the broadcaster, but it cannot selectively send the broadcaster valid transcoded segments for challenged segments and garbage data for un-challenged segments, because it would've already sent the segments off-chain by the time it can determine the challenged segments when it calls `claimWork`.

dob:
[...]

Issue #46, opened 2017-07-27
Think about the incentives of calling EndJob


yondonfu:
The incentives for calling EndJob() are currently unclear. Who would call it, the transcoder or the broadcaster and would they be incentivized to call it such that the benefits outweigh the costs of an on-chain transaction?

dob:
How about when the broadcaster requests the job they specify how many blocks they expect transcoded commitment for (default 2 hrs or something), and they pay for longer commitments? Would this make endjob unnecessary on chain and it could move to just being an off chain convenience? Everyone would know the job ended after the expiration block.

________________________________

yondonfu:
Additional thoughts:

Currently `endJob` sets an end block such that a transcoder can still claim work and verify challenged segments in the time period before the job's end block. This is to prevent a broadcaster from receiving transcoded segments from the transcoder and calling `endJob` before a transcoder can claim work and verify challenged segments in an effort to avoid paying the fees associated with a transcoded stream.

As a type of griefing attack, a transcoder transcode all the segments, but withhold the segments from the broadcaster. The broadcaster has no way to prove the transcoder is withholding segments, so the only course of action available is to call `endJob`. However, since `endJob` only sets an end block for the job, a transcoder is still able to claim work and verify challenged segments during the time period until the end block, thereby allowing the transcoder to take fees from the broadcaster's deposit even though the transcoder is withholding segments from the broadcaster. A job's ending period could be adjusted such that a transcoder does not have enough time to go through a full claim work / verify flow once `endJob` is called, but then this would again allow a broadcaster to call `endJob` to prevent a transcoder from being able to get the fees it is owed when the transcoder has already sent all the segments to the broadcaster.

yondonfu:
`endJob` is currently removed since jobs are assigned an end block upon creation so this can be closed

PR #108, opened 2017-11-08
Restrictions on broadcaster withdrawals

### LINKED ISSUE #76: Restrictions on withdrawing job deposits
Currently, a broadcaster can withdraw deposits at any time using the `withdraw` method. As is, this method is a vulnerability because a broadcaster could call `withdraw` after a transcoder has done work, but before it actually claims work on-chain (which would transfer funds from the broadcaster's deposit to job escrow). Tracking this here as a reminder to fix this

yondonfu:
Current approach to restricting broadcaster deposit withdrawals:

The JobsManager keeps track of the number of active jobs for a broadcaster. A job can be in 3 states - `Active`, `Ended` and `Inactive`. When a job is created it becomes `Active`. When either a transcoder or broadcaster calls `endJob(jobId)` for a job, the job becomes `Ended` and an end block is set for the job. After a job's end block, it becomes `Inactive`. A broadcaster can only withdraw its deposit if it has 0 active jobs, and the latest end block for any of its ended jobs is less than or equal to the current block. A consequence of this is that a broadcaster must call `endJob` for all of its active jobs in order to withdraw its deposit which is not ideal from a UX/cost point of view, but let me know what you guys think.

As a solution to mitigate the possibility of a miner influencing the block hash used to determine which segments are challenged, we use the block hash of claimBlock + 1 which a miner can only influence if it wins the race to mine the block after the claimBlock.

Closes #76 #103

### DIFF
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -449,10 +449,10 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
      */
     function jobStatus(uint256 _jobId) public view returns (JobStatus) {
         if (jobs[_jobId].endBlock <= block.number) {
-            // A job is inactive if its end block is set and the current block is greater than or equal to the job's end block
+            // A job is inactive if the current block is greater than or equal to the job's end block
             return JobStatus.Inactive;
         } else {
-            // A job is active if it does not have an end block
+            // A job is active if the current block is less than the job's end block
             return JobStatus.Active;
         }
     }
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -22,9 +22,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
     // % of verifications you can fail before being slashed
     uint64 public verificationFailureThreshold;
 
-    // Time between when endJob() is called for a job and when the job is considered inactive. Denominated in blocks
-    uint256 public jobEndingPeriod;
-
     // Time after a transcoder calls claimWork() that it has to complete verification of claimed work
     uint256 public verificationPeriod;
 
@@ -42,7 +39,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
 
     struct Broadcaster {
         uint256 deposit;         // Deposited tokens for jobs
-        uint256 activeJobs;      // Active jobs that have not been ended
         uint256 withdrawBlock;   // Block at which a deposit can be withdrawn
     }
 
@@ -64,7 +60,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
     }
 
     // States of a job
-    enum JobStatus { Inactive, Ended, Active }
+    enum JobStatus { Inactive, Active }
 
     // Represents a transcode claim
     struct Claim {
@@ -113,7 +109,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
 
     function initialize(
         uint64 _verificationRate,
-        uint256 _jobEndingPeriod,
         uint256 _verificationPeriod,
         uint256 _slashingPeriod,
         uint64 _failedVerificationSlashAmount,
@@ -127,7 +122,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         finishInitialization();
 
         verificationRate = _verificationRate;
-        jobEndingPeriod = _jobEndingPeriod;
         verificationPeriod = _verificationPeriod;
         slashingPeriod = _slashingPeriod;
         failedVerificationSlashAmount = _failedVerificationSlashAmount;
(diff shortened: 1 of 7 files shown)


contracts/jobs/JobsManager.sol
```diff
@@ -198,8 +210,19 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         // Sender must be the job's broadcaster or elected transcoder
         require(job.broadcasterAddress == msg.sender || job.transcoderAddress == msg.sender);
 
+        // Decrement broadcaster's active jobs
+        broadcasters[msg.sender].activeJobs = broadcasters[msg.sender].activeJobs.sub(1);
+
+        uint256 endBlock = block.number.add(jobEndingPeriod);
+
         // Set end block for job
         job.endBlock = block.number.add(jobEndingPeriod);
```
dob: Should we use the variable `endBlock` in this assignment since we just calculated it one line above?

contracts/jobs/libraries/JobLib.sol
```diff
@@ -51,13 +49,23 @@ library JobLib {
             return false;
         }
 
-        if (uint256(keccak256(_claimBlock, block.blockhash(_claimBlock), _segmentNumber)) % _verificationRate == 0) {
+        // Use block hash and block number of the block after a claim to determine if a segment
+        // should be verified
+        if (uint256(keccak256(_claimBlock + 1, block.blockhash(_claimBlock + 1), _segmentNumber)) % _verificationRate == 0) {
```
dob: Is the fact that `block.blockhash` is only available for the most recent 256 blocks a problem here? Does it mean there's only a ~64 minute window for verification and slashing to take place before this code wouldn't work?

yondonfu:
An alternative might be for there to either be a manually set or default end block for a job and after that block the job is considered inactive. When a broadcaster creates a job, it updates its own withdraw block to whatever is later, its current withdraw block or the end block of the new job. A broadcaster could optionally extend the length of a job via a function like `extendJob(uint256 _jobId, uint256 _blocks)` which updates the end block of the job. `withdraw` will just check if the broadcaster's withdraw block is less than or equal to the current block thereby avoiding the need for a broadcaster to call `endJob` for all of its active jobs. This would also mean jobs only need to have 2 states again `Active` or `Inactive`.

However, not directly related to withdrawals, I think there might still be potential issues with attacks related to the lifecycle of a job that I note here: https://github.com/livepeer/protocol/issues/46#issuecomment-342950346

contracts/jobs/libraries/JobLib.sol
yondonfu: Right...I forgot about that restriction. We probably cannot calculate the block hash in `shouldVerifySegment` and instead have to store the block hash at some point so that it can be used later

dob: Ok, need to think about this more...since it seems like this limitation is a problem for us potentially.

yondonfu:
After discussing offline, the interim solution we are thinking of rolling with right now is as follows:

When a broadcaster creates a job, it can pass in an end block at which a job becomes inactive. Clients interacting with the contracts can set a default if a broadcaster does not manually set one (maybe 24 hours, maybe roundLength). As mentioned in the previous comment, the end block for a job also becomes the broadcaster's withdraw block if the end block > the broadcaster's withdraw block. `withdraw` checks if the current block >= the broadcaster's withdraw block.

contracts/jobs/JobsManager.sol
```diff
@@ -238,6 +236,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
                 segmentRange: _segmentRange,
                 claimRoot: _claimRoot,
                 claimBlock: block.number,
+                blockHash: block.blockhash(block.number - 1),
```
yondonfu: Since the previous code before this PR had the problem of using block.blockhash to reference the claim block when shouldVerifySegment is called it was possible for shouldVerifySegment to fail since block.blockhash only works for up to 256 blocks into the past.

Now we're storing a block hash when the claim is made to be used later. The block hash has to be in the past because the block hash of the current block during a claim is not known until the block is actually mined

yondonfu:
I updated the code to use the interim solution we discussed for withdrawal restrictions such that each newly created job is assigned an end block and the latest end block of a broadcaster's jobs becomes the broadcaster's withdraw block.

I also updated the segment challenge code to use a block hash stored during a claim. This does not solve the miner attack originally mentioned for this PR, but it does fix the block.blockhash issue that @dob pointed out that I describe in my inline comment. A fix for the miner attack can probably be pushed out of this PR and this PR can just focus on the withdrawal restrictions

contracts/jobs/JobsManager.sol
dob: Ok, this is definitely something we'll have to address before launch. We need to use a future value. I'm thinking an oracle service that submits blockhashes for block n would be ok, or restricting the verificationPeriod to 250 blocks, which would mean that the block + 1 hash would be available on chain, and it could be written down during the verify call.

dob:
Ok, looks good to me. I wonder if it will be a source of confusion for users that they have to set their own endblock, but since it can be passed in anyway I'm not sure how to provide the default without overloading the Job() function. Doing so would probably create more confusion. So let's leave it as a passed in value and set a default in the client. :ship:

PR #111, opened 2017-11-09
Temporarily remove proxies from migrations and testing with Geth/Parity

yondonfu:
We were encountering some issues with interacting with ManagerProxy contracts. Still investigating the cause of the issues, but for the time being the proxies are removed from the migrations and instead you can interact with the target contracts directly. Once we have a fix for the ManagerProxy issue, we can add the proxies back in.

In an effort to prevent scenarios where we spend a while deploying contracts to a test network with longer block times only to discover a basic bug using the deployed contracts, I'm trying to improve the testing infrastructure to run the integration tests against Geth and Parity to catch those types of bugs earlier (as opposed to previously, the test suite only included unit tests which instantiated contracts individually instead of using the deployed versions from the migrations). Currently using docker-compose to spin up containers for both clients. I still have to include them in CircleCI.


migrations/migrations.config.js
```diff
@@ -17,8 +21,8 @@ module.exports = {
         roundLength: 50
     },
     faucet: {
-        faucetAmount: 100000000000000000000,
-        requestAmount: 1000,
+        faucetAmount: new BigNumber(1000000000000000000000).mul(TOKEN_UNIT),
+        requestAmount: new BigNumber(1000000).mul(TOKEN_UNIT),
```
dob: Nice. Big upgrade to the faucet.

migrations/5_init_protocol.js
```diff
@@ -35,42 +31,25 @@ module.exports = function(deployer, network) {
...
+            controller.setContract(ethUtil.bufferToHex(ethAbi.soliditySHA3(["string"], ["RoundsManager"])), RoundsManager.address)
         ])
     }).then(() => {
-        // Deploy proxy contracts
         return Promise.all([
-            ManagerProxy.new(Controller.address, ethUtil.bufferToHex(ethAbi.soliditySHA3(["string"], ["BondingManagerTarget"]))),
-            ManagerProxy.new(Controller.address, ethUtil.bufferToHex(ethAbi.soliditySHA3(["string"], ["JobsManagerTarget"]))),
-            ManagerProxy.new(Controller.address, ethUtil.bufferToHex(ethAbi.soliditySHA3(["string"], ["RoundsManagerTarget"])))
-        ])
-    }).then(proxies => {
```
dob: Is the idea that after you debug the proxy implementation you'd just change this migration basically back to the original state?

yondonfu: Pretty much, just taking it out for now to not block any efforts to do end to end testing on the test net

PR #113, opened 2017-11-10
docker-compose based tests with CircleCI

yondonfu:
Added docker-compose based tests to the CircleCI config. Now the CI will spin up containers for each ETH client we want to test against. Our unit tests run against TestRPC since they rely heavily on RPC endpoints native to TestRPC, but the integration tests run against all of the clients. For some reason, the Parity tests work locally, but do not work on CircleCI - I keep getting a transaction not processed error so it seems like something is timing out when running in CircleCI. So, I've commented out the Parity related tests for now and can add them back in later once we figure out the CircleCI issue.

The CircleCI config file is also now using CircleCI 2.0 which has native Docker support instead of CircleCI 1.0


run_tests.sh
```diff
@@ -0,0 +1,79 @@
...
+    exit -1
+fi
+
+echo "Starting linting..."
+# Wait for lint
+LINT_TEST_EXIT_CODE=`docker wait lint`
+# Output lint logs
+docker logs lint
+# Check unit-tests output
+if [ -z ${LINT_TEST_EXIT_CODE+x} ] || [ "$LINT_TEST_EXIT_CODE" -ne 0 ] ; then
```
ericxtang: what does this `+x` do?

yondonfu: `${FOO+x}` evaluates to null if FOO is undefined and evaluates to "x" if FOO is defined. The check `[ -z ${FOO+x} ] ` is to check if the exit code is undefined. I think `[ -z ${FOO} ]` would fail to differentiate between FOO being undefined and FOO being an empty string

PR #114, opened 2017-11-13
Double claim segment slashing condition. claimBlock + 1 for challenged segments

### LINKED ISSUE #106: claimedDuplicateSegments() slashing condition
As mentioned in #99, a short term solution for preventing a transcoder from claiming a segment twice is to add a slashing condition. A watcher can call

`claimedDuplicateSegments(segmentNumber, claimID1, claimID2)`

The code could check the segment ranges on those two claims and if the segmentNumber appeared in both, it would enact the slashing condition.

We want to prevent this from getting invoked twice. And we should also consider whether other claims for the same job would be eligible for payout or slashable.

yondonfu:
Added a double claim segment slashing condition to prevent transcoders from claiming a segment more than once. Watchers can trigger this slashing condition and earn a finder's fee.

Triggering this slashing condition kicks out the transcoder and fully refunds a broadcaster for the job which I think makes sense. There needs to be some work done on the logic for the failed verification and missed verification slashing conditions which currently only refund a broadcaster for a claim, but also kick out the transcoder (meaning there are funds stuck in escrow) - I'll track that in a separate issue

Using `block.blockhash(claimBlock + 1)` for challenging segments - acknowledging the limitations of using this value and placing restrictions on the total length of the verification and slashing period.

Closes #106

### DIFF
--- a/contracts/jobs/libraries/JobLib.sol
+++ b/contracts/jobs/libraries/JobLib.sol
@@ -44,14 +44,15 @@ library JobLib {
         view
         returns (bool)
     {
+        // Claim block + 1 must be within the last 256 blocks from the current block
+        require(_claimBlock + 1 > block.number - 256);
         // Segment must be in segment range
         if (_segmentNumber < _segmentRange[0] || _segmentNumber > _segmentRange[1]) {
             return false;
         }
 
         // Use block hash and block number of the block after a claim to determine if a segment
-        // should be verified. Note: this check will only work if _claimBlock + 1 is within
-        // the last 256 blocks from the current block
+        // should be verified
         if (uint256(keccak256(_claimBlock + 1, block.blockhash(_claimBlock + 1), _segmentNumber)) % _verificationRate == 0) {
             return true;
         } else {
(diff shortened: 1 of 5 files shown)


contracts/jobs/libraries/JobLib.sol
```diff
@@ -51,8 +50,9 @@ library JobLib {
         }
 
         // Use block hash and block number of the block after a claim to determine if a segment
-        // should be verified
-        if (uint256(keccak256(_claimBlock, _claimBlockHash, _segmentNumber)) % _verificationRate == 0) {
+        // should be verified. Note: this check will only work if _claimBlock + 1 is within
+        // the last 256 blocks from the current block
+        if (uint256(keccak256(_claimBlock + 1, block.blockhash(_claimBlock + 1), _segmentNumber)) % _verificationRate == 0) {
```
dob: Should we add an `assert()` here to make sure that `_claimBlock + 1 > block.number - 256` ? I forget what block.blockhash returns if the requested block is outside the 256 range, but if it's 0 then this hash could still evaluate to something that `% verificationRate == 0`.

yondonfu: Good call. `block.blockhash` returns 0x0 if the block number is outside the 256 range. The verification period + slashing period check upon initialization of the JobsManager should prevent `shouldVerifySegment` from ever being called with the value of claimBlock + 1 being outside the 256 block range (since it is only ever called within the verificationPeriod and slashingPeriod), but the assert would be useful to enforce/specify the invariant

Issue #115, opened 2017-11-13
Penalties and thresholds for non perfectly attributable slashing conditions


yondonfu:
Currently, a transcoder is automatically kicked out of the active transcoder set and slashed if it triggers the failed verification or missed verification slashing condition. However, a broadcaster is only refunded for a particular claim rather than the entire job. I think the idea here originally was that a transcoder could trigger one of these non perfectly attributable slashing conditions (non perfectly attributable in that a triggered slashing condition does not necessarily always equate to a misbehaving transcoder), but still get the fees for other claims as long as it passes some % threshold of successful verifications. At the moment, the threshold is not implemented so transcoders are kicked out and broadcasters are only partially refunded thereby introducing the possibility of some funds locked in escrow when a transcoder is slashed. The broadcaster should either be completely refunded for the job and the transcoder automatically kicked out upon being slashed OR the broadcaster should be partially refunded just for a claim and the transcoder stays active as long as it still meets the % threshold of successful verifications

yondonfu:
Went with fully refunding broadcaster and automatically kicking out transcoder upon being slashed with the caveat that delegators remain bonded to the transcoder post slashing

PR #116, opened 2017-11-14
Slashing post conditions

yondonfu:
A proposed solution for #115 

When a transcoder is slashed:
- The broadcaster is fully refunded for the job and all escrowed fees are returned
- The transcoder is removed from the active set and the total active set is reduced by the transcoder's active total stake
- The transcoder is removed from the transcoder pools
- Transcoder's delegators remain bonded
- As a result of the above, if the transcoder maintains a sufficient total stake (such that delegators do not delegate to someone else), it can rejoin the transcoder pools next round

The alternative was to implement a threshold such that a transcoder could trigger slashing conditions a certain number of times before it is actually slashed, but that would require additional state management and complexity. This approach is imperfect, but is simpler and does not introduce much additional logic.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -419,12 +419,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         if (penalty > del.bondedAmount) {
             // Decrease transcoder's total stake by transcoder's stake
             transcoderPools.decreaseTranscoderStake(_transcoder, del.bondedAmount);
+            // Decrease delegate's delegated amount
+            delegators[del.delegateAddress].delegatedAmount = delegators[del.delegateAddress].delegatedAmount.sub(del.bondedAmount);
             // Set transcoder's bond to 0 since
             // the penalty is greater than its stake
             del.bondedAmount = 0;
         } else {
             // Decrease transcoder's total stake by the penalty
             transcoderPools.decreaseTranscoderStake(_transcoder, penalty);
+            // Decrease delegate's delegated amount
+            delegators[del.delegateAddress].delegatedAmount = delegators[del.delegateAddress].delegatedAmount.sub(penalty);
             // Decrease transcoder's stake
             del.bondedAmount = del.bondedAmount.sub(penalty);
         }
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -417,21 +417,23 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         Delegator storage del = delegators[_transcoder];
 
         if (penalty > del.bondedAmount) {
-            // Decrease transcoder's total stake by transcoder's stake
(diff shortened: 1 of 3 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -421,20 +417,23 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         Delegator storage del = delegators[_transcoder];
 
         if (penalty > del.bondedAmount) {
-            // Decrease transcoder's total stake by transcoder's stake
```
yondonfu: The transcoder is being removed from the pools in this function call anyway so we don't need to update its total stake

contracts/bonding/BondingManager.sol
```diff
@@ -421,20 +417,23 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            // Decrease delegate's delegated amount
+            delegators[del.delegateAddress].delegatedAmount = delegators[del.delegateAddress].delegatedAmount.sub(penalty);
             // Decrease transcoder's stake
             del.bondedAmount = del.bondedAmount.sub(penalty);
         }
 
-        // Set withdraw round for delegators
-        transcoders[msg.sender].delegatorWithdrawRound = roundsManager().currentRound().add(unbondingPeriod);
+        // Decrease total active stake
+        totalActiveTranscoderStake = totalActiveTranscoderStake.sub(activeTranscoderTotalStake(_transcoder));
```
dob: Is this calculation affected by the lines above which decreased the transcoder's "delegated amount"? I'm not sure if the total stake needs to account for the penalty, or if it's just copied from some value which was written down before the penalty was subtracted.

yondonfu: It's not because this line is subtracting the transcoder's active stake which was locked in for the round. After the transcoder is slashed none of its stake counts toward the total active stake so we shouldn't need to account for the penalty here - we just need to account for the penalty for the transcoder's delegated amount since delegators will still be bonded to it after it is slashed

contracts/bonding/BondingManager.sol
```diff
@@ -287,7 +282,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded) {
             // Withdraw bonded amount which is now unbonded
             amount = amount.add(delegators[msg.sender].bondedAmount);
-            delete delegators[msg.sender];
+            delegators[msg.sender].bondedAmount = 0;
+            delegators[msg.sender].withdrawRound = 0;
```
ericxtang: Should we also set the `delegatedAmount` to 0?

yondonfu: Didn't set `delegatedAmount` to 0 here for consistency - if delegator A is delegated to delegator B and delegator B withdraws, delegator A is still delegated to delegator B since we don't have a way to automatically remove delegator A's delegate. Since delegator A remains delegated to delegator B we can keep `delegatedAmount` for delegator B and only change it if delegator A bonds to another delegator or unbonds

PR #121, opened 2017-11-28
Use job creation block as psuedorandom seed for assigning a transcoder

### LINKED ISSUE #120: Seed used for job assignment
We currently use `block.blockhash(job.creationBlock - 1)` as the seed when running the weighted pick algorithm for selecting a transcoder for a job based upon its stake relative to the total active stake in a round. This is not great as the seed since it is not random as the previous block hash is known when someone calls `job` and an active transcoder is assigned to the job. The weakness here is that someone playing the roles of both the broadcaster and transcoder can wait for opportune times such that the previous block hash when used as a seed in the weighted pick algorithm would lead to the the job being assigned to oneself. Additionally, if the person is also a miner and the value from self-dealing as a transcoder exceeds the value from mining the Ethereum block, it can try to manipulate a block hash and then submit a job the next block that it knows will be assigned to itself.

A workaround is to use `block.blockhash(job.creationBlock + 1)` so that someone does not know the seed at the time of job creation. This is still vulnerable to miner manipulation which is dependent on the value at stake for self-dealing - this is an area for further research. This will change how the client works - transcoders will need to watch for all job events and then compute off-chain for each job whether they are assigned to it. They will then either have to call an additional function `acceptJob` after job.creationBlock + 1 is mined which will check if the transcoder was assigned the job, or this logic can be included in the first `claimWork` call by the transcoder. With either solution, the transcoder would need to take action within 256 blocks of job.creationBlock due to the restrictions on `block.blockhash`. If a transcoder does not take action in time, the simple solution would be that it is out of luck and cannot claim work for the job.

yondonfu:
This uses the block hash of job creation block + 1 as the pseudorandom seed for assigning a transcoder to a job instead of the block hash of job creation block - 1 (which is known ahead of time). A consequence of this jobs are not immediately assigned to a transcoder at the time of creation. Instead, transcoders have to watch for all job creation events and check off-chain which jobs should be assigned to them. Changes will need to be made in the client accordingly.

When calling `claimWork` for the first time for a job, the contract will verify that the caller should be assigned the job based on the block hash of job creation block + 1. Note that since we use `block.blockhash()` to get the block hash for a block, we are limited to the past 256 blocks which means that a transcoder must call `claimWork` to assign itself the job within 256 blocks of the job creation block - at the moment, transcoders simply lose the ability to claim work for the job after those 256 blocks. Also, transcoders need to wait for job creation block + 1 to be mined before it can call `claimWork`. For subsequent calls to `claimWork`, the assigned transcoder is stored for the job, so the contract just checks if the caller is the stored assigned transcoder.

Closes #120

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -59,10 +59,15 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     // Candidate and reserve transcoders
     SortedDoublyLL.Data transcoderPool;
 
-    // Current active transcoders for current round
-    address[] public activeTranscoders;
-    // Total stake of all active transcoders
-    uint256 public totalActiveTranscoderStake;
+    // Represents the active transcoder set
+    struct ActiveTranscoderSet {
+        address[] transcoders;
+        mapping (address => bool) isActive;
+        uint256 totalStake;
+    }
+
+    // Keep track of active transcoder set for each round
+    mapping (uint256 => ActiveTranscoderSet) public activeTranscoderSet;
 
     // Only the RoundsManager can call
     modifier onlyRoundsManager() {
@@ -100,8 +105,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         numActiveTranscoders = _numActiveTranscoders;
         // Set up transcoder pool
         transcoderPool.setMaxSize(_numTranscoders);
-        // Set length of active transcoders array
-        activeTranscoders.length = _numActiveTranscoders;
      }
 
     /*
@@ -159,6 +162,13 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     {
         // Sender must be registered transcoder
         require(transcoderStatus(msg.sender) == TranscoderStatus.Registered);
+
+        uint256 currentRound = roundsManager().currentRound();
+        if (activeTranscoderSet[currentRound].isActive[msg.sender]) {
+            // If transcoder is active for the round set it as inactive
+            activeTranscoderSet[currentRound].isActive[msg.sender] = false;
+        }
+
         // Remove transcoder from pools
         transcoderPool.remove(msg.sender);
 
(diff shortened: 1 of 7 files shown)


contracts/jobs/JobsManager.sol
```diff
@@ -227,11 +224,20 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
...
         require(_segmentRange[1] >= _segmentRange[0]);
 
+        if (job.transcoderAddress != address(0)) {
+            // If transcoder already assigned, check if sender is
+            // the assigned transcoder
+            require(job.transcoderAddress == msg.sender);
+        } else {
+            // If transcoder is not already assigned, check if sender
+            // should be assigned and that job creation block + 1 has been mined and it has been <= 256 blocks since the job creation block
+            require(block.number > job.creationBlock + 1 && block.number <= job.creationBlock + 256 && bondingManager().electActiveTranscoder(job.maxPricePerSegment, job.creationBlock + 1) == msg.sender);
```
dob: If we're using the block that the job creation was mined in, then we should only check that block.number > job.creationBlock.

contracts/bonding/BondingManager.sol
```diff
@@ -501,7 +501,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             address electedTranscoder = availableTranscoders[numAvailableTranscoders - 1];
 
             // Pseudorandomly pick an available transcoder weighted by its stake relative to the total stake of all available transcoders
-            uint256 r = uint256(block.blockhash(block.number - 1)) % totalAvailableTranscoderStake;
+            uint256 r = uint256(block.blockhash(_block)) % totalAvailableTranscoderStake;
```
ericxtang: This means we have to set the round time to be greater than 256 blocks right?  Otherwise would `totalAvailableTranscoderStake` change between job creation and claim?

yondonfu:
Question though...whether we use `job.creationBlock` vs. `job.creationBlock + 1` seems like it shouldn't matter for when a transcoder can start transcoding? Once a job transaction is mined, a transcoder can immediately determine off-chain if it should be assigned to the job and request the stream from the broadcaster knowing that it will be able to claim work for the transcoding it does as long as it calls claimWork for the first time within 256 blocks of the job creation block

ericxtang:
But if it's `creationBlock+1`, the transcoder would need to wait for that block hash right?

yondonfu:
Hm right right, never mind!

yondonfu:
Changed to use `job.creationBlock` instead of `job.creationBlock + 1` for the pseudorandom seed.

In response to @ericxtang 's comment about the race condition of a changing active transcoder set if the round changes between the time of job creation and the time of the first `claimWork` call for the job - added a mapping to map rounds to active transcoder sets so `electActiveTranscoder` can be called with `job.creationRound` and the weighted pick algorithm will use the total active stake and the active transcoder set for that round instead of the current round

PR #122, opened 2017-11-29
Parameter updates + migration scripts + access control

yondonfu:
Still need to do some clean up and fix the failing integration tests (likely due to the yet to be solved weird proxy failure) but pushing this up to get some thoughts on the parameter update mechanism and the new access control logic

- Controller maintains a permissions mapping that describes whether a source address can call a function at a particular target address
- Instead of each manager contract dictating which contracts can call which functions on it, the manager uses a `onlyAuthorized` modifier that checks the controller's permission list if the caller has the appropriate permission to call the function. The only address that can add/revoke permissions in the Controller is its owner
- Permission addition is done in the migration scripts
- Parameter initialization/updates occur using the `setParameters` function defined in each manager contract. Since the parameters can differ in type/number across each manager, the function signature for `setParameters` unfortunately can be different and can't be shared across each manager, so this is a downside. The function uses the `onlyAuthorized` modifier and the idea is that the Controller's owner sets itself as the only address that can call the function. In the beginning this would be a multisig and later, potentially some governance DAO
- Rewrote migration scripts to use async/await syntax to improve readability and control flow logic
- Controller starts off paused so parameter initialization can take place before any other functions can be executed. Manager contracts can also be paused individually - the idea here being that if we run a delegation game, the JobsManager could be paused while the other manager contracts are not

Let me know what thoughts are on this new access control design and the use of a single `setParameters` function for protocol parameter updates


contracts/Controller.sol
```diff
@@ -9,6 +9,13 @@ import "zeppelin-solidity/contracts/lifecycle/Pausable.sol";
 contract Controller is Pausable, IController {
     // Track contract ids and their mapped addresses
     mapping (bytes32 => address) registry;
+    // Track function permissions on registered contracts
+    mapping (address => mapping (address => mapping (bytes4 => bool))) permissions;
```
dob: the first two keys (address[address])....are they `owner_address[contract_address]`, or `contract_address[owner_address]`?

contracts/Controller.sol
```diff
@@ -37,4 +36,42 @@ contract Controller is Pausable, IController {
...
         return IManager(registry[_id]).setController(_controller);
     }
+
+    /*
+     * @dev Add a function permission for a contract
+     * @param _src Source address (caller)
+     * @param _target Target address (contract called)
+     * @param _sig Function signature at target address
+     */
+    function addPermission(address _src, address _target, bytes4 _sig) external onlyOwner {
```
dob: if we update a contract, and the signature changes, then we have to be sure to update permissions as well to use the new signature, right? Can we do this atomically? Or perhaps we can add the permission first before deploying the update?

contracts/bonding/BondingManager.sol
```diff
@@ -95,8 +82,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
     function BondingManager(address _controller) Manager(_controller) {}
 
-    function initialize(uint64 _unbondingPeriod, uint256 _numTranscoders, uint256 _numActiveTranscoders) external beforeInitialization returns (bool) {
-        finishInitialization();
```
dob: If we get rid of the `finishedINitialization()` call and the modifier that checks if the manager was initialized, are we still certain that the contract has been initialized and is in the proper state to be used?

contracts/token/Minter.sol
```diff
@@ -26,35 +26,21 @@ contract Minter is Manager, IMinter {
...
-        require(msg.sender == controller.getContract(keccak256("BondingManager")) || msg.sender == controller.getContract(keccak256("JobsManager")));
-        _;
-    }
-
     function Minter(address _controller, uint256 _initialTokenSupply, uint8 _yearlyInflation) Manager(_controller) {
         initialTokenSupply = _initialTokenSupply;
         yearlyInflation = _yearlyInflation;
     }
 
+    function setParameters(uint8 _yearlyInflation) external onlyAuthorized {
```
dob: Should this be `setParameters` or `setYearlyInflation`?

contracts/verification/LivepeerVerifier.sol
```diff
@@ -23,34 +19,18 @@ contract LivepeerVerifier is Manager, IVerifier {
...
-
-            isSolver[_solvers[i]] = true;
-        }
-        solvers = _solvers;
+    function LivepeerVerifier(address _controller, string _verificationCodeHash) Manager(_controller) {
         // Set verification code hash
         verificationCodeHash = _verificationCodeHash;
     }
 
+    function setParameters(string _verificationCodeHash) external onlyAuthorized {
```
dob: `setParameters` or should it be `setVerificaitonCodeHash()`?

migrations/3_deploy_contracts.js
```diff
@@ -9,31 +11,104 @@ const IdentityVerifier = artifacts.require("IdentityVerifier")
...
+
+        const bondingManager = await deployProxyAndRegister(deployer, controller, BondingManager, "BondingManager", controller.address)
+        const jobsManager = await deployProxyAndRegister(deployer, controller, JobsManager, "JobsManager", controller.address)
+        const roundsManager = await deployProxyAndRegister(deployer, controller, RoundsManager, "RoundsManager", controller.address)
+
+        const owner = await controller.owner()
+
+        deployer.logger.log("Adding permissions...")
+
+        await controller.addPermission(owner, bondingManager.address, functionSig("setParameters(uint64,uint256,uint256)"))
```
dob: I could see a need for a future test that checks all of these hardcoded signatures against actual signatures in our compiled contract ABIs to report if any are not present. This could be an area of pain if we don't have checks, as developers would have to know to come in here to modify this if we ever make a signature change, and that wouldn't be obvious from a failing test.

dob:
Also, what's it mean that the controller starts off paused but the manager defaults `paused = false`? Just a little confused.

contracts/bonding/BondingManager.sol
yondonfu: No, but the idea is that at genesis, since the Controller starts off as paused, all contract functions are disabled and the Controller's owner can "initialize" the contracts by calling `setParameters` on each of the contracts and then afterwords once the Controller is unpaused the contract will have been initialized with the correct parameter values. When we upgrade a contract, the proxy contract will still have the values set by that call to `setParameters`.

This safety here is very dependent on the owner of the Controller being the only address that is allowed to call this function, but I think we have that dependency no matter how we implement parameter updates

contracts/token/Minter.sol
yondonfu: I was thinking that the function used to set parameter values on each contract is the same across all contracts, but open to other thoughts

yondonfu:
Re: controller starting off paused, but the manager defaults to unpaused - the idea was to allow managers to be paused without the controller being paused, but if the controller is paused, all managers are paused. Perhaps, it might be clearer for managers to be `locallyPaused` and a manager is paused if `controller.paused() || locallyPaused` and is unpaused if `!controller.paused() && !locallyPaused`

yondonfu:
For readability purposes, switched back to explicitly defining access controls in the manager contracts themselves instead of storing permissions in the controller.

Defined individual setters for protocol parameters on each manager in addition to a batch setter that can be used for initialization. All protocol parameter setters use the onlyControllerOwner modifier to restrict access exclusively for the controller's owner.

Unfortunately, the majority of our meaningful integration tests would involve fast forwarding block time (TestRPC offers the `evm_mine` RPC call which can accomplish this, but Geth does not), so for now I've removed Geth from the integration test CI. I'll look into other ways we can leverage multiple clients for testing...

Issue #124, opened 2017-11-30
Protocol parameters representing percents need more granular representation


dob:
Right now we use `uint` to represent protocol parameters denominated in percents such as slash percentages and inflation rates. We need something more granular since we may want values < 1% or even intermediate values like 24.334%.

I'd suggest using a CONSTANT represented denominator of 10^6 (1,000,000), and storing the parameters as the numerator. So for example, a slash rate of 1.3% would be stored as the uint 13000.

Anywhere the contracts currently multiply by the param and divide by 100, we would instead multiply by the param and divide by the CONSTANT DIVISOR.

yondonfu:
Closed by #125

Issue #98, opened 2017-10-19
Parameter update mechanism for inflation rate


dob:
When the protocol goes live, we'll have to parameterize it with certain values for inflation, verification rate, etc. We need a mechanism for these parameters to move in the future. Ideally we work towards decentralized governance or automatically adjusting values, but in the early days they may be moveable by more centralized mechanisms like a multisig or contract upgrade.

dob:
Related to automatically adjusting inflation, I wrote a simulation that used the following logic:

1. Set the target bonded participation (in my case 50%).
2. At each round, calculate the % that we're off target. So if 40% is bonded, we're 10% off.
3. Move the inflation rate by the sqrt of the difference. So in this example, move by sqrt(10).

As to how this effected bonding, I randomly chose to increase or decrease the bonded amount in response to the incentives by a random number that was no less than 100,000 and no greater than 2 * the bonded difference from the target. So if we are 4,000,000 bonded with a 5,000,000 target the random number may be as large as 2,000,000.

Since we're dealing with integers, we always move at least 1% in either direction each round. 

This, of course, could be predicted since the movement is forced and restricted based on the above formula, but it pretty quickly converges to a standard inflation range. In reality, I'm not sure how quickly humans would respond to the automatically moving incentives, so we should update the simulation.

I'm interested in other suggestions for the movement function besides sqrt(difference).

dob:
Another thing to look into: If we update parameters which are set as constants in manager contracts, then the storage value may not actually update on the forwarding contract when you deploy a new version of a target contract.

dob:
Updating `N` the number of transcoders would potentially cause issues with the data structures that store the active and transcoder candidate pools.

yondonfu:
Regarding updating parameters which are set as constants in manager contracts:

We could hardcode constants for protocol parameters in the `initialize` function for each manager contract which is only callable by the Controller. So, whenever we deploy a new version of the target contract we could call `upgradeContract(proxyId, targetId, newTargetContract)` in the Controller which does the following:

1. Check if the proxy contract for `proxyId` exists
2. Check if the target contract id for the proxy contract is `targetId`
3. Register `newTargetContract` as `targetId`
4. Call `initialize` on the proxy contract for `proxyId` which will forward the `initialize` call to `newTargetContract`. The `initialize` call will use the logic implemented in `newTargetContract`, but will execute in the storage context of the proxy contract for `proxyId`

[...]

PR #125, opened 2017-12-06
Constant inflation change model

### LINKED ISSUE #16: Block-time proof the inflation in the protocol
If the block times change it could affect our inflation rate. We need a solution that maintains our yearly inflation target regardless of underlying block times.

yondonfu:
- Implemented a simple constant inflation change model. If we are below the target bonding rate, we increase the inflation by a small amount. If we are above the target bonding rate, we decrease the inflation by a small amount. As discussed offline, this implementation is shooting for simplicity
- Representing percentages as numerator values with the denominator set as `PERC_DIVISOR = 1000000` which allows us to specify percentage values with greater precision
- Changed the inflation rate to a per round value (it was previously a per year value)
- Since we no longer attempt to compute the number of rounds in a year, we do not have a way to spread redistributed tokens across the rounds in a year. Instead, right now tokens in the redistribution pool are split amongst the transcoders in the next round. This might not be the right approach, so this can be up for discussion

Closes #16 #98 #124


contracts/Manager.sol
```diff
@@ -8,6 +8,9 @@ contract Manager is IManager {
     // Controller that contract is registered with
     IController public controller;
 
+    // Divisor used for representing percentages
+    uint256 public constant PERC_DIVISOR = 1000000;
```
dob: think we should add an assert anywhere in the constructor to make sure that `PERC_DIVISOR` is > 0, or even that `PERC_DIVISOR == 1000000`. It would be disasterous if it ever got changed upon deployment of a new contract, right? I guess the constructor is the only place, but I don't know if it's necessary.

contracts/token/Minter.sol
```diff
@@ -95,14 +110,35 @@ contract Minter is Manager, IMinter {
      * @dev Set the reward token amounts for the round. Only callable by the RoundsManager
      */
     function setCurrentRewardTokens() external onlyRoundsManager whenSystemNotPaused returns (bool) {
+        setInflation();
+
         currentMintableTokens = mintedTokensPerRound();
         currentMintedTokens = 0;
-        currentRedistributableTokens = redistributableTokensPerRound();
+        currentRedistributableTokens = redistributionPool;
```
dob: I know that redistributing in the current round might be simplest, but do you think it would be wrong just to spread it over some number `n` rounds, rather than all at once? It could arbitrarily be 100 rounds?

contracts/token/Minter.sol
```diff
@@ -95,14 +110,35 @@ contract Minter is Manager, IMinter {
...
         return true;
     }
 
+    /*
+     * @dev Set inflation based upon the current bonding rate
+     */
+    function setInflation() internal {
+        uint256 currentBondingRate = (bondingManager().getTotalBonded() * PERC_DIVISOR) / livepeerToken().totalSupply();
+
+        uint256 newInflation = inflation;
```
dob: any reason to use the `newInflation` variable? Or could we eliminate it just be updating inflation itself in the if statement?

contracts/token/Minter.sol
```diff
@@ -95,14 +110,35 @@ contract Minter is Manager, IMinter {
...
+
+        if (currentBondingRate < targetBondingRate) {
+            // Bonding rate is below the target - increase inflation
+            newInflation = newInflation.add(inflationChange);
+        } else if (currentBondingRate > targetBondingRate) {
+            // Bonding rate is above the target - decrease inflation
+            newInflation = newInflation.sub(inflationChange);
```
dob: do we have to manually check for underflow? We don't want `.sub()` to throw if inflation goes to 0...we just want it to remain at 0.

contracts/token/Minter.sol
```diff
@@ -95,14 +110,35 @@ contract Minter is Manager, IMinter {
...
+        }
```
ericxtang: Should we add a check here to make sure inflation stays between some range?  (at least above 0?)

yondonfu: Discussed previously offline with @dob - adding upper/lower bounds might make sense, but they also introduce additional parameters that we do not know the right values for. Additionally, there might be some benefit in allowing the inflation rate to float freely and allow  participation by delegators with various incentives and liquidity preferences adjust the rate without us trying to prescribe what we believe to be the right bounds. An example could be if the inflation does end up starting to increase out of control, anyone with belief in the network value can be play a sort of "lender of last resort" role (ala a central bank except anyone can play this role) and bond to increase the participation rate to the target rate and stop the inflation growth. That being said, I think this is just an initial implementation to add the functionality and we can definitely do more research and discuss better models

contracts/token/Minter.sol
yondonfu: Picking an arbitrary number of rounds feels a bit weird, but I also don't like redistributing the entire pool in the next round since its an immediate benefit for any transcoder that chose to distribute fees after its delegators unbonded in the previous round. The latter might be worse in terms of how transcoder incentives are affected so maybe spreading it out even if it is over an arbitrary number of rounds is better at least for the time being

contracts/Manager.sol
yondonfu: This is currently set as a constant so it wouldn't be changed in the constructor - if it is for some reason changed it would because the hardcoded constant value is changed in a newly deployed target contract. We could include asserts for `PERC_DIVISOR == 1000000` whenever we calculate percentages, but we would need to change all the assert statements as well if the value of the divisor does happen to change. 1000000 allows us to represent percentages as low as .0001% so I'm not sure if we would ever change the divisor, what do you think?

dob: Ok yah I don't expect us to change it intentionally. I was worried about the scenario where we deploy an updated contract and accidentally remove or add a 0. But that's probably overthinking it. I certainly don't think it needs asserts everywhere.

PR #127, opened 2017-12-07
More events. Remove unnecessary bool return values

### LINKED ISSUE #31: Events
We should define events so that clients can listen for occurrences of common actions on chain.

* Job()
* EndJob()
* ClaimJob()
* VerificationPassed()
* VerificationFailed()
* Deposit()
* Withdrawl()
* Bond()
* Unbond()
* Slash()
* Reward()
* Transcoder()
* NewRound()

I'm sure there are more.

yondonfu:
- Added more events that I thought might be useful for a client interacting with the contracts
- Removed unnecessary bool returns values for state mutating functions. Since our convention is to fail hard and fail early by reverting transactions upon any type of error, bool return values are never useful. There are some cases that a bool return value might be useful even if we're reverting, but I don't believe there are any cases of that in our code right now.
- Removed the constructor and fallback function from LivepeerToken because they shouldn't be needed - no ETH can be sent to the contract if the fallback function is not payable and the fallback function is non-payable by default

Current list of events (excluding LivepeerVerifier and LivepeerToken):

ParameterUpdate
SetContract
SetController
TranscoderUpdate
TranscoderEvicted
TranscoderResigned
Bond
Unbond
Reward
NewRound
NewJob
NewClaim
Verify
DistributeFees
ReceivedVerification
NewInflation
SetCurrentRewardTokens

Let me know if there are any additional events in particular that anyone thinks might be particularly helpful. Maybe @jozanza has some thoughts relevant for the SDK?

Closes #31

### DIFF
--- a/contracts/Migrations.sol
+++ b/contracts/Migrations.sol
@@ -10,15 +10,15 @@ contract Migrations {
             _;
     }
 
-    function Migrations() {
+    function Migrations() public {
         owner = msg.sender;
     }
 
-    function setCompleted(uint completed) restricted {
+    function setCompleted(uint completed) public restricted {
         last_completed_migration = completed;
     }
 
-    function upgrade(address new_address) restricted {
+    function upgrade(address new_address) public restricted {
         Migrations upgraded = Migrations(new_address);
         upgraded.setCompleted(last_completed_migration);
     }
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -172,7 +172,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         t.pendingFeeShare = _feeShare;
         t.pendingPricePerSegment = _pricePerSegment;
 
-        uint256 currentRound = roundsManager().currentRound();
         uint256 delegatedAmount = delegators[msg.sender].delegatedAmount;
 
         // Check if transcoder is not already registered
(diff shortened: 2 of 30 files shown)


dob:
What about events for 

* WIthdrawing funds
* Failing Verification
* Slashing

contracts/Migrations.sol
```diff
@@ -10,15 +10,15 @@ contract Migrations {
             _;
     }
 
-    function Migrations() {
+    function Migrations() public {
```
dob: These methods that are marked `public`...they're `public` by default and we're just adding this to be explicit, right? Good idea.

contracts/bonding/BondingManager.sol
```diff
@@ -282,7 +292,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             }
         }
 
-        return true;
+        Bond(_to, msg.sender, currentRound);
```
dob: Are we including `currentRound` in all the events for convenience and searchability on the frontend? Presumably the client would know which block or round it occured in based upon the event metadata itself, right...but this would just allow us to filter or easily view a timeline?

contracts/Migrations.sol
yondonfu: Yep, the compiler also emits warnings if you don't add explicit visibility declarations - a non-functionality related todo is to do clean up any compiler warnings and make the CI fail if we have any warnings

contracts/bonding/BondingManager.sol
yondonfu: Yeah good point - the current round can always be computed from the block number during which an event is received

yondonfu:
Removed round values from events as they can always be recalculated using the block number associated with an event. Also added `Withdraw`, `Deposit` and verification events with more specific naming (`PassedVerification` and `FailedVerification`)

PR #129, opened 2017-12-08
Add a lock period during a round s.t. transcoders cannot change rates

### LINKED ISSUE #62: Transcoder shouldn't be able to update their pending fees/price immediately before new round
There needs to be a period of time where delegators can review a transcoders fees when they aren't subject to change. Otherwise there is an attack where transcoders advertise great rates, and then switch them at the last minute before being elected in.

I think the simplest temporary implementation might be that the `transcoder()` function throws if you're within `transcoderFreezeBlocks` of the start of the next round. Perhaps this is a couple hours worth? It's not great UX though.

The other option is that this is just defended against socially. If people notice this happening, then they delegate away, and we don't need to update the implementation. 

Finally, people could set a client side flag that says that if a transcoder changes the rates beyond a certain threshold, their client automatically watches and unbonds.

yondonfu:
Added a lock period during a round during which transcoders cannot join the transcoder pool or change their rates. The lock period is calculated as a percentage of the round length.

Closes #62

### DIFF
--- a/contracts/rounds/RoundsManager.sol
+++ b/contracts/rounds/RoundsManager.sol
@@ -14,6 +14,10 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
     // Round length in blocks
     uint256 public roundLength;
 
+    // Lock period of a round as a % of round length
+    // # of blocks in the lock period = (roundLength * roundLockAmount) / PERC_DIVISOR
+    uint256 public roundLockAmount;
+
     // Last initialized round. After first round, this is the last round during which initializeRound() was called
     uint256 public lastInitializedRound;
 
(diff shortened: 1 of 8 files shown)


contracts/rounds/RoundsManager.sol
```diff
@@ -14,18 +14,25 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
     // Round length in blocks
     uint256 public roundLength;
 
+    // Lock period of a round as a % of round length
```
dob: Let's just add one line comment that says something like: "Transcoders can not change their rates within the `roundLockAmount` % of the end of a round to prevent a last minute change after delegators have bonded."

PR #131, opened 2017-12-12
blockNum() in RoundsManager

yondonfu:
- Fix a bug in the BondingManager where a delegator that has never bonded to anyone before has a `lastClaimTokenPoolsSharesRound = 0` and when calling a bonding related function tries to claim rewards and fees in every round since then through the current round. A delegator that has never bonded to anyone before should not try to claim reward and fees when calling a bonding related function and instead should just update its `lastClaimTokenPoolsSharesRound` to the current round.
- Add a `blockNum()` function in the RoundsManager that returns the current block number. This allows us to mock out the value returned by `blockNum()` in tests instead of relying on the TestRPC specific `evm_mine` and `evm_increaseTime` RPC methods. AdjustableRoundsManager is used in integration tests and overrides `blockNum()` to return a value that can be manually set.
- As a result of being able to mock the value returned by `blockNum()`, we can target Geth/Parity with our integration tests again since the tests do not rely on any TestRPC specific features


contracts/bonding/BondingManager.sol
```diff
@@ -820,27 +820,29 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function updateDelegatorWithTokenPoolsShares(address _delegator, uint256 _endRound) internal {
         Delegator storage del = delegators[_delegator];
 
-        uint256 currentBondedAmount = del.bondedAmount;
-        uint256 currentUnbondedAmount = del.unbondedAmount;
+        if (del.lastClaimTokenPoolsSharesRound > 0) {
```
yondonfu: If this condition is not true, then the delegator has never been bonded to anyone before

contracts/jobs/JobsManager.sol
```diff
@@ -378,6 +382,9 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         // Sender must be elected transcoder
         require(job.transcoderAddress == msg.sender);
 
+        uint256 blockNum = roundsManager().blockNum();
+        // Claim block + 1 must be within the last 256 blocks from the current block
+        require(blockNum < 256 || claim.claimBlock + 1 >= blockNum - 256);
```
yondonfu: This check was originally in `JobLib.shouldVerifySegment`, but `JobLib` does not have access to RoundsManager to get the block number, so this check is moved out of `JobLib` and into `JobsManager`

contracts/rounds/AdjustableRoundsManager.sol
```diff
@@ -0,0 +1,22 @@
+pragma solidity ^0.4.17;
+
+import "./RoundsManager.sol";
+
+
+contract AdjustableRoundsManager is RoundsManager {
```
yondonfu: This contract is just used for integration tests and not for production deployment. It overrides `blockNum()` to return a state variable that can be modified by custom setters

ericxtang:
Makes sense.  🚢

PR #134, opened 2017-12-26
Compute percentage points first when computing reward

yondonfu:
Currently, in `minter.createReward`, we calculate the mintable reward portion as `(mintableAmount * fracNum) / fracDenom)`. In the extreme case, both `mintableAmount` and `fracNum` can be very large (an example being when there is a single transcoder that is capturing all of the inflationary rewards for a few rounds leading to both a large `mintableAmount` and `fracNum`). In this case, it is possible for `mintableAmount * fracNum` to overflow.

As a proposed fix, `minter.createReward` can instead compute the percentage points represented by `fracNum` and `fracDenom` as `(fracNum * PERC_DIVISOR) / fracDenom)`. Since `PERC_DIVISOR` is known to be a constant 10^6, the risk of overflow when computing `fracNum * PERC_DIVISOR)` is less than the risk of overflow when computing `mintableAmount * fracNum` since both values here are unbounded. 

We then compute the mintable reward portion as `(mintableAmount * percPoints) / PERC_DIVISOR)`. `percPoints` is also bounded in magnitude by the magnitude of `PERC_DIVISOR`.

This doesn't completely eliminate the risk of overflow because if `fracNum` ever gets very very large (it is the delegated stake of a transcoder) it can still overflow when multiplied by `PERC_DIVISOR`. This fix just mitigates the risk somewhat.

### DIFF
--- a/contracts/token/Minter.sol
+++ b/contracts/token/Minter.sol
@@ -78,8 +78,10 @@ contract Minter is Manager, IMinter {
      * @param _fracDenom Denominator of fraction (total active stake)
      */
     function createReward(uint256 _fracNum, uint256 _fracDenom) external onlyBondingManager whenSystemNotPaused returns (uint256) {
+        uint256 percPoints = _fracNum.mul(PERC_DIVISOR).div(_fracDenom);
+
         // Compute fraction of redistributable tokens to include in reward
-        uint256 redistributeAmount = currentRedistributableTokens.mul(_fracNum).div(_fracDenom);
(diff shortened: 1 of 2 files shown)


dob:
Might be interesting to deploy the updated contract to our testnet and use the upgrade mechanism. Do we have the key for the owner account?

________________________________

yondonfu:
I think with `PERC_DIVISOR = 1000000` as long as the total issued token and the mintable tokens for a round are less than `(2^256) / 1000000)` we should be ok.

I believe Eric has the owner account key.

PR #136, opened 2017-12-30
Use ETH for job deposits and fee payments

yondonfu:
This PR changes job deposits and fee payouts to be denominated in ETH rather than LPT.

Broadcasters now use the payable `deposit` function in JobsManager to make a ETH deposit to pay for transcode jobs. Similar to with LPT, all ETH is forwarded to the Minter, which is the interim owner of all funds in the protocol until they are paid out.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -41,7 +41,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     // Represents a delegator's current state
     struct Delegator {
         uint256 bondedAmount;                    // The amount of bonded tokens
-        uint256 unbondedAmount;                  // The amount of unbonded tokens
+        uint256 fees;                            // The amount of fees collected
         address delegateAddress;                 // The address delegated to
         uint256 delegatedAmount;                 // The amount of tokens delegated to the delegator
         uint256 startRound;                      // The round the delegator transitions to bonded phase and is delegated to someone
@@ -279,19 +279,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         }
 
         if (_amount > 0) {
-            if (_amount > del.unbondedAmount) {
-                // If amount to bond is greater than the delegator's unbonded amount
-                // use the delegator's unbonded amount and transfer the rest from the sender
-                uint256 transferAmount = _amount.sub(del.unbondedAmount);
-                // Set unbonded amount to 0
-                del.unbondedAmount = 0;
-                // Transfer the token to the Minter
-                livepeerToken().transferFrom(msg.sender, minter(), transferAmount);
-            } else {
-                // If the amount to bond is less than or equal to the delegator's unbonded amount
-                // just use the delegator's unbonded amount
-                del.unbondedAmount = del.unbondedAmount.sub(_amount);
-            }
+            // Transfer the LPT to the Minter
+            livepeerToken().transferFrom(msg.sender, minter(), _amount);
         }
 
         Bond(_to, msg.sender);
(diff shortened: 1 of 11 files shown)


contracts/token/Minter.sol
```diff
@@ -107,6 +107,22 @@ contract Minter is Manager, IMinter {
...
+     * @param _amount Amount of ETH
+     */
+    function transferETH(address _to, uint256 _amount) external onlyBondingManagerOrJobsManager whenSystemNotPaused {
+        _to.transfer(_amount);
+    }
+
+    /*
+     * @dev Receive ETH from JobsManager
+     */
+    function receiveETH() external payable onlyJobsManager whenSystemNotPaused returns (bool) {
```
yondonfu: Since the Minter shouldn't accept ETH from any address except for the JobsManager I thought it would make sense to define this specific function with the modifier restricting access to the JobsManager rather than making the fallback function payable and then doing an additional check there for the sender address

contracts/bonding/BondingManager.sol
```diff
@@ -347,22 +347,22 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        uint256 amount = 0;
-
         if (delegators[msg.sender].unbondedAmount > 0) {
-            // Withdraw unbonded amount
-            amount = amount.add(delegators[msg.sender].unbondedAmount);
+            // Withdraw unbonded amount (ETH)
+            uint256 unbondedAmount = delegators[msg.sender].unbondedAmount;
             delegators[msg.sender].unbondedAmount = 0;
+
+            minter().transferETH(msg.sender, unbondedAmount);
```
dob: I'm a little confused about what "withdrawing" after the unbonding period, has to do with ETH. Maybe the term is overloaded...perhaps withdrawing a deposit should be separate than withdrawing after the unbonding period has ended, since one will be withdrawing ETH and the other withdrawing LPT?

dob: maybe `withdrawDeposit` and `withdrawStake`?

contracts/jobs/JobsManager.sol
```diff
@@ -233,15 +233,14 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
...
+     * @dev Deposit ETH for jobs
      */
-    function deposit(uint256 _amount) external whenSystemNotPaused {
-        broadcasters[msg.sender].deposit = broadcasters[msg.sender].deposit.add(_amount);
-        // Transfer tokens for deposit to Minter. Sender needs to approve amount first
-        livepeerToken().transferFrom(msg.sender, minter(), _amount);
+    function deposit() external payable whenSystemNotPaused {
+        broadcasters[msg.sender].deposit = broadcasters[msg.sender].deposit.add(msg.value);
+        // Transfer ETH for deposit to Minter
+        minter().receiveETH.value(msg.value)();
```
dob: Maybe name this function in the minter `deposit()` as well? Up to you. It feels a little weird to call a function called "receive"

contracts/jobs/JobsManager.sol
```diff
@@ -253,7 +252,7 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
 
         uint256 amount = broadcasters[msg.sender].deposit;
         delete broadcasters[msg.sender];
-        minter().transferTokens(msg.sender, amount);
+        minter().transferETH(msg.sender, amount);
```
dob: Similarly, maybe we should call this `withdraw` as well?

yondonfu:
- Separated the withdraw function in the BondingManager into two separate functions `withdrawStake`, which withdraws a delegator's bonded stake after the unbonding period, and `withdrawFees`, which withdraws a delegator's fees (currently tracked through the `fees` field).
- Changed the delegator field previously named `unbondedAmount` to `fees` since it is exclusively used for tracking the amount of fees a delegator has collected
- ETH related functions in the Minter are now named `depositETH` and `withdrawETH`

yondonfu:
Also updated naming:
- `delegatorStake` is now `pendingStake`
- `delegatorFees` is now `pendingFees`

contracts/bonding/BondingManager.sol
```diff
@@ -41,7 +41,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     // Represents a delegator's current state
     struct Delegator {
         uint256 bondedAmount;                    // The amount of bonded tokens
-        uint256 unbondedAmount;                  // The amount of unbonded tokens
+        uint256 fees;                            // The amount of fees collected
```
dob: Ah, see that this was updated in a subsequent commit.

PR #139, opened 2018-01-09
Add git commit hash as version identifier to contract info in Controller

yondonfu:
Proposing the use of the SHA1 hash of the head git commit at the time of deployment as the version identifier for contracts. The Controller would store a contract's address in addition to the Git commit hash at the time of contract deployment. 

Ex.

If we deploy an updated version of the BondingManager contract, we can examine the Git commit hash associated with the contract id `keccak256("BondingManagerTarget")` to determine what version of the target contract that the proxy contract is using.

I thought using this hash as a version identifier would be more meaningful since serve as a direct reference in the Git repository - I don't immediately see a need for an additional version string besides this, but open to thoughts.

### DIFF
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,6 +1,8 @@
 FROM node:8.5.0
 
 WORKDIR /app/
+# Need .git so we can get the git head commit hash
+COPY .git /app/.git
 COPY .babelrc /app/.babelrc
 COPY .soliumrc.json /app/.soliumrc.json
 COPY .soliumignore /app/.soliumignore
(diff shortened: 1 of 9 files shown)


yondonfu:
Interesting point about building separate code paths in the client...

I don't think the client would be using a version identifier to determine which code path to use in the case of an upcoming upgrade. If an upgrade is announced to occur at a block N in the future that involves a change in a contract's ABI, the client could include a code path that incorporates those changes and then once the current block >= N the client would use the code path with the changes. The client should always be using the Controller to retrieve the contract addresses, so it should never have to decide what contract to talk to. The version identifier is more important for anyone to confirm that the deployed contract code matches with the contract code stored at a particular Git commit (if we wanted to take this one step further we could also include the hash of the bytecode...).

dob:
Got it. Sounds good.

PR #142, opened 2018-01-15
Inflation change and verifier solver setters. Tweak initial parameters

yondonfu:
- Setter for inflation change
- Setter for adding solvers to LivepeerVerifier
- Fix LivepeerVerifier tests
- Tweak initial parameters for test deployments

### DIFF
--- a/contracts/verification/LivepeerVerifier.sol
+++ b/contracts/verification/LivepeerVerifier.sol
@@ -55,6 +55,14 @@ contract LivepeerVerifier is Manager, IVerifier {
         verificationCodeHash = _verificationCodeHash;
     }
 
+    function addSolver(address _solver) external onlyControllerOwner {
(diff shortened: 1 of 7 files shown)


contracts/verification/LivepeerVerifier.sol
```diff
@@ -55,6 +55,14 @@ contract LivepeerVerifier is Manager, IVerifier {
         verificationCodeHash = _verificationCodeHash;
     }
 
+    function addSolver(address _solver) external onlyControllerOwner {
+        // Must not already be a solver
+        require(!isSolver[_solver]);
```
ericxtang: We did `require(!isSolver[_solvers[i]] && _solvers[i] != address(0));` earlier, should we do `address(0)` check here too?

PR #144, opened 2018-01-16
Genesis state as a part of deployment

yondonfu:
Adding genesis state setup to the deployment which involves:
- Deploying all the Livepeer protocol contracts
- Deploying the TokenDistribution and GenesisManager contracts
- Transferring ownership of LivepeerToken to the GenesisManager
- Setting the initial token allocations in the GenesisManager
- Starting genesis and the token distribution
- Adding token grants during genesis and the token distribution
- After the token distribution ends, vesting tokens start their vesting schedule and time locked tokens are released
- Vesting tokens and time locked tokens are held by separate contracts for each recipient
- After genesis ends, ownership of LivepeerToken is transferred to the protocol Minter

The current deployment scripts use a mock token distribution contract that can be manually ended. In reality, we would probably deploy all the contracts and kick off genesis, wait until the token distribution completes and then run an additional script afterwords to end genesis.

Also created a `ContractDeployer` module that wraps the Truffle deployer to enable async await syntax which we can reuse in additional migration scripts

Looking for review/thoughts on this set up first and then I'll add tests

### DIFF
--- a/contracts/GenesisManager.sol
+++ b/contracts/GenesisManager.sol
@@ -1,0 +1,264 @@
+pragma solidity ^0.4.17;
+
+import "./token/ILivepeerToken.sol";
+import "./token/ITokenDistribution.sol";
+
+import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "zeppelin-solidity/contracts/ownership/Ownable.sol";
+import "zeppelin-solidity/contracts/token/TokenVesting.sol";
+import "zeppelin-solidity/contracts/token/TokenTimelock.sol";
+
+
+contract GenesisManager is Ownable {
+    using SafeMath for uint256;
+
+    // LivepeerToken contract
+    ILivepeerToken public token;
+    // TokenDistribution contract
+    ITokenDistribution public tokenDistribution;
+
+    // Address of the Livepeer bank multisig
+    address public bankMultisig;
+    // Address of the Minter contract in the Livepeer protocol
+    address public minter;
+
+    // Initial token supply issued
+    uint256 public initialSupply;
+    // Crowd's portion of the initial token supply
+    uint256 public crowdSupply;
+    // Company's portion of the initial token supply
+    uint256 public companySupply;
+    // Team's portion of the initial token supply
+    uint256 public teamSupply;
+    // Investors' portion of the initial token supply
+    uint256 public investorsSupply;
+    // Community's portion of the initial token supply
+    uint256 public communitySupply;
+
+    // Token amount in grants for the team
+    uint256 public teamGrantsAmount;
+    // Token amount in grants for investors
+    uint256 public investorsGrantsAmount;
+    // Token amount in grants for the community
+    uint256 public communityGrantsAmount;
+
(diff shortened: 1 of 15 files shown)


contracts/test/TokenDistributionMock.sol
```diff
@@ -0,0 +1,44 @@
...
+        over = false;
+    }
+
+    function finalize() external onlyOwner {
+        require(!isOver());
+
+        over = true;
+
+        // Send this contract's balance to the faucet
+        uint256 balance = token.balanceOf(this);
```
yondonfu: This mock token distribution contract just receives the initial crowd supply and then transfers it to the faucet when it ends

contracts/GenesisManager.sol
```diff
@@ -0,0 +1,268 @@
...
+    function addVestingGrant(
+        address _receiver,
+        uint256 _amount,
+        uint256 _timeToCliff,
+        uint256 _vestingDuration
+    )
+        internal
+    {
+        // Receiver must not have already received a grant with a vesting schedule
+        require(vestingHolders[_receiver] == address(0));
```
dob: we'll need two different addresses if the same person is receiving multiple grants. that's ok.

contracts/GenesisManager.sol
```diff
@@ -0,0 +1,268 @@
...
+        // Create a vesting holder contract to act as the holder of the grant's tokens
+        // Note: the vesting grant is revokable
+        TokenVesting holder = new TokenVesting(_receiver, startTime, _timeToCliff, _vestingDuration, true);
+        vestingHolders[_receiver] = holder;
+
+        // Transfer ownership of the vesting holder to the owner of this contract
+        // giving the owner of this contract to revoke the grant
+        holder.transferOwnership(owner);
+
+        token.transfer(holder, _amount);
```
dob: look at TokenVesting contract to see how it is aware of how much token is initially granted, and therefore how it knows how much to vest. What if more token is sent to the vesting contract after its begun.

contracts/GenesisManager.sol
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /**
+     * @dev End genesis
+     */
+    function end() external onlyOwner atStage(Stages.GenesisStart) {
+        // Token distribution must be over
+        require(tokenDistribution.isOver());
```
dob: Up for discussion whether this returns true when the registration period is over or the actual initial distribution is over.

migrations/4_genesis.js
```diff
@@ -0,0 +1,75 @@
...
+
+        deployer.logger.log("Adding community token grants...")
+
+        await Promise.all(genesis.communityGrants.map(grant => {
+            return genesisManager.addCommunityGrant(grant.receiver, grant.amount)
+        }))
+
+        deployer.logger.log("Finalizing token distribution...")
+
+        await tokenDistribution.finalize()
```
dob: In reality this line and everything after would occur in a separate script since there is a time gap between setting up genesis and finalizing.

PR #145, opened 2018-01-17
Update linting, CI and enable offline signing for deployment

yondonfu:
- Update linter to use the latest Solium version which includes a new config file format and comes with a default security plugin. We use most of the security plugin rules, but turn off a few of them such as the ones for no block members (we use block numbers and timestamps) and no inline assembly (we use inline assembly to implement the delegate proxy pattern for upgradeable contracts)
- Moved OraclizeVerifier into its own `verification` directory as it is not used at the moment and it generates a lot of warnings from the Solidity compiler since it imports `usingOraclize.sol` which does not comply with a lot of the recommendations from the Solidity compiler
- Small fixes in the contracts to adhere to the Solium rule set and to silence warnings from the Solidity compiler. The remaining compiler warnings are either for mock test contracts (compiler complains about unused variables in mocked functions) or for opcodes that are only available after the Byzantium hardfork (i.e. returndatacopy. This warning should go away in a future update to the compiler)
- Updated CI process. The build process will compile and push a built Docker image containing the contracts to our DockerHub. The built image is used to run our unit tests and integration tests targeting Parity and Geth, and can also be pulled from DockerHub from anyone that wants access to a container with all the dependencies installed and contracts compiled. The tests should also run slightly quicker during the process
- Replaced ethereumjs-testrpc with its new replacement ganache-cli (pretty much just a naming change)
- Enabled offline transaction signing for deployments to LP test net (maybe we should come up with a one word name for this...) by using truffle-hdwallet-provider-privkey. Instead of relying on the account being unlocked on the node that we are connected to we can sign all the transactions for deployments locally and then send them off to the node


contracts/verification/LivepeerVerifier.sol
```diff
@@ -98,6 +107,7 @@ contract LivepeerVerifier is Manager, IVerifier {
      * @param _requestId Request identifier
      * @param _result Result of verification computation - keccak256 hash of transcoded segment data
      */
+    // solium-disable-next-line mixedcase
```
dob: is this a directive to the linter to ignore the warning?

truffle.js
```diff
@@ -28,8 +24,9 @@ module.exports = {
             gas: 6700000
         },
         lpTestNet: {
-            host: "localhost",
-            port: 8545,
+            provider: () => {
+                return new HDWalletProvider(process.env.LPTESTNET_PRIV_KEY, "http://ethrpc-testnet.livepeer.org:8545")
```
dob: this is cool

contracts/verification/LivepeerVerifier.sol
yondonfu: Yeah we never use mixedcase except for here because it follows the convention used by OraclizeVerifier which inherits `__callback` from the Oraclize API contract. Debatable whether we should follow that convention, but figured silencing the warning here, but enforcing the rule everywhere else would suffice for now

yondonfu:
Yeah agreed on the smaller, focused PRs bit, I got a little carried away with this one...

PR #148, opened 2018-01-19
GenesisManager unit test

yondonfu:
- GenericMock contract that is meant to be used for mocking out any return values that a contract depends on from another contract. Ideally, we can switch to using this in all of our tests once I tweak it a little more rather than having a bunch of different mock contracts
- GenesisManager unit test

Integration test for genesis state to come separately

### DIFF
--- a/contracts/test/GenericMock.sol
+++ b/contracts/test/GenericMock.sol
@@ -11,9 +11,10 @@ contract GenericMock {
         bytes32 bytes32Value;
         bool boolValue;
         MockValueType valueType;
+        bool set;
     }
 
-    enum MockValueType { Uint256, Bytes32, Bool, None }
+    enum MockValueType { Uint256, Bytes32, Bool }
 
     // Track function selectors and mapped mock values
     mapping (bytes4 => MockValue) mockValues;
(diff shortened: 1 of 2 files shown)


contracts/test/GenericMock.sol
```diff
@@ -0,0 +1,123 @@
...
+    function setMockBool(bytes4 _func, bool _value) external {
+        mockValues[_func].valueType = MockValueType.Bool;
+        mockValues[_func].boolValue = _value;
+        mockValues[_func].set = true;
+    }
+
+    /*
+     * @dev Return mock value for a functione
+     */
+    function() public {
```
dob: Just curious....What's the deal with there being a really restricted gas limit by default in the fallback function again? Is it only on default sends which include a fixed gas amount? 

Would the fact that this method does more than just log an event be a problem in any way?

yondonfu: Yeah the `<address>.send` and `<address>.transfer` functions get a 2300 gas stipend so the fallback function can only log an event. 

Shouldn't be a problem here since we only ever use GenericMock to mock out return values for defined getter functions

yondonfu:
Yep I'll follow up with an integration test to validate post genesis state including the state of the various vesting/timelock contracts, token balances, Controller, token ownership, etc.

PR #150, opened 2018-01-22
Fix job assignment

### LINKED ISSUE #147: Job routing is always routing jobs to highest staked transcoder and not routing proportionally
Currently on testnet the transcoder delegation looks like this:

```
+----------------------+
|REGISTERED TRANSCODERS|
+----------------------+
+----+--------------------------------------------+--------+------------------------------+----------------+---------------+-------+------------------------+-----------------------+---------------+
| ID |                  ADDRESS                   | ACTIVE |       DELEGATED STAKE        | REWARD CUT (%) | FEE SHARE (%) | PRICE | PENDING REWARD CUT (%) | PENDING FEE SHARE (%) | PENDING PRICE |
+----+--------------------------------------------+--------+------------------------------+----------------+---------------+-------+------------------------+-----------------------+---------------+
|  0 | 0xFeBF35b94e16E018Ab5cA9aA1f66aaC1a1AB29E0 | true   | 69828.859299524624845138 LPT |             20 |            10 | 5 WEI |                     20 |                    10 | 5 WEI         |
|  1 | 0x97Ab31E8E0BE434CdD02a0C017266894E81Cdd82 | true   | 69646.494846492990017785 LPT |             10 |             5 | 1 WEI |                     10 |                     5 | 1 WEI         |
|  2 | 0x102551aDEaC5b4d043b6B2423e774dc3Db004031 | true   | 43664315968921069 LPTU       |             10 |             5 | 1 WEI |                     10 |                     5 | 1 WEI         |
+----+--------------------------------------------+--------+------------------------------+----------------+---------------+-------+------------------------+-----------------------+---------------+
```

A local broadcaster has his broadcasting profile set to offer 10 LPT/segment. The jobs should roughly be split between transcoders 0 and 1, but all jobs are being routed to transcoder 0.

yondonfu:
Fixes #147 

Implements solution #3 described in the issue. The Go client will require a small change to pass in the job creation block hash instead of the job creation block number as well. Includes an integration test for job assignment (the test might be flaky sometimes because it runs a bunch of trials and computes the percentage of assigned jobs for each transcoder and compares the computed percentage against the expected percentage with some acceptable delta)

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -522,7 +522,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * Returns address of elected active transcoder and its price per segment
      * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
      */
-    function electActiveTranscoder(uint256 _maxPricePerSegment, uint256 _block, uint256 _round) external view returns (address) {
+    function electActiveTranscoder(uint256 _maxPricePerSegment, bytes32 _challengeHash, uint256 _round) external view returns (address) {
         uint256 activeSetSize = activeTranscoderSet[_round].transcoders.length;
         // Create array to store available transcoders charging an acceptable price per segment
         address[] memory availableTranscoders = new address[](activeSetSize);
@@ -548,7 +548,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             address electedTranscoder = availableTranscoders[numAvailableTranscoders - 1];
 
             // Pseudorandomly pick an available transcoder weighted by its stake relative to the total stake of all available transcoders
-            uint256 r = uint256(roundsManager().blockHash(_block)) % totalAvailableTranscoderStake;
+            uint256 r = uint256(_challengeHash) % totalAvailableTranscoderStake;
             uint256 s = 0;
 
             for (uint256 j = 0; j < numAvailableTranscoders; j++) {
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -521,8 +521,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @dev Pseudorandomly elect a currently active transcoder that charges a price per segment less than or equal to the max price per segment for a job
      * Returns address of elected active transcoder and its price per segment
      * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
+     * @param _blockHash Job creation block hash used as a pseudorandom seed for assigning an active transcoder
+     * @param _round Job creation round
      */
-    function electActiveTranscoder(uint256 _maxPricePerSegment, bytes32 _rand, uint256 _round) external view returns (address) {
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "ganache": "./node_modules/.bin/ganache-cli -l 0x663BE0",
     "compile": "rm -rf build/contracts && truffle compile --all",
     "test:unit": "bash scripts/run_ganache.sh && truffle test test/unit/*",
-    "test:integration": "truffle test test/integration/*",
+    "test:integration": "truffle test test/integration/Delegation.js && truffle test test/integration/JobAssignment.js",
```
yondonfu: This is a hack to get the integration tests to pass - Truffle is supposed to redeploy all contracts between test runs to guarantee a clean room environment, but this does not seem to be working for some reason so for the time being this NPM script will manually invoke each integration test

contracts/bonding/BondingManager.sol
```diff
@@ -522,7 +522,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * Returns address of elected active transcoder and its price per segment
      * @param _maxPricePerSegment Max price (in LPT base units) per segment of a stream
      */
-    function electActiveTranscoder(uint256 _maxPricePerSegment, uint256 _block, uint256 _round) external view returns (address) {
+    function electActiveTranscoder(uint256 _maxPricePerSegment, bytes32 _rand, uint256 _round) external view returns (address) {
```
dob: Should we rename `_rand` to `_blockHash`, and add `@param` comments for this param and `_round`?

contracts/bonding/IBondingManager.sol
```diff
@@ -20,7 +20,7 @@ contract IBondingManager {
     function setActiveTranscoders() external;
     function updateTranscoderWithFees(address _transcoder, uint256 _fees, uint256 _round) external;
     function slashTranscoder(address _transcoder, address _finder, uint256 _slashAmount, uint256 _finderFee) external;
-    function electActiveTranscoder(uint256 _maxPricePerSegment, uint256 _block, uint256 _round) external view returns (address);
+    function electActiveTranscoder(uint256 _maxPricePerSegment, bytes32 _rand, uint256 _round) external view returns (address);
```
dob: same comment about `_rand`, as it's not obvious what that is supposed to represent.

package.json
yondonfu: Never mind this doesn't play well with the CI - will use a different fix

yondonfu:
AdjustableRoundsManager inherits from RoundsManager and allows callers to manually set the block number and the block hash. This is only used for tests so that we can control block numbers and block hashes within the tests (ganache-cli aka testrpc has a custom RPC method for controlling the block number, but this method is not available on clients like Geth or Parity so we would not be able to rely on the custom RPC method when running tests on those clients). The regular RoundsManager would be used in a real deployment.

package.json
yondonfu: Switched the integration tests to use ganache-cli for now because there seems to be a bug in Truffle with redeploying contracts between test runs which I've filed an issue for

PR #163, opened 2018-01-30
Slashing fixes with accompanying tests

### LINKED ISSUE #162: Slashing fixes
Tracking a few slashing related fixes in this issue:

- `slashTranscoder` in `BondingManager` currently always fails for an active transcoder because `activeTranscoderTotalStake` will always fail if the transcoder being slashed is set as inactive first. The order of operations in `slashTranscoder` should be changed such that the total active stake for the round is decremented by the value returned by `activeTranscoderTotalStake` BEFORE setting the transcoder as inactive.
- `slashTranscoder` should not check that a transcoder is registered. Currently, the check causes 2 problems:
    - A transcoder that unbonds and automatically resigns as a transcoder becomes unslashable
    - A transcoder that is slashed can still call `claimWork` and fault, but it becomes unslashable
    the second time since it is no longer registered

yondonfu:
Closes #162 

Addresses the two problems mentioned in the issue. Added integration tests for each slashing condition in: `FailedVerificationSlashing.js`, `MissedVerificationSlashing.js` and `DoubleClaimSegmentSlashing.js`. Also added `SlashingEdgeCases.js` that tests that the bugs/exploits described in the issue are fixed - there is one test for slashing a transcoder after it has unbonded and resigned and one test for slashing a transcoder after it has already been slashed because it committed another fault (the scenario used in the test is the transcoder first double claims segments, is slashed and then misses verification for a subsequent transcode claim). Additional clarifying comments made in-line.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -797,6 +797,14 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     }
 
     /*
+     * @dev Return whether a transcoder is registered
+     * @param _transcoder Transcoder address
+     */
+    function isRegisteredTranscoder(address _transcoder) public view returns (bool) {
+        return transcoderStatus(_transcoder) == TranscoderStatus.Registered;
+    }
+
+    /*
      * @dev Remove transcoder
      */
     function resignTranscoder(address _transcoder) internal {
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -477,24 +477,25 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         uint256 currentRound = roundsManager().currentRound();
 
         if (activeTranscoderSet[currentRound].isActive[_transcoder]) {
-            // Set transcoder as inactive
-            activeTranscoderSet[currentRound].isActive[_transcoder] = false;
             // Decrease total active stake for the round
             activeTranscoderSet[currentRound].totalStake = activeTranscoderSet[currentRound].totalStake.sub(activeTranscoderTotalStake(_transcoder, currentRound));
+            // Set transcoder as inactive
+            activeTranscoderSet[currentRound].isActive[_transcoder] = false;
         }
 
         // Remove transcoder from pools
         transcoderPool.remove(_transcoder);
 
-        // Award finder fee
-        if (penalty > 0 && _finder != address(0)) {
+        // Account for penalty
+        if (penalty > 0) {
             uint256 burnAmount = penalty;
 
+            // Award finder fee if there is a finder address
             if (_finder != address(0)) {
-                // Award finder fee
                 uint256 finderAmount = MathUtils.percOf(penalty, _finderFee);
                 minter().transferTokens(_finder, finderAmount);
 
+                // Subtract finder fee from the amount to be burned
                 burnAmount = burnAmount.sub(finderAmount);
             }
 
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -464,12 +464,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
         Delegator storage del = delegators[_transcoder];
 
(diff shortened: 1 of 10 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -457,44 +457,48 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
-        // Remove transcoder from pools
-        transcoderPool.remove(_transcoder);
+        // If registered transcoder, remove from pool
+        if (transcoderStatus(_transcoder) == TranscoderStatus.Registered) {
+            transcoderPool.remove(_transcoder);
+        }
 
-        // Award finder fee
-        if (penalty > 0 && _finder != address(0)) {
```
yondonfu: This was a bug - previously if there was not a finder address provided, `slashTranscoder` would not burn the slashed tokens. This change makes sure that `slashTranscoder` always burns the slashed tokens and if there is a finder address provided, it subtracts the finder's fee from the amount to be burned.

contracts/jobs/JobsManager.sol
```diff
@@ -488,9 +488,9 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
             // Protocol slashes transcoder for failing verification (no finder)
             bondingManager().slashTranscoder(transcoder, address(0), failedVerificationSlashAmount, 0);
 
-            PassedVerification(transcoder, _jobId, _claimId, _segmentNumber);
```
yondonfu: This was a bug - previously `receiveVerification` would fire the `PassedVerification` event when verification fails...This change makes sure `PassedVerification` is fired when verification succeeds.

contracts/rounds/AdjustableRoundsManager.sol
```diff
@@ -26,7 +26,7 @@ contract AdjustableRoundsManager is RoundsManager {
     }
 
     function blockHash(uint256 _block) public view returns (bytes32) {
-        require(_block <= blockNum() - 256);
```
yondonfu: This was a bug (though it only affected tests and not a live deployed contract system because in production `AdjustableRoundsManager` would never be used - only `RoundsManager`). This change makes sure that `blockHash` reverts if the block number provided is not greater than the current block - 256 to match up with the behavior of `blockHash` in `RoundsManager`.

migrations/3_deploy_contracts.js
```diff
@@ -20,13 +19,7 @@ module.exports = function(deployer, network) {
...
         const token = await lpDeployer.deployAndRegister(LivepeerToken, "LivepeerToken")
         await lpDeployer.deployAndRegister(Minter, "Minter", controller.address, config.minter.inflation, config.minter.inflationChange, config.minter.targetBondingRate)
-
-        if (network === "development" || network === "testrpc" || network === "parityDev" || network === "gethDev") {
-            await lpDeployer.deployAndRegister(IdentityVerifier, "Verifier", controller.address)
-        } else {
-            await lpDeployer.deployAndRegister(LivepeerVerifier, "Verifier", controller.address, config.verifier.solvers, config.verifier.verificationCodeHash)
-        }
-
+        await lpDeployer.deployAndRegister(LivepeerVerifier, "Verifier", controller.address, config.verifier.solvers, config.verifier.verificationCodeHash)
```
yondonfu: I don't think there is any practical use for `IdentityVerifier` anymore because even in tests we can just use the `LivepeerVerifier` and if we want to skip verification we would just not have a solver callback to the contract with a result.

yondonfu:
Once the transcoder is slashed once, it would have to register again, wait until the next round and if the job is still active, it would have to explicitly try to trigger the failed verification slashing condition on its own in order to be slashed by the whitelisted LivepeerVerifier solver again (there would need to be another request created in LivepeerVerifier because the original request is deleted the first time the solve calls `__callback()`). 

I think we still need the check in `claimWork()` to make sure a transcoder is registered because without it the following attack is still possible: transcoder unbonds and resigns, waits until 1 block before the unbonding period ends, calls `claimWork()` for a still active job to drain the broadcaster, calls `withdrawStake()` right after before a watcher can do anything

PR #170, opened 2018-02-02
Ensure roundLength can be increased/decreased

### LINKED ISSUE #160: Ensure that roundLength can be updated without ending up with duplicate round numbers
At the moment, `currentRound()` in `RoundsManager` can produce a round number that has already occurred if `roundLength` is increased. We should ensure that the protocol never sees the same round number twice. @dob proposed a solution based on keeping track of the last round and block that `roundLength` was updated.

So, (blockNum() - lastUpdatedRoundLengthBlock) / roundLength = rounds since roundLength was updated the result being we always maintain the number of rounds that have happened given a particular value of roundLength that can be added with the highest round number we reached previously with a different value of roundLength. I think we would also need to record lastUpdatedRoundLengthRound such that the current round = lastUpdatedRoundLengthRound + (blockNum() - lastUpdatedRoundLengthBlock) / roundLength. So when we call setRoundLength(), lastUpdatedRoundLengthRound = currentRound(), lastUpdatedRoundLengthBlock = blockNum(). initializeRound() would continue to set lastInitializedRound = currentRound() whenever it is called as long as lastInitializedRound < currentRound().

yondonfu:
Fixes #160 

Makes sure that the roundLength parameter can be updated properly - previously if we increased roundLength, we would end up with duplicate round numbers. Also refactored RoundsManager to improve path coverage. It is now using a unit test fixture that uses the GenericMock contract to stub out the rest of the contract dependencies. The goal is to move all our unit tests to this architecture, with the RoundsManager unit test being the first one to move over and the plan is to move the rest of the unit tests over in separate PRs. Additional comments in-line

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sol linguist-language=Solidity
--- a/migrations/3_deploy_contracts.js
+++ b/migrations/3_deploy_contracts.js
@@ -45,6 +45,9 @@ module.exports = function(deployer, network) {
             config.jobsManager.doubleClaimSegmentSlashAmount,
             config.jobsManager.finderFee
         )
-        await roundsManager.setParameters(config.roundsManager.roundLength, config.roundsManager.roundLockAmount)
+
+        // Set RoundsManager parameters
+        await roundsManager.setRoundLength(config.roundsManager.roundLength)
+        await roundsManager.setRoundLockAmount(config.roundsManager.roundLockAmount)
     })
 }
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "eslint-config-google": "^0.9.1",
     "ethereumjs-abi": "^0.6.4",
     "ethereumjs-util": "^5.1.2",
-    "ganache-cli": "^6.0.3",
+    "ganache-cli": "^6.1.0-beta.0",
     "mkdirp": "^0.5.1",
     "solium": "^1.1.2",
     "truffle": "^4.0.4",
--- a/contracts/rounds/IRoundsManager.sol
+++ b/contracts/rounds/IRoundsManager.sol
@@ -1,10 +1,11 @@
 pragma solidity ^0.4.17;
 
 
-/*
- * @title Interface for RoundsManager
+/**
+ * @title RoundsManager interface
  */
-contract IRoundsManager {
+interface IRoundsManager {
+    // Events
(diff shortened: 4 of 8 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.sol linguist-language=Solidity
```
yondonfu: Enables syntax highlighting for Solidity on Github

package.json
```diff
@@ -42,7 +42,7 @@
     "eslint-config-google": "^0.9.1",
     "ethereumjs-abi": "^0.6.4",
     "ethereumjs-util": "^5.1.2",
-    "ganache-cli": "^6.0.3",
+    "ganache-cli": "^6.1.0-beta.0",
```
yondonfu: Updated to new beta version for ganache-cli which is supposed to have a fix for the issue of random crashes during tests which has been plaguing us for awhile

migrations/3_deploy_contracts.js
```diff
@@ -45,6 +45,9 @@ module.exports = function(deployer, network) {
             config.jobsManager.doubleClaimSegmentSlashAmount,
             config.jobsManager.finderFee
         )
-        await roundsManager.setParameters(config.roundsManager.roundLength, config.roundsManager.roundLockAmount)
+
+        // Set RoundsManager parameters
+        await roundsManager.setRoundLength(config.roundsManager.roundLength)
```
yondonfu: Removed `setParameters` in favor of just using individual setters. Plan is to do the same for other contracts in separate PRs

ericxtang:
@yondonfu I think here is a case that screws with the lastRoundLengthUpdateStartBlock.

 Round | 0 | 299 | 300 | 301
-- | -- | -- | -- | --
setRoundLength | 1000 | 100 | 1000 | 100
LRLURound | 0 | 2 | 3 | 3
LRLUStartBlock | 0 | 200 | 300 | 3000(incorrect)

yondonfu:
I think the most recently pushed commit should fix that case:

Now `setRoundLength()` uses `currentRoundStartBlock()` to set `lastRoundLengthUpdateStartBlock` as `lastRoundLengthUpdateStartBlock + (rounds since last update) * the old round length`. This makes sure we always calculate the new value for `lastRoundLengthUpdateStartBlock` based off of the old value

ericxtang:
🚢

PR #172, opened 2018-02-05
Small fixes

### LINKED ISSUE #156: Contracts should inherit their associated interfaces
Many of our contracts such as `BondingManager` and `JobsManager` have associated interfaces such as `IBondingManager` and `IJobsManager`. We should make sure that all contracts inherit from their associated interfaces for consistency. For example, `LivepeerToken` should inherit from `ILivepeerToken`, but currently it does not.

### LINKED ISSUE #157: Create a getter for the max size of the transcoder pool
The max size of the transcoder pool is currently set by the function `setNumTranscoders` in `BondingManager`, but there is no getter in `BondingManager` that allows an external caller to retrieve the max size of the transcoder pool.

### LINKED ISSUE #161: transferTokens() in Minter can only be called by BondingManager
The current logic of the modifier is a remnant from when payments were still denominated in LPT so the JobsManager had to call transferTokens() on the Minter when a broadcaster withdrew its LPT deposit. Since payments are now denominated in ETH the JobsManager calls depositETH() and withdrawETH() instead. As a result, transferTokens() should only be called by BondingManager.

### LINKED ISSUE #164: Ensure that user cannot try to claim rewards and fees past the current round
Currently, `claimTokenPoolsShares` accepts an `_endRound` parameter. There should be a validation check to make sure `_endRound <= currentRound`.

yondonfu:
Fixes #156 
Fixes #157 
Fixes #164 
Fixes #161 

This includes a bunch of smaller fixes for open issues.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -563,7 +563,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @param _endRound The last round for which to claim token pools shares for a delegator
      */
     function claimTokenPoolsShares(uint256 _endRound) external {
+        // End round must be after the last claim round
         require(delegators[msg.sender].lastClaimTokenPoolsSharesRound < _endRound);
+        // End round must not be after the current round
+        require(_endRound <= roundsManager().currentRound());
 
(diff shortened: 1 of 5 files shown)


ericxtang:
🚢

PR #178, opened 2018-02-06
Small JobsManager fixes and additional coverage in unit test

### LINKED ISSUE #165: doubleClaimSegmentSlash fails if one of the claims is already complete
**Description**

Currently, doubleClaimSegmentSlash() accepts two claim ids and checks that both claims are still pending. If either claim is not pending (as in the transcoder has submitted it, but has not called distributeFees() for it yet), the slashing invocation fails.

**Exploit**

A transcoder submits transcode claim A, passes verification for challenged segments and calls distributeFees() to mark transcode claim A as complete. She then submits transcode claim B with the same segment range as transcode claim B. A watcher tries to trigger the double claim segment slashing condition against the transcoder, but the function call fails because even though transcode claim B is pending, transcode claim A is complete.

**Solution**

The require statement in doubleClaimSegmentSlash() should instead check that both claims are either complete or pending. If both claims are complete that means, the transcoder was able to distribute fees for both claims - the watcher can retroactively slash the transcoder if no one slashed the transcoder for double claiming segments previously.

yondonfu:
Fixes #165 

Small fixes for JobsManager and some work around improving coverage for its associated unit test. Additional comments in-line.


contracts/jobs/JobsManager.sol
```diff
@@ -527,9 +492,9 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         uint256 blockNum = roundsManager().blockNum();
         uint256 challengeBlock = claim.claimBlock + 1;
         // Must be after verification period
-        require(blockNum > claim.endVerificationBlock);
+        require(blockNum >= claim.endVerificationBlock);
```
yondonfu: Treating the `endVerificationBlock` as the block in which the verification period is over

contracts/test/GenericMock.sol
```diff
@@ -62,10 +63,21 @@ contract GenericMock {
...
+        mockValues[_func].valueType = MockValueType.Address;
+        mockValues[_func].addressValue = _value;
+        mockValues[_func].set = true;
+    }
+
     /*
      * @dev Return mock value for a functione
      */
-    function() public {
+    function() public payable {
```
yondonfu: Allows this mock contract to receive ETH (relevant for testing `deposit()` in JobsManager which forwards ETH to the Minter)

test/unit/JobsManager.js
```diff
@@ -40,176 +47,288 @@ contract("JobsManager", accounts => {
...
-            const inactiveJobId = 1
-            // Transcoder claims work for inactive job
-            await expectThrow(jobsManager.claimWork(inactiveJobId, segmentRange, claimRoot, {from: electedTranscoder}))
+        it("should fail if the transcoder address is not set and the caller is not assigned", async () => {
+            await expectThrow(jobsManager.claimWork(1, segmentRange, claimRoot, {from: accounts[2]}))
         })
 
-        it("should fail if the transcoder is assigned and the sender is not the assigned transcoder", async () => {
-            await jobsManager.claimWork(jobId, segmentRange, claimRoot, {from: electedTranscoder})
+        it("should set the transcoder address if this is the first claim and the caller is assigned", async () => {
```
ericxtang: Do unit tests in truffle have some kind of guaranteed ordering?  Feels like we should check for `jInfo[4] == 0`, and then `claimWork`, and see that `jInfo[4]` changed.

test/unit/JobsManager.js
```diff
@@ -316,314 +404,288 @@ contract("JobsManager", accounts => {
...
         })
 
-        it("should update job escrow and set claim as complete", async () => {
-            // Fast foward through verificaiton and slashing period
-            const verificationPeriod = await jobsManager.verificationPeriod.call()
-            const slashingPeriod = await jobsManager.slashingPeriod.call()
-            await fixture.roundsManager.mineBlocks(verificationPeriod.add(slashingPeriod).toNumber() + 1)
-
-            await jobsManager.distributeFees(jobId, claimId, {from: electedTranscoder})
+        it("should fail if the claim is not pending", async () => {
```
ericxtang: Should probably check for both `Slashed` and `Complete`.

test/unit/JobsManager.js
```diff
@@ -658,205 +720,281 @@ contract("JobsManager", accounts => {
...
+            await jobsManager.missedVerificationSlash(0, 0, 0, {from: watcher})
 
-            // Should fail because claim is already slashed
-            await expectThrow(jobsManager.missedVerificationSlash(jobId, claimId, segmentNumber, {from: accounts[2]}))
+            const cStatus = (await jobsManager.getClaim(0, 0))[5]
+            assert.equal(cStatus, 1, "wrong claim status")
         })
     })
 
     describe("doubleClaimSegmentSlash", () => {
```
ericxtang: We should probably make sure the transcoder can be slashed in any status (add it to bondingManager.js?)

test/unit/JobsManager.js
```diff
@@ -40,176 +47,288 @@ contract("JobsManager", accounts => {
...
+            assert.equal(bInfo[1], currentBlock + 50, "wrong broadcaster withdrawBlock")
+        })
 
-            const endBlock = 100 + 500
-            const creationBlock = 100
-            await jobsManager.job(streamId, transcodingOptions, maxPricePerSegment, endBlock, {from: broadcaster})
+        it("should not change the broadcaster's withdraw block if the job's end block <= broadcaster's withdraw block", async () => {
+            await jobsManager.deposit({from: broadcaster, value: 1000})
+            await jobsManager.job("foo", transcodingOptions, 1, currentBlock + 50, {from: broadcaster})
+            await jobsManager.job("foo", transcodingOptions, 1, currentBlock + 50, {from: broadcaster})
```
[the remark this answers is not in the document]
yondonfu: Yeah since the second job is created with the same end block as the first job, this is testing that the creation of a second job with an end block <= the broadcaster's withdraw block leaves the broadcaster's withdraw block unchanged.

test/unit/JobsManager.js
yondonfu: The beforeEach hook for this describe block ensures that each test starts with the same exact state so that test ordering does not matter so `jInfo[4]` will always be 0 before calling `claimWork` and the tests for `job` test that `jInfo[4]` is initialized to be the null address.

test/unit/JobsManager.js
yondonfu: Yep planning on adding the tests for the BondingManager unit test

contracts/jobs/JobsManager.sol
```diff
@@ -104,49 +104,14 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
...
-     * @dev Batch set protocol parameters. Only callable by the controller owner
-     * @param _verificationRate % of segments to be verified
-     * @param _verificationPeriod Number of blocks to complete verification of claimed work
-     * @param _slashingPeriod Number of blocks after the verification period to submit slashing proofs
-     * @param _failedVerificationSlashAmount % of stake slashed for failed verification
-     * @param _missedVerificationSlashAmount % of stake slashed for missed verification
-     * @param _doubleClaimSegmentSlashAmount % of stake slashed of double claiming a segment
-     * @param _finderFee % of slashed amount awawded to finder
-     */
-    function setParameters(
```
dob: What's the reason for removing this? We have individual setters, so we can reduce contract size and redundant functionality that needs to be maintained?

test/unit/JobsManager.js
```diff
@@ -40,176 +47,288 @@ contract("JobsManager", accounts => {
...
+
+        it("should zero out broadcaster's deposit and withdraw block", async () => {
+            // Fast forward through end block
+            await fixture.roundsManager.setMockUint256(functionSig("blockNum()"), endBlock)
+
+            await jobsManager.withdraw({from: broadcaster})
+
+            const bInfo = await jobsManager.broadcasters.call(broadcaster)
+            assert.equal(bInfo[0], 0, "wrong deposit")
+            assert.equal(bInfo[1], 0, "wrong withdraw block")
```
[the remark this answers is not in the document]
yondonfu: The Minter dependency is mocked such that `minter().withdrawETH()` just returns a bool, but the function call itself doesn't do anything because the mocked Minter dependency doesn't actually implement `withdrawETH()` - we just tell the mocked dependency to return a bool when it receives that function call. Testing the ETH transfer triggered by a call to the Minter could be done in an integration test that uses a non-mocked Minter

contracts/jobs/JobsManager.sol
yondonfu: Yeah primarily to reduce contract size and redundant functionality

yondonfu:
Sounds good and agree with being more vigilant with reviewing tests during code reviews and treating them as a form of spec/docs!

PR #180, opened 2018-02-06
Set jobs as inactive when a transcoder is slashed

### LINKED ISSUE #175: A transcoder can be slashed, come back and register and then claim work for that job again
Currently, if a job has an end block far in the future, a transcoder can be slashed, come back and register and then claim work for that job again. This might not be a big deal though because if the transcoder faults again it will be slashed. 

Should we just set a job as inactive if a transcoder is slashed? Is there ever a scenario where the transcoder could come back and start doing honest work for the job again? I feel like setting the job as inactive would simplifiy logic while reducing flexibity, but I’m not sure the flexibility is needed. However, in a future where multiple transcoders might be able to handle a job perhaps we should just leave the job active
If a transcoder is slashed for a claim and comes back and submits a claim that shares segments with the slashed claim except it is honestly transcoding this time should it be able to be slashed?

yondonfu:
Fixes #175 

As described in the issue, the minor downside to the approach taken in this PR is that it removes the previous invariant: a broadcaster cannot withdraw its deposit if it has any active jobs. If a job is set as inactive by setting its end block to the current block when the job's transcoder is slashed, the contract does not maintain enough state (and ideally shouldn't have to) to also update the broadcaster's withdraw block to reflect the possibility that the broadcaster no longer has any active jobs. As a result, the restriction is no longer that a broadcaster cannot withdraw its deposit if it has any active jobs since a broadcaster might not be able to withdraw even if it does not have any active jobs. The rules for withdrawal are now:

-  A broadcaster commits to keeping a deposit on-chain until the largest end block it defines at the time of job creation
- The largest end block provided for a new job is the broadcaster's withdraw block
- A broadcaster cannot withdraw its deposit until its withdraw block
- A broadcaster's withdraw block can only become larger, it cannot become smaller

I think this is an OK tradeoff because in the current system, whether we set the job as inactive by setting its end block to the current block or introduce additional state to check if the transcoder was previously slashed for a job (in order to prevent the transcoder from claiming work for the job again), no other transcoder will be able to claim work for that job. Plus, either way we don't have a good way to update the broadcaster's withdraw block - so in both scenarios a broadcaster must wait until its withdraw block. Though in the future, we might want to add the feature where transcoders can fill in as the elected transcoder for a job if the originally elected transcoder was slashed, I don't think setting jobs as inactive for now will impact the implementation of that feature in the future.

### DIFF
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -676,6 +676,8 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         uint256 fees = job.escrow;
         job.escrow = job.escrow.sub(fees);
         broadcasters[job.broadcasterAddress].deposit = broadcasters[job.broadcasterAddress].deposit.add(fees);
+        // Set end block of job to current block - job becomes inactive
+        job.endBlock = roundsManager().blockNum();
     }
 
     /*
--- a/test/unit/JobsManager.js
+++ b/test/unit/JobsManager.js
@@ -24,6 +24,17 @@ contract("JobsManager", accounts => {
     const DOUBLE_CLAIM_SEGMENT_SLASH_AMOUNT = 40 * PERC_MULTIPLIER
     const FINDER_FEE = 4 * PERC_MULTIPLIER
 
+    const JobStatus = {
+        Inactive: 0,
+        Active: 1
+    }
+
+    const ClaimStatus = {
+        Pending: 0,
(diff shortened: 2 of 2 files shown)


yondonfu:
Yeah...that's an unfortunate result of the current implementation and I'm not sure if there is a better solution for that other than being very careful about how the client sets the end block for users, provide ample warning to users if they choose to craft their own job creation transactions and in the worst case deploying an update that removes the withdraw block check

PR #183, opened 2018-02-06
Fix stuck tokens bug if a delegator unbonds and then immediately re-bonds

### LINKED ISSUE #166: Delegators that unbond and then bond before the unbonding period is over become stuck
@ericxtang found this bug where a delegator can end up in a weird state such that it cannot perform any bonding related actions including withdrawing its stake:

Delegator A is bonded to transcoder B.  A calls `unbond()` - as a result A now has a `withdrawRound`, a null `delegateAddress` and is in the unbonding state. Additionally, in the process of calling `unbond()`, A's last claim round is set to the current round N. 1 round passes. Before the unbonding period is over, A calls `bond()` towards B. Now, A has a `withdrawRound`, a `delegateAddress` set to the address of B and a `startRound` set to 0 (`bond()` only sets your `startRound` if you are in the unbonded state OR you are already delegated towards someone and are delegating to someone else - in this case A is in the unbonding state and is not already delegated to someone). Note that 1) the claimable stake for B's token pools in round N + 1 does not include A's bonded stake because it was locked in for that round before A bonded toward B 
 2) A's last claim round is still set to N, but the current round is N + 1. Let's say 5 rounds pass and now the current round is N + 6 and B's token pools for each round was initialized with its claimable stake. Now, if A tries to call any bonding related function thereby automatically claiming rewards and fees from past rounds, it will try to claim rewards and fees starting from round N. As a part of claiming rewards and fees for a round, A's bonded stake is subtracted from B's claimable stake for that round. In this case, A's bonded stake > B's claimable stake which causes the transaction to fail.

Fixing this requires two things:
- Ensuring that a delegator cannot call `bond()` if it is in the unbonding state - if a delegator calls `unbond()` it must wait through the unbonding period before it can call `bond()` again if it changes its mind and decides that it does not want to withdraw its stake and instead wants to bond to someone again
- If a delegator claims rewards and fees through a certain round by calling `unbond()`, if it decides to `bond()` again after the unbonding period, it should now claim rewards and fees starting from the last claim round because from its last claim round through the current round it was not bonded to anyone. My initial instinct is to just zero out the last claim round if you call `unbond()` and treat the next time you call `bond()` as if you were calling `bond()` for the first time (instead of iterating from 0 through current round, you just skip the claiming process and just set the last claim round to the current round because you were not bonded to anyone from round 0 through current round)

### LINKED ISSUE #167: Should delegators be able to unbond and rebond to cancel the delegation period?
I don't think the protocol currently supports it.  @dob pointed out that "If we have a multi week or month unbonding period, it would be kind of lame if you couldn’t cancel it to rebond towards someone else. Especially since we don’t have partial unbonding."

### LINKED ISSUE #181: JobExists() vs. _jobId < numJobs
Small thing - is there a reason we use _jobId < numJobs instead of jobExists() here?  

https://github.com/livepeer/protocol/blob/936bb90db0476e5e826cd861defb5e6972ca733f/contracts/jobs/JobsManager.sol#L351

yondonfu:
Fixes #167 
Fixes #166 
Fixes #181

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -216,6 +216,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundInitialized
         autoClaimTokenPoolsShares
     {
+        // Caller cannot be in the unbonding state
+        require(delegatorStatus(msg.sender) != DelegatorStatus.Unbonding);
+
         Delegator storage del = delegators[msg.sender];
 
         uint256 currentRound = roundsManager().currentRound();
@@ -302,6 +305,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         del.delegateAddress = address(0);
         // Unbonding delegator does not have a start round
         del.startRound = 0;
+        // Unbonding delegator's last claim round is reset
+        del.lastClaimTokenPoolsSharesRound = 0;
 
         // If caller is a registered transcoder, resign
         // In the future, with partial unbonding there would be a check for 0 bonded stake as well
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -327,7 +327,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         external
         whenSystemNotPaused
         currentRoundInitialized
-        autoClaimTokenPoolsShares
     {
         // Delegator must be in the unbonded state
         require(delegatorStatus(msg.sender) == DelegatorStatus.Unbonded);
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -308,8 +308,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         del.delegateAddress = address(0);
         // Unbonding delegator does not have a start round
         del.startRound = 0;
-        // Unbonding delegator's last claim round is reset
-        del.lastClaimTokenPoolsSharesRound = 0;
 
         // If caller is a registered transcoder, resign
         // In the future, with partial unbonding there would be a check for 0 bonded stake as well
@@ -867,7 +865,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function updateDelegatorWithTokenPoolsShares(address _delegator, uint256 _endRound) internal {
         Delegator storage del = delegators[_delegator];
 
-        if (del.lastClaimTokenPoolsSharesRound > 0) {
+        // Only will have earnings to claim if you have a delegate
+        // If not delegated, skip the earnings claim process
+        if (del.delegateAddress != address(0)) {
             uint256 currentBondedAmount = del.bondedAmount;
             uint256 currentFees = del.fees;
 
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -220,14 +220,12 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
         uint256 currentRound = roundsManager().currentRound();
 
-        if (delegatorStatus(msg.sender) != DelegatorStatus.Bonded) {
+        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded || delegatorStatus(msg.sender) == DelegatorStatus.Unbonding) {
             // New delegate
             // Set start round
+            // Don't set start round if delegator is in pending state because the start round would not change
             del.startRound = currentRound.add(1);
-        }
-
-        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonding) {
-            // If transitioning from unbonding state
+            // If transitioning from unbonding or unbonded state
             // make sure to zero out withdraw round
             del.withdrawRound = 0;
         }
@@ -249,6 +247,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
                 // Decrease old transcoder's total stake
                 transcoderPool.updateKey(del.delegateAddress, transcoderPool.getKey(del.delegateAddress).sub(del.bondedAmount), address(0), address(0));
             }
+        }
+
+        if (del.delegateAddress == address(0)) {
+            // If no existing delegate, delegation amount = bonded stake + amount included in this function call
+            delegationAmount = delegationAmount.add(del.bondedAmount);
         }
 
         del.delegateAddress = _to;
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -216,17 +216,20 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundInitialized
         autoClaimTokenPoolsShares
     {
-        // Caller cannot be in the unbonding state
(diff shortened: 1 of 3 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -220,12 +220,18 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        if (delegatorStatus(msg.sender) != DelegatorStatus.Bonded) {
             // New delegate
             // Set start round
             del.startRound = currentRound.add(1);
         }
 
+        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonding) {
+            // If transitioning from unbonding state
+            // make sure to zero out withdraw round
+            del.withdrawRound = 0;
```
yondonfu: If a delegator was in the unbonding state and chooses to bond again before the unbonding period is over we reset its withdraw round to 0 - a delegator should only have a withdraw round if it is in the unbonding or unbonded state

contracts/bonding/BondingManager.sol
```diff
@@ -319,7 +327,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         external
         whenSystemNotPaused
         currentRoundInitialized
-        autoClaimTokenPoolsShares
```
yondonfu: Currently, you cannot withdraw your bonded stake unless you're in the unbonded state and you can only be in the unbonded state if you previously called `unbond()` and waited through the unbonding period without bonding again. As a result, you wouldn't have any earnings that you need to claim if you're calling `unbond()` since you would not have been delegated to a transcoder during the unbonding period.

yondonfu:
Hm I think are some additional issues that need to be addresses...hold off on reviewing this!

contracts/bonding/BondingManager.sol
```diff
@@ -860,7 +865,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function updateDelegatorWithTokenPoolsShares(address _delegator, uint256 _endRound) internal {
         Delegator storage del = delegators[_delegator];
 
-        if (del.lastClaimTokenPoolsSharesRound > 0) {
+        // Only will have earnings to claim if you have a delegate
+        // If not delegated, skip the earnings claim process
+        if (del.delegateAddress != address(0)) {
```
yondonfu: This check makes sure that a delegator does not try to claim earnings since its last claim round if it does not currently have a delegate because it could not possibly have any earnings to claim in this case. We do not check if the delegate is also a registered transcoder here even though the delegate would have to be a registered transcoder in order for there to be claimable earnings - this is to account for the case in which a transcoder gets slashed, but its delegators still need to claim earnings from past rounds while they were delegated to the now slashed transcoder

contracts/bonding/BondingManager.sol
```diff
@@ -220,12 +220,18 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
         uint256 currentRound = roundsManager().currentRound();
 
-        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded) {
+        if (delegatorStatus(msg.sender) != DelegatorStatus.Bonded) {
```
yondonfu: Previously, we set the delegator's start round only if it was in the unbonded state or if it had a delegate and was delegating to someone else. Now, we set the delegator's start round no matter what unless its already bonded. If it is already bonded, we set its starts only if it is changing its delegation, otherwise we do not set its start round. This means that if you call `bond()` from the unbonding, unbonded or pending state your start round is updated.

yondonfu:
Ok now ready for review!

test/integration/Delegation.js
```diff
@@ -45,69 +49,153 @@ contract("Delegation", accounts => {
...
+        await bondingManager.withdrawStake({from: delegator1})
+        const endDelegatorBalance = await token.balanceOf(delegator1)
+        const endMinterBalance = await token.balanceOf(minterAddr)
+
+        dInfo = await bondingManager.getDelegator(delegator1)
+        currentRound = await roundsManager.currentRound()
+        assert.equal(dInfo[0], 0, "wrong bonded amount")
+        assert.equal(dInfo[5], 0, "wrong withdraw round")
+        assert.equal(endDelegatorBalance.sub(startDelegatorBalance).toNumber(), 2000, "wrong token balance increase")
+        assert.equal(startMinterBalance.sub(endMinterBalance), 2000, "wrong minter balance decrease")
```
ericxtang: Let's also check the transcoder delegatedAmount

test/integration/Delegation.js
```diff
@@ -45,69 +49,153 @@ contract("Delegation", accounts => {
...
+        await bondingManager.bond(0, transcoder1, {from: delegator2})
+        await roundsManager.mineBlocks(1)
+
+        dInfo = await bondingManager.getDelegator(delegator2)
+        currentRound = await roundsManager.currentRound()
+        assert.equal(await bondingManager.delegatorStatus(delegator2), 0, "delegator 2 should be in pending state")
+        assert.equal(dInfo[2], transcoder1, "wrong delegate")
+        assert.equal(dInfo[4], currentRound.add(1).toNumber(), "wrong start round")
+        assert.equal(dInfo[5], 0, "wrong withdraw round")
+        assert.equal(dInfo[6], currentRound.toNumber(), "wrong last claim round")
```
ericxtang: Let's also check the transcoder delegatedAmount

test/integration/Delegation.js
```diff
@@ -45,69 +49,153 @@ contract("Delegation", accounts => {
...
+        await roundsManager.initializeRound()
+
+        // Delegator 2 delegates its bonded stake which has yet to be withdrawn to transcoder 2
+        // Should not have to claim rewards and fees for any rounds and should instead just skip to the current round
+        await bondingManager.bond(0, transcoder2, {from: delegator2})
+
+        const dInfo = await bondingManager.getDelegator(delegator2)
+        const currentRound = await roundsManager.currentRound()
+        assert.equal(dInfo[2], transcoder2, "wrong delegate")
+        assert.equal(dInfo[6], currentRound.toNumber(), "wrong last claim round")
```
ericxtang: Let's also check the transcoder delegatedAmount

contracts/bonding/BondingManager.sol
```diff
@@ -220,12 +220,18 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
             // New delegate
```
ericxtang: No necessarily new delegate anymore - let's update the comment here

contracts/bonding/BondingManager.sol
```diff
@@ -245,6 +249,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             }
         }
 
+        if (del.delegateAddress == address(0)) {
+            // If no existing delegate, delegation amount = bonded stake + amount included in this function call
```
yondonfu: If a delegator unbonds it will not have an existing delegate, but will still retain its bonded amount unless it waits through the unbonding period and calls `withdrawStake()`. If the delegator calls `bond()` with an amount of 0 it should expect to delegate its bonded amount to its target delegate. Previously, this wasn't the case and `bond()` would just delegate 0 to its target delegate because `bond()` did not handle the case where a delegator still has bonded stake but does not have a delegate. Now, this change will make sure to account for the delegator's existing bonded amount

ericxtang:
🚢

PR #186, opened 2018-02-07
Upgradeable Minter with additional tests

### LINKED ISSUE #171: Upgrading the Minter
Here’s another possibility: keep the Minter as a regular contract that is not upgradeable using the delegate proxy upgrade pattern and replace transferTokenOwnership() with migrateToNewMinter() - this function is only callable by the Controller owner and transfers ownership of the token as well as the current Minter’s token balance to a new Minter. The function can do some additional sanity checks to make sure the new Minter’s controller is the same as the current Minter’s controller (and thus the same owner). The upgrade path for the Minter then becomes: 1) Deploy new Minter 2) Call migrateToNewMinter() on the current Minter 3) Register new Minter with the Controller

yondonfu:
Fixes #171 

Also label ETH and token transfer functions in the Minter that always trust their caller as "trusted".

### DIFF
--- a/contracts/token/IMinter.sol
+++ b/contracts/token/IMinter.sol
@@ -1,14 +1,23 @@
 pragma solidity ^0.4.17;
 
+import "../IController.sol";
 
+
+/**
+ * @title Minter interface
+ */
 contract IMinter {
-    event NewInflation(uint256 inflation);
-    event SetCurrentRewardTokens(uint256 mintableTokens);
+    // Events
+    event SetCurrentRewardTokens(uint256 currentMintableTokens, uint256 currentInflation);
 
+    // External functions
     function createReward(uint256 _fracNum, uint256 _fracDenom) external returns (uint256);
-    function transferTokens(address _to, uint256 _amount) external;
-    function burnTokens(uint256 _amount) external;
+    function trustedTransferTokens(address _to, uint256 _amount) external;
+    function trustedBurnTokens(uint256 _amount) external;
+    function trustedWithdrawETH(address _to, uint256 _amount) external;
     function depositETH() external payable returns (bool);
-    function withdrawETH(address _to, uint256 _amount) external;
     function setCurrentRewardTokens() external;
+
+    // Public functions
+    function getController() public view returns (IController);
 }
--- a/contracts/token/Minter.sol
+++ b/contracts/token/Minter.sol
@@ -53,6 +53,12 @@ contract Minter is Manager, IMinter {
         _;
     }
 
+    // Checks if caller is either the currently registered Minter or JobsManager
+    modifier onlyMinterOrJobsManager() {
+        require(msg.sender == controller.getContract(keccak256("Minter")) || msg.sender == controller.getContract(keccak256("JobsManager")));
+        _;
+    }
+
     /**
      * @dev Minter constructor
      * @param _inflation Base inflation rate as a percentage of current total token supply
@@ -119,6 +125,8 @@ contract Minter is Manager, IMinter {
         livepeerToken().transferOwnership(_newMinter);
         // Transfer current Minter's token balance to new Minter
         livepeerToken().transfer(_newMinter, livepeerToken().balanceOf(this));
+        // Transfer current Minter's ETH balance to new Minter
+        _newMinter.depositETH.value(this.balance)();
     }
 
     /**
@@ -167,9 +175,9 @@ contract Minter is Manager, IMinter {
     }
 
     /**
-     * @dev Deposit ETH to this contract. Only callable by JobsManager
+     * @dev Deposit ETH to this contract. Only callable by the currently registered Minter or JobsManager
      */
-    function depositETH() external payable onlyJobsManager whenSystemNotPaused returns (bool) {
+    function depositETH() external payable onlyMinterOrJobsManager whenSystemNotPaused returns (bool) {
         return true;
     }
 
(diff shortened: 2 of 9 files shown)


contracts/token/IMinter.sol
```diff
@@ -1,14 +1,23 @@
...
 
+
+/**
+ * @title Minter interface
+ */
 contract IMinter {
-    event NewInflation(uint256 inflation);
-    event SetCurrentRewardTokens(uint256 mintableTokens);
+    // Events
+    event SetCurrentRewardTokens(uint256 currentMintableTokens, uint256 currentInflation);
```
yondonfu: Collapsed `NewInflation` and `SetCurrentRewardTokens` into the same event that is fired with both the current mintable tokens and current inflation for the round so a client would only need to listen to a single event if it cares about these values as opposed to listening for two events

contracts/token/Minter.sol
```diff
@@ -99,53 +148,61 @@ contract Minter is Manager, IMinter {
...
         _to.transfer(_amount);
     }
 
-    /*
-     * @dev Deposit ETH from the JobsManager
+    /**
+     * @dev Deposit ETH to this contract. Only callable by the currently registered Minter or JobsManager
      */
-    function depositETH() external payable onlyJobsManager whenSystemNotPaused returns (bool) {
+    function depositETH() external payable onlyMinterOrJobsManager whenSystemNotPaused returns (bool) {
```
yondonfu: The modifier is changed to check if the caller is either the currently registered Minter or the JobsManager. The idea behind adding a check for the currently registered Minter was to restrict ETH transfer to this contract so that the contract cannot receive random ETH - it can only receive ETH from these whitelisted addresses if `depositETH` is called. So when we need to migrate to a new Minter, the old Minter is still registered with the Controller so it can transfer its ETH balance to the new Minter using `depositETH`.

contracts/token/Minter.sol
```diff
@@ -76,12 +104,33 @@ contract Minter is Manager, IMinter {
...
+     */
+    function migrateToNewMinter(IMinter _newMinter) external onlyControllerOwner {
+        // New Minter cannot be the current Minter
+        require(_newMinter != this);
+        // Check for null address
+        require(address(_newMinter) != address(0));
+
+        IController newMinterController = _newMinter.getController();
+        // New Minter must have same Controller as current Minter
+        require(newMinterController == controller);
```
yondonfu: These next two checks don't actually guarantee that the contract address being passed into `migrateToNewMinter` is a valid Minter contract, but more so just serve as sanity checks to make sure that the provided address at least implements the `getController` function of the Minter interface, the Controller returned matches the Controller of the old Minter and that the Controller still has the old Minter registered (might be debatable how meaningful this last check really is...)

test/unit/Minter.js
```diff
@@ -36,234 +33,319 @@ contract("Minter", accounts => {
...
-            await fixture.jobsManager.deposit({from: accounts[0], value: 100})
-            await fixture.jobsManager.setWithdrawAmount(100)
+        it("should transfer ETH to receiving address when caller is BondingManager", async () => {
+            await fixture.jobsManager.execute(minter.address, functionSig("depositETH()"), {from: accounts[1], value: 100})
+            await fixture.bondingManager.execute(minter.address, functionEncodedABI("trustedWithdrawETH(address,uint256)", ["address", "uint256"], [accounts[1], 100]))
 
-            const startBalance = web3.eth.getBalance(accounts[1])
-            await fixture.jobsManager.withdraw(true, accounts[1], {from: accounts[2]})
-            const endBalance = web3.eth.getBalance(accounts[1])
+            assert.equal(web3.eth.getBalance(minter.address), 0, "wrong minter balance")
```
ericxtang: Should also check the balance for accounts[1]

test/unit/Minter.js
```diff
@@ -36,234 +33,319 @@ contract("Minter", accounts => {
...
+        })
+
+        it("should transfer ETH to receiving address when caller is JobsManager", async () => {
+            await fixture.jobsManager.execute(minter.address, functionSig("depositETH()"), {from: accounts[1], value: 100})
+            await fixture.jobsManager.execute(minter.address, functionEncodedABI("trustedWithdrawETH(address,uint256)", ["address", "uint256"], [accounts[1], 100]))
 
-            assert.equal(endBalance.sub(startBalance).toNumber(), 100, "balance did not update correctly")
+            assert.equal(web3.eth.getBalance(minter.address), 0, "wrong minter balance")
```
ericxtang: should also check the balance for accounts[1]

contracts/token/Minter.sol
```diff
@@ -76,12 +104,33 @@ contract Minter is Manager, IMinter {
...
-        livepeerToken().transferOwnership(_newOwner);
-    }
-
-    /*
-     * @dev Create reward based on a fractional portion of the mintable tokens and redistributable funds for a round
+    /**
+     * @dev Migrate to a new Minter by transferring ownership of the token as well
+     * as the current Minter's token balance to the new Minter
+     * @param _newMinter Address of new Minter
+     */
```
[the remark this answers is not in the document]
yondonfu: Yeah that's probably a good idea to cover the edge case where there is a gap between calling `migrateToNewMinter()` and registering the new Minter with the Controller - if anyone tries withdrawing funds or creating inflationary rewards in that gap those calls will fail

ericxtang:
🚢

PR #190, opened 2018-02-12
Update BondingManager unit test. Small fixes in BondingManager.sol

### LINKED ISSUE #132: Add a guard to check if delegator is trying to claim rewards/fees for too many rounds
The `autoClaimTokenPoolsShares` modifier and the `claimTokenPoolsShares` function iterate from a delegator's `lastClaimTokenPoolsSharesRound` through an end round (the current round in the case of `autoClaimTokenPoolsShares`). If the number of rounds to iterate through becomes too large, both of these operations will run out of gas. The client interacting with the protocol can check how many rounds there are to iterate through and break the total number of rounds into smaller portions to iterate through, but it could also be helpful to add a guard for `autoClaimTokenPoolsShares` and `claimTokenPoolsShares` to revert if the number of rounds to iterate through is exceedingly large. This would be an extra defense on the contract level to protect users from wasting gas on trying to claim rewards/fees for too many rounds only to have the transaction to fail because the transaction ran out of gas.

### LINKED ISSUE #187: Explicitly mark visibility in functions and state variables
Going through the Consensys audit checklist I found a few variables that didn't have their visibility explicitly marked:

Controller.registry
BondingManager.delegators
BondingManager.transcoders
BondingManager.totalBonded
BondingManager.transcoderPools

yondonfu:
Fixes #187 
Fixes #132 

Updated BondingManager unit test. Found some small issues in the BondingManager contract in the process - fixes described in line.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -13,6 +13,10 @@ import "zeppelin-solidity/contracts/math/Math.sol";
 import "zeppelin-solidity/contracts/math/SafeMath.sol";
 
 
+/**
+ * @title BondingManager
+ * @dev Manages bonding, transcoder and rewards/fee accounting related operations of the Livepeer protocol
+ */
 contract BondingManager is ManagerProxyTarget, IBondingManager {
     using SafeMath for uint256;
     using SortedDoublyLL for SortedDoublyLL.Data;
@@ -46,21 +50,21 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         uint256 delegatedAmount;                 // The amount of tokens delegated to the delegator
         uint256 startRound;                      // The round the delegator transitions to bonded phase and is delegated to someone
         uint256 withdrawRound;                   // The round at which a delegator can withdraw
-        uint256 lastClaimTokenPoolsSharesRound;  // The last round during which the delegator claimed its share of a transcoder's reward and fee pools
+        uint256 lastClaimRound;                  // The last round during which the delegator claimed its earnings
     }
 
     // The various states a delegator can be in
     enum DelegatorStatus { Pending, Bonded, Unbonding, Unbonded }
 
     // Keep track of the known transcoders and delegators
-    mapping (address => Delegator) delegators;
-    mapping (address => Transcoder) transcoders;
+    mapping (address => Delegator) private delegators;
+    mapping (address => Transcoder) private transcoders;
 
     // Keep track of total bonded tokens
-    uint256 totalBonded;
+    uint256 private totalBonded;
 
     // Candidate and reserve transcoders
-    SortedDoublyLL.Data transcoderPool;
+    SortedDoublyLL.Data private transcoderPool;
 
     // Represents the active transcoder set
     struct ActiveTranscoderSet {
@@ -90,34 +94,20 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         _;
     }
 
-    // Automatically claim token pools shares from lastClaimTokenPoolsSharesRound through the current round
-    modifier autoClaimTokenPoolsShares() {
-        updateDelegatorWithTokenPoolsShares(msg.sender, roundsManager().currentRound());
+    // Automatically claim earnings from lastClaimRound through the current round
+    modifier autoClaimEarnings() {
+        updateDelegatorWithEarnings(msg.sender, roundsManager().currentRound());
         _;
     }
 
+    /**
+     * @dev BondingManager constructor. Only invokes constructor of base Manater contract with provided Controller address
+     * @param _controller Address of Controller that this contract will be registered with
+     */
     function BondingManager(address _controller) public Manager(_controller) {}
 
-    /*
-     * @dev Batch set protocol parameters. Only callable by the controller owner
-     * @param _unbondingPeriod Rounds between unbonding and possible withdrawl
-     * @param _numTranscoders Max number of registered transcoders
-     * @param _numActiveTranscoders Number of active transcoders
-     */
-    function setParameters(uint64 _unbondingPeriod, uint256 _numTranscoders, uint256 _numActiveTranscoders) external onlyControllerOwner {
-        unbondingPeriod = _unbondingPeriod;
-
-        // Max number of transcoders must be greater than or equal to number of active transcoders
-        require(_numTranscoders >= _numActiveTranscoders);
-
-        transcoderPool.setMaxSize(_numTranscoders);
-        numActiveTranscoders = _numActiveTranscoders;
-
-        ParameterUpdate("all");
-    }
-
-    /*
-     * @dev Set unbonding period. Only callable by the controller owner
+    /**
+     * @dev Set unbonding period. Only callable by Controller owner
      * @param _unbondingPeriod Rounds between unbonding and possible withdrawal
      */
     function setUnbondingPeriod(uint64 _unbondingPeriod) external onlyControllerOwner {
@@ -126,8 +116,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         ParameterUpdate("unbondingPeriod");
     }
 
-    /*
-     * @dev Set max number of registered transcoders. Only callable by the controller owner
+    /**
+     * @dev Set max number of registered transcoders. Only callable by Controller owner
      * @param _numTranscoders Max number of registered transcoders
      */
     function setNumTranscoders(uint256 _numTranscoders) external onlyControllerOwner {
@@ -139,8 +129,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         ParameterUpdate("numTranscoders");
     }
 
-    /*
-     * @dev Set number of active transcoders. Only callable by the controller owner
+    /**
+     * @dev Set number of active transcoders. Only callable by Controller owner
      * @param _numActiveTranscoders Number of active transcoders
      */
     function setNumActiveTranscoders(uint256 _numActiveTranscoders) external onlyControllerOwner {
@@ -152,11 +142,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         ParameterUpdate("numActiveTranscoders");
     }
 
-    /*
+    /**
      * @dev The sender is declaring themselves as a candidate for active transcoding.
      * @param _blockRewardCut % of block reward paid to transcoder by a delegator
      * @param _feeShare % of fees paid to delegators by a transcoder
-     * @param _pricePerSegment Price per segment (denominated in LPT units) for a stream
+     * @param _pricePerSegment Price per segment (denominated in Wei) for a stream
      */
     function transcoder(uint256 _blockRewardCut, uint256 _feeShare, uint256 _pricePerSegment)
         external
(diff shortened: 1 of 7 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -186,14 +176,12 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         if (transcoderStatus(msg.sender) == TranscoderStatus.NotRegistered) {
             if (transcoderPool.isFull()) {
                 address lastTranscoder = transcoderPool.getLast();
+                // Must have more stake than the last transcoder in the pool
+                require(delegatedAmount > transcoderPool.getKey(lastTranscoder));
```
yondonfu: Previously, if the caller did not have enough delegated stake to join the transcoder pool (the transcoder pool is full and the caller's delegated stake <= the delegated stake of the transcoder at the bottom of the pool), this function would just fail silently (fields are updated, but the transcoder would not actually be registered in the transcoder pool). Now, this function reverts if the caller does not have enough delegated stake to join the transcoder pool

contracts/bonding/BondingManager.sol
```diff
@@ -249,43 +238,39 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             }
         }
 
-        if (del.delegateAddress == address(0)) {
-            // If no existing delegate, delegation amount = bonded stake + amount included in this function call
-            delegationAmount = delegationAmount.add(del.bondedAmount);
-        }
-
+        // Delegation amount must be > 0 - cannot delegate to someone without having bonded stake
+        require(delegationAmount > 0);
```
yondonfu: I don't think there is any reason for someone to have a delegate unless they also have bonded stake, so this check makes sure the delegation amount (sum of current bonded stake and provided amount) is greater than 0

contracts/bonding/BondingManager.sol
```diff
@@ -90,34 +94,20 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
     }
 
-    function BondingManager(address _controller) public Manager(_controller) {}
-
-    /*
-     * @dev Batch set protocol parameters. Only callable by the controller owner
-     * @param _unbondingPeriod Rounds between unbonding and possible withdrawl
-     * @param _numTranscoders Max number of registered transcoders
-     * @param _numActiveTranscoders Number of active transcoders
+    /**
```
j0sh: Curious, can we auto-generate docs from this style of function documentation?

contracts/bonding/BondingManager.sol
dob: I *thought* that it was ok if this transaction failed silently because the caller would still be registered candidate transcoder so they would be eligible to receive other stake and eventually make it into the pool, even if their initial transaction didn't have enough stake to make it into the pool.

If we revert here, then it's impossible to get others to delegate towards you (sponsor you) into the pool since you need enough stake in possession to make it on your own.

Or was I misunderstanding how it worked before?

contracts/bonding/BondingManager.sol
```diff
@@ -566,80 +541,76 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
-    /*
-     * @dev Returns pending bonded stake for a delegator. Includes reward pool shares since lastClaimTokenPoolsSharesRound
+    /**
+     * @dev Returns pending bonded stake for a delegator from its lastClaimRound through an end round
      * @param _delegator Address of delegator
+     * @param _endRound The last round to compute pending stake from
      */
-    function pendingStake(address _delegator) public view returns (uint256) {
+    function pendingStake(address _delegator, uint256 _endRound) public view returns (uint256) {
```
yondonfu: `pendingStake` and `pendingFees` can be viewed as the read-only counterparts to `claimEarnings`. While `claimEarnings` will compute the claimable rewards/fees for each round and credit the caller with those values, `pendingStake` and `pendingFees` just compute claimable rewards/fees for each round and what the caller's stake/collected fees would be if it called `claimEarnings`. This `endRound` parameter just allows the caller to configure which round to compute the pending rewards and fees up through

contracts/bonding/BondingManager.sol
```diff
@@ -814,21 +785,23 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
-    /*
+    /**
      * @dev Remove transcoder
      */
     function resignTranscoder(address _transcoder) internal {
         uint256 currentRound = roundsManager().currentRound();
         if (activeTranscoderSet[currentRound].isActive[_transcoder]) {
-            // If transcoder is active for the round set it as inactive
+            // Decrease total active stake for the round
```
yondonfu: Whenever we resign a transcoder we can use this function which will not only remove the transcoder from the pool, but also check if the transcoder is active for the round and if so, set decrease the total active stake for the current round and set the transcoder as inactive

contracts/bonding/BondingManager.sol
```diff
@@ -860,21 +833,28 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
         // Only will have earnings to claim if you have a delegate
         // If not delegated, skip the earnings claim process
         if (del.delegateAddress != address(0)) {
+            // Cannot claim earnings for more than 20 rounds
+            // This is an arbitrary number to cause transactions to fail early if
+            // we know they will require too much gas to loop through all the necessary rounds to claim earnings
+            // The user should instead manually invoke `claimEarnings` to split up the claiming process
+            // across multiple transactions
+            require(_endRound.sub(del.lastClaimRound) <= 20);
```
dob: is 20 arbitrary or somewhat empirically observed as a reasonable value?

yondonfu: So this number is pretty arbitrary and it might be debatable whether this early failure is *that* useful - the idea is that if the number of rounds to claim through is known to be too big, `claimEarnings` and any bonding function with the `autoClaimEarnings` modifier will fail early instead of trying to loop through all the rounds thereby consuming all the necessary gas. This can always be determined off-chain, but this check just adds an extra safeguard in case anyone does call a function that tries to claim earnings through too many rounds

contracts/bonding/BondingManager.sol
yondonfu: Yeah the small updates to the format of these comments were so that they adhere to the [Ethereum Natural Specification](https://github.com/ethereum/wiki/wiki/Ethereum-Natural-Specification-Format) and there are tools for auto-generating docs based upon that format

contracts/bonding/BondingManager.sol
yondonfu: The function actually uses the caller's delegated amount (which includes the amount other people have delegated to the caller) as opposed to its bonded amount (which is just the amount that the caller has bonded by itself) so it would still be possible for other people to "sponsor" you so you could have enough delegated stake to join the pool. Once you have enough people sponsoring you such that you have a sufficient delegated stake to join the pool, you can call `transcoder` to evict the transcoder with the lowest delegated stake in the pool and join the pool.

Before this change, the transaction would fail silently such that you end up not being in the transcoder pool (and thus not deemed a registered transcoder because `transcoderStatus`  considers you registered only if you are in the transcoder pool), but you still updated your reward cut, fee share, price, etc.

contracts/bonding/BondingManager.sol
yondonfu: Its arbitrary right now, but I can do some gas testing to see what a reasonable number of rounds to claim through is...

yondonfu:
@j0sh Good point re: breaking up the changes into individual commits (I've also been trying to create smaller PRs). I got a little carried away updating the unit test and finding small issues so they ended up getting lumped into a single commit, but I'll try to be more cognizant of the size/content of the commit going forward 😅

j0sh:
Indeed, especially with writing tests -- it's easy (and quite fun!) to keep the momentum going once started, so that's certainly not to be discouraged 😄  For what it's worth, `git add -p` is a lifesaver for teasing apart large chunks of diffs.

test/unit/BondingManager.js
```diff
@@ -32,742 +47,1417 @@ contract("BondingManager", accounts => {
...
+            await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound)
 
-        // Mock distribute fees params
-        const fees = 300
-        const jobCreationRound = 7
-        const transcoderTotalStake = 4000
+            await bondingManager.bond(1000, transcoder, {from: transcoder})
+            await bondingManager.transcoder(5, 10, 1, {from: transcoder})
+
+            await fixture.roundsM
```
ericxtang: Should check finder got its fee

test/unit/BondingManager.js
```diff
@@ -32,742 +47,1417 @@ contract("BondingManager", accounts => {
...
+            await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound + 1 + unbondingPeriod.toNumber())
+            await bondingManager.withdrawStake({from: delegator})
 
-            const expFeePool = fees
-            const tokenPools = await bondingManager.getTranscoderTokenPoolsForRound(tAddr, jobCreationRound)
-            const feePool = tokenPools[1]
-            assert.equal(feePool, expFeePool, "transcoder fee pool incorrect")
+            const dInfo = await bondingManager.getDelegator(delegator)
+            assert.equal(dInfo[0], 0, "should set bondedAmount to 0")
+            assert.equal(dInfo[5], 0, "should set withdrawRound to 0")
```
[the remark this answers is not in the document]
yondonfu: So we can't actually check the token balance in these unit tests because the LivepeerToken contract dependency is mocked out such that the mock doesn't keep track of any actual data - instead whenever the function under test relies on a value returned by the dependency, we just tell the mock to return a specific value. The idea was that for each unit test, we only test state changes specific to the contract under test and leave any side effects created by a function call for a contract to be tested in an integration test where the dependencies are not mocked out.

test/unit/BondingManager.js
```diff
@@ -32,742 +47,1417 @@ contract("BondingManager", accounts => {
...
+            describe("caller is changing delegate", () => {
+                it("should set startRound to next round", async () => {
+                    await bondingManager.bond(0, transcoder1, {from: delegator})
 
-            await fixture.jobsManager.setBondingManager(bondingManager.address)
-            await fixture.roundsManager.setBondingManager(bondingManager.address)
-            await fixture.roundsManager.setCurrentRoundInitialized(true)
-            await fixture.roundsManager.setCurrentRound(5)
-            await fixture.token.setApproved(true)
+                    assert.equal((await bondingManager.getDelegator(delegator))[4], currentRound + 2, "wrong startRound")
```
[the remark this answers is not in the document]
yondonfu: I think this is tested for on L314 https://github.com/livepeer/protocol/blob/7c845e772d0b20380813edcef255d6526325e919/test/unit/BondingManager.js#L314

test/unit/BondingManager.js
```diff
@@ -32,742 +47,1417 @@ contract("BondingManager", accounts => {
...
+
+        it("should withdraw caller's fees", async () => {
+            await bondingManager.claimEarnings(currentRound + 1, {from: transcoder0})
+            assert.isAbove((await bondingManager.getDelegator(transcoder0))[1], 0, "caller should have non-zero fees")
+
+            await bondingManager.withdrawFees({from: transcoder0})
+
+            const dInfo = await bondingManager.getDelegator(transcoder0)
+            assert.equal(dInfo[6], currentRound + 1, "should set caller's lastClaimRound")
+            assert.equal(dInfo[1], 0, "should set caller's fees to zero")
```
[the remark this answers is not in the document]
yondonfu: Kind of along the lines of my comment about testing state changes specific to the contract under test - the actual ETH transfer is normally performed by a call to `trustedWithdrawETH` on the Minter, but since the Minter is mocked out in this unit test, that function call doesn't actually do anything. Similar to with the LivepeerToken contract, the idea was that actual ETH movement side effects that rely on the Minter could be tested in an integration test and this unit test would just focus on how the BondingManager state changes

dob:
This is such a great set of tests for the bonding manager. Thanks for pulling this together before the audit. While it was a little too massive to line by line on the testing conditions, reading through each scenario gave me confidence that we're well covered for regressions in far more areas than we were before. I approve the merge as Eric's comments are addressed.

yondonfu:
Left some comments re: checking for things such as token balances and ETH balances that the tests currently do not check for because these interactions depend on external contract dependencies and the current unit test approach is to isolate state changes for the contract under test and leave side effects for integration tests. I think this makes it a little easier to test specific state changes for a single contract, but let me know your thoughts!

ericxtang:
@yondonfu I think your approach makes sense here.  It just means we need have a place to run the integration tests with the additional dependencies.  It may feel a little redundant, but I think worth doing soon.

contracts/bonding/BondingManager.sol
jozanza: Agreed. @yondonfu created an issue #191 where we can maybe try and figure out what that could mean as far as implementation goes.

PR #192, opened 2018-02-13
Transcoder price matching with a price floor during round lock period

### LINKED ISSUE #188: Proposal - On chain availability + Round lock price matching solution to availability/job assignment problem
## The problems

1. Currently transcoders have no way to indicate they're unavailable. This creates bad broadcaster experience.
2. Even if we had a way to indicate they were proactively unavailable, what is their incentive to call this?
3. The transcoder with the lowest price in a round, essentially has right of first refusal on all jobs. Even if other transcoders are priced only 1 wei higher, they may not get a single job. This creates a "game" in pricing to try to end up as the lowest priced transcoder. And it leads to centralization of all jobs processed by 1 or a few nodes.

## A proposed solution

1. On chain availability counter-based tracking. Set a network parameter `MaxConcurrency` with a value such as `4`. When a transcoder gets assigned a job, this parameter reduces. *(Challenge or potential blocker, we don't know when they are assigned a job on chain.)* When it hits zero, they are unavailable until they indicate availability.
   1. They also have the ability to proactively mark as unavailable at any time
   1. This solves for letting the transcoder go offline, and if they happen to go offline they'll eventually be made inactive.
    1. It also ensures that if they reach the `MaxConcurrency` then the network can react, adjust price upward, and pass work to the next available transcoder.
    1. If you want your work definitely accepted by a transcoder during a time of high usage of the network, set your price higher so that the pool of potentially available transcoders is bigger. If you only enable 1, then your transaction may fail if someone beats you in the race to claim that transcoder's final concurrency spot.
2. Downward price adjustment during Round Lock Period allowed. During this period the lowest price transcoder's price is locked. Transcoders can adjust their price down, with a limit of matching the lowest price transcoders.
   1. This allows the pool of people willing to do the work at the lowest network price to grow, encouraging decentralization and throughput at that price.
   1. If the price is too low for people to match, that's fine...maybe they'll converge around the second lowest price. And this will become the price when `MaxConcurrency` is exceeded or if the low price transcoder is inactive or goes offline.
   1. *Note, one weakness here is that it might be hard to converge on the second lowest price, as this becomes another "game" in which people may try to change at the last second to gain that position.*

### Implementation

This is clearly not fully spec'ed out, but here are some ideas:

* Availability is a map or array maintained on chain indicating the counter for each active transcoder by index.
* This is updated by the `Job(), Claim(), Available(), and Unavailable()` methods
* When a new job is created, an availability bitmap is copied into the job to be used by `electActiveTranscoder()`
* `transcoder()` function is updated to support downward price setting during the lock period if this is not already the low priced transcoder.

### Questions

* What are the effects of an arbitrarily low `MaxConcurrency` in the early days? 
   * Probably that work would be more likely to get distributed around the network and fees may be a little higher on average than they could be if all work went to the cheapest.

* Would there be games played around pricing to try and sneak into certain positions?
   * I guess for the alpha I'm less worried about this in the early days and more interested in the social + long term alignment of transcoders. I think them aligning towards the cheapest price is actually a very good outcome. And if an attacker spends stake to set an arbitrarily low price, well the rest will align around the second lowest and our client defaults will set the price a little higher such that it's likely the job gets assigned to someone available anyway.

* The blocker around not determining who a job was assigned to in the `Job()` transaction is a concern. If we wait until the `claim()` function to decrement their availability counter, then this doesn't help defend against the case where they are actually offline.

yondonfu:
Fixes #188 

This implements the ability for registered transcoders to price match the lowest price set by any registered transcoder during the round lock period.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -165,41 +165,76 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         whenSystemNotPaused
         currentRoundInitialized
     {
-        // Only callable if current round is not locked
-        require(!roundsManager().currentRoundLocked());
-        // Block reward cut must be a valid percentage
-        require(MathUtils.validPerc(_blockRewardCut));
-        // Fee share must be a valid percentage
-        require(MathUtils.validPerc(_feeShare));
-
+        Transcoder storage t = transcoders[msg.sender];
         Delegator storage del = delegators[msg.sender];
 
-        // Must have a non-zero amount bonded to self
-        require(del.delegateAddress == msg.sender && del.bondedAmount > 0);
+        if (roundsManager().currentRoundLocked()) {
+            // If it is the lock period of the current round
+            // the lowest price previously set by any transcoder
+            // becomes the price floor and the caller can lower its
+            // own price to a point greater than or equal to the price floor
 
-        Transcoder storage t = transcoders[msg.sender];
-        t.pendingBlockRewardCut = _blockRewardCut;
-        t.pendingFeeShare = _feeShare;
-        t.pendingPricePerSegment = _pricePerSegment;
+            // Caller must already be a registered transcoder
+            require(transcoderStatus(msg.sender) == TranscoderStatus.Registered);
 
-        uint256 delegatedAmount = del.delegatedAmount;
+            // Iterate through the transcoder pool to find the price floor
+            // Since the caller must be a registered transcoder, the transcoder pool size will always at least be 1
+            // Thus, we can safely set the initial price floor to be the pendingPricePerSegment of the first
+            // transcoder in the pool
+            address currentTranscoder = transcoderPool.getFirst();
+            uint256 priceFloor = transcoders[currentTranscoder].pendingPricePerSegment;
+            for (uint256 i = 0; i < transcoderPool.getSize(); i++) {
+                if (transcoders[currentTranscoder].pendingPricePerSegment < priceFloor) {
+                    priceFloor = transcoders[currentTranscoder].pendingPricePerSegment;
+                }
 
-        // Check if transcoder is not already registered
-        if (transcoderStatus(msg.sender) == TranscoderStatus.NotRegistered) {
-            if (transcoderPool.isFull()) {
-                address lastTranscoder = transcoderPool.getLast();
-                // Must have more stake than the last transcoder in the pool
-                require(delegatedAmount > transcoderPool.getKey(lastTranscoder));
-
-                transcoderPool.remove(lastTranscoder);
-
-                TranscoderEvicted(lastTranscoder);
+                currentTranscoder = transcoderPool.getNext(currentTranscoder);
             }
 
-            transcoderPool.insert(msg.sender, delegatedAmount, address(0), address(0));
-        }
+            // Provided pricePerSegment must greater than or equal to the price floor and
+            // less than or equal to the previously set pricePerSegment
+            require(_pricePerSegment >= priceFloor && _pricePerSegment <= t.pendingPricePerSegment);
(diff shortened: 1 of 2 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -165,41 +165,76 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            // Caller must already be a registered transcoder
+            require(transcoderStatus(msg.sender) == TranscoderStatus.Registered);
+
+            // Iterate through the transcoder pool to find the price floor
+            // Since the caller must be a registered transcoder, the transcoder pool size will always at least be 1
+            // Thus, we can safely set the initial price floor to be the pendingPricePerSegment of the first
+            // transcoder in the pool
+            address currentTranscoder = transcoderPool.getFirst();
+            uint256 priceFloor = transcoders[currentTranscoder].pendingPricePerSegment;
+            for (uint256 i = 0; i < transcoderPool.getSize(); i++) {
```
dob: I wonder if it's worth it to track the priceFloor per round (if it's already been calculated) in a data structure so that we don't have to loop through to determine it every time.

contracts/bonding/BondingManager.sol
```diff
@@ -165,41 +165,76 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        // Must have a non-zero amount bonded to self
-        require(del.delegateAddress == msg.sender && del.bondedAmount > 0);
+        if (roundsManager().currentRoundLocked()) {
+            // If it is the lock period of the current round
+            // the lowest price previously set by any transcoder
+            // becomes the price floor and the caller can lower its
+            // own price to a point greater than or equal to the price floor
+
+            // Caller must already be a registered transcoder
+            require(transcoderStatus(msg.sender) == TranscoderStatus.Registered);
```
dob: should we also require that `_feeShare` and `_blockRewardCut` are not different than the pending values? I'm not sure...either the values won't update silently which could be weird, or else they risk wasting gas if they don't realize they have to set these exactly the same.

contracts/bonding/BondingManager.sol
yondonfu: The tricky part with tracking the priceFloor is computing the new priceFloor if the transcoder with the previous priceFloor resigns. I think we would have to maintain a data structure (likely a sorted one) that can always return the correct priceFloor even if transcoders enter/leave the transcoder pool during a round which would require a bunch of additional storage writes to handle this case

contracts/bonding/BondingManager.sol
yondonfu: Hm yeah I do think that ignoring the values silently is a bit weird...I would lean towards adding the require checks for `feeShare` and `blockRewardCut` to make the intent of the function call clear: "I am leaving my feeShare and blockRewardCut unchanged by providing the same values as the pending values and changing the pricePerSegment" instead of "I am  providing arbitrary values for feeShare and blockRewardCut that will be ignored". I think the defenses against a user accidentally providing incompatible feeShare and blockRewardCut values can be added in the client

yondonfu:
Updated with require checks that provided blockRewarCut and feeShare values during lock period match the pending values that were previously set by the registered transcoder

contracts/bonding/BondingManager.sol
dob: Got it. Ok let's leave this for now. I can envision adding additional accounting to handle this case, such as a bool mapping that returns false if the floor isn't locked for the round, and then sets this to false if a transcoder resigns. But I also don't like the idea of adding all this special purpose accounting all over the place. I'd rather keep it simple than optimize for gas savings at this point.

dob:
Explicitly approving the updated require checks :ship:

PR #195, opened 2018-02-13
Update rest of unit tests

yondonfu:
Update the rest of the unit tests to use the new unit test approach of mocking out every external contract dependency using the GenericMock contract and the unit test fixture. This update allows us to delete the old mock contract dependencies.


package.json
```diff
@@ -40,7 +40,7 @@
     "bignumber.js": "^4.0.1",
     "eslint": "^4.3.0",
     "eslint-config-google": "^0.9.1",
-    "ethereumjs-abi": "^0.6.4",
+    "ethereumjs-abi": "github:yondonfu/ethereumjs-abi#abi-offset-fixed-arrays",
```
yondonfu: There is an ABI encoding bug in the ethereumjs-abi package - I've opened a PR, but until that is merged in this will install the branch with the fix on my fork

PR #196, opened 2018-02-13
Some other additional JobsManager fixes

### LINKED ISSUE #182: Transcoders can leave money on the table if they submit a transcode claim right before a job's end block
We require the job to be active in verify(). If the transcoder calls claimWork() in the last block before job.endBlock, there is a chance one of the segments should be verified, and the transcoder will not have an opportunity to do so because the job will have ended after you call claimWork().  This means you’d always have to “leave some money on the table” during the last few rounds of a job to protect yourself as a transcoder.

Yondon: Ah...yeah good point. Perhaps we remove the requirement for a job to be active for verify() so if a transcoder calls claimWork() right before the job’s end block, it can still verify any challenged segments for that last transcode claim. The additional implication is that a transcoder can call verify() for an inactive job, but I don’t think that is meaningful because the claimWork() is the function call that can actually move funds.

### LINKED ISSUE #194: Additional checks when receiving verification result
Eric's comment: In `receiveVerification`, we don’t check for the verification has already been done, or the claim has already been slashed.  Should we do that?  I think it depends on our assumption - if we can COMPLETELY trust the compute oracles, then we’ll be fine.  If we want to minimize the attack surface from a mal-functioning or malicious oracle, we could mark the claim as Slashed.

These additional checks can be placed in `receiveVerification` in JobsManager or `__callback` in LivepeerVerifier

yondonfu:
Fixes #194 
Fixes #182

### DIFF
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -354,8 +354,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         Job storage job = jobs[_jobId];
         Claim storage claim = job.claims[_claimId];
 
-        // Job cannot be inactive
-        require(jobStatus(_jobId) != JobStatus.Inactive);
         // Sender must be elected transcoder
         require(job.transcoderAddress == msg.sender);
 
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -451,11 +451,22 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         external
         whenSystemNotPaused
         onlyVerifier
+        jobExists(_jobId)
     {
-        address transcoder = jobs[_jobId].transcoderAddress;
+        Job storage job = jobs[_jobId];
+        Claim storage claim = job.claims[_claimId];
+        // Claim must not be slashed
+        require(claim.status != ClaimStatus.Slashed);
+        // Segment must have been submitted for verification
+        require(claim.segmentVerifications[_segmentNumber]);
+
+        address transcoder = job.transcoderAddress;
 
         if (!_result) {
+            // Refund broadcaster
             refundBroadcaster(_jobId);
+            // Set claim as slashed
+            claim.status = ClaimStatus.Slashed;
             // Protocol slashes transcoder for failing verification (no finder)
             bondingManager().slashTranscoder(transcoder, address(0), failedVerificationSlashAmount, 0);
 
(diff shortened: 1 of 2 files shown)


contracts/jobs/JobsManager.sol
```diff
@@ -451,11 +449,22 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         external
         whenSystemNotPaused
         onlyVerifier
+        jobExists(_jobId)
     {
-        address transcoder = jobs[_jobId].transcoderAddress;
+        Job storage job = jobs[_jobId];
```
yondonfu: Assuming that the verifier contract being used is the LivepeerVerifier, these checks should prevent a rogue whitelisted solver from triggering this function via `__callback` in LivepeerVerifier` to slash a transcoder. However, there aren't any checks preventing a rogue whitelisted solver from triggering this function via `__callback` with result set to true, but in that case the only thing that happens is the `PassedVerification` event is fired so I think that should be ok

contracts/jobs/JobsManager.sol
```diff
@@ -354,8 +354,6 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
         Job storage job = jobs[_jobId];
         Claim storage claim = job.claims[_claimId];
 
-        // Job cannot be inactive
-        require(jobStatus(_jobId) != JobStatus.Inactive);
```
yondonfu: I can't immediately think of any issues that might arise from a transcoder being able to call `verify()` for an inactive job - the benefit here is that it solves the problem of a transcoder leaving money on the table if the current block is very close to the job's end block and it wants to claim as many segments as possible by waiting until the very end - this would allow the transcoder to submit a claim at the very end, and then submit for verification after the job's end block

dob: Is it weird if they call verify after the slashing period has ended? So for example call `claim()` at the end of the round, and then not call `verify()` for 1000 blocks? I assume they would already be slashed but would any code here screw up the state if executed later?

contracts/jobs/JobsManager.sol
yondonfu: If the claim does not exist, `Claim storage claim = job.claims[_claimId]` should throw with an out of bounds error https://github.com/livepeer/protocol/pull/196/files#diff-8f91d5b27bb5716159a6dc1367d9dffbR572

ericxtang:
Looks good to me.  Feel free to 🚢 after clarifying @dob's question.

contracts/jobs/JobsManager.sol
yondonfu: I think it is assumed that if the transcoder misses verification because it does not call `verify()` for a long time that a watcher will slash the transcoder during the verification slashing period. If the transcoder is not slashed, calls `verify()` many blocks later and passes verification, nothing should happen because a successful verification only fires an event

dob: Ah, I see. So since a successful verificaiton is equivalent to a lack of slashing during the slash period, there's actually no benefit to calling verify after the fact. Sounds good. :ship:

PR #200, opened 2018-02-14
Allow transcoders with insufficient delegated stake to set rewardCut, feeShare, price

### LINKED ISSUE #191: On-chain price signaling for a user that is campaigning to become a transcoder
At the moment, we prevent users from registering as a transcoder if the transcoder pool is full and they have insufficient delegated stake to evict the transcoder in the pool with the lowest delegated stake. As a result, these users have to campaign off-chain to have other users delegate tokens towards them (effectively as "sponsors") in order to build up enough delegated stake to eventually join the transcoder pool. 

A question is whether the protocol should allow these users to invoke the `transcoder` function to just set the `rewardCut`, `feeShare` and `pricePerSegment` that the user *would* charge if it was a registered transcoder as a form of on-chain price signaling. The benefits of this are 1) increased transparency 2) queryable on-chain data to build a dapp interface for delegating to users that are campaigning to become transcoders. The potential disadvantages of this are 1) data management is less clean because you end up with transcoder data structures that have some fields populated, but are not registered while the alternative is to ensure that all fields in a transcoder data structure are zeroed out if the user is not registered.

yondonfu:
Closes #191 

Allows a user campaigning to become a transcoder to still invoke `transcoder` to set rewardCut, feeShare and pricePerSegment even if the transcoder pool is full and they do not have sufficient delegated stake to join the pool

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -203,12 +203,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
             t.pendingPricePerSegment = _pricePerSegment;
 
-            TranscoderUpdate(msg.sender, t.pendingRewardCut, t.pendingFeeShare, _pricePerSegment);
+            TranscoderUpdate(msg.sender, t.pendingRewardCut, t.pendingFeeShare, _pricePerSegment, true);
         } else {
             // It is not the lock period of the current round
             // Caller is free to change rewardCut, feeShare, pricePerSegment as it pleases
             // If caller is not a registered transcoder, it can also register and join the transcoder pool
             // if it has sufficient delegated stake
+            // If caller is not a registered transcoder and does not have sufficient delegated stake
+            // to join the transcoder pool, it can change rewardCut, feeShare, pricePerSegment
+            // as information signals to delegators in an effort to camapaign and accumulate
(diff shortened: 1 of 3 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -226,20 +230,25 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+                if (!transcoderPool.isFull()) {
+                    // If pool is not full add new transcoder
+                    transcoderPool.insert(msg.sender, delegatedAmount, address(0), address(0));
+                } else {
                     address lastTranscoder = transcoderPool.getLast();
-                    // Must have more stake than the last transcoder in the pool
-                    require(delegatedAmount > transcoderPool.getKey(lastTranscoder));
 
-                    transcoderPool.remove(lastTranscoder);
+                    if (delegatedAmount > transcoderPool.getKey(lastTranscoder)) {
```
dob: Should this be `delegatedAmount` or `delegatedAmount + t.<current-bonded-stake>`? Because they may already have some amount delegated and this new delegatedAmount could push them over the edge?

dob: Ah, nevermind this. There is no additional amount bonded in this function so we don't need to do additional accounting.

PR #201, opened 2018-02-14
Add coverage testing for contracts

yondonfu:
Uses [solidity-coverage](https://github.com/sc-forks/solidity-coverage) to generate coverage reports for contract tests.


yondonfu:
Ready for review

package.json
```diff
@@ -12,11 +12,14 @@
     "solium": "./node_modules/.bin/solium --dir ./contracts",
     "ganache": "./node_modules/.bin/ganache-cli -l 0x663BE0",
     "compile": "rm -rf build/contracts && truffle compile --all",
+    "test:coverage": "SOLIDITY_COVERAGE=true bash scripts/run_ganache.sh && ./node_modules/.bin/solidity-coverage || true",
```
yondonfu: solidity-coverage uses a forked version of ganache-cli/testrpc called testrpc-sc which is not up to date with the beta version of ganache-cli we are using that includes a fix for intermittent crashes during testing and a change in how block hashes are retrieved. As a result, 2 tests fail when running unit tests with coverage enabled. As a workaround, we run the unit tests once with coverage enabled just to get the coverage report for coveralls and a second time to actually make sure they all pass. Maybe there is a better way to do this until testrpc-sc is up to date with the ganache-cli beta....

PR #202, opened 2018-02-27
Additional test cases for uncovered branches

yondonfu:
Added some additional test cases for uncovered code branches. Also renamed `TokenPools.sol` to `EarningsPool.sol` to match the terminology used in `BondingManager`.


contracts/libraries/SortedDoublyLL.sol
```diff
@@ -311,14 +311,14 @@ library SortedDoublyLL {
         address nextId = _nextId;
 
         if (prevId != address(0)) {
-            if (!contains(self, prevId) || _key < self.nodes[prevId].key) {
+            if (!contains(self, prevId) || _key > self.nodes[prevId].key) {
```
yondonfu: Previously, the inequality comparison for keys was flipped - now it is fixed

contracts/token/Minter.sol
```diff
@@ -35,12 +35,6 @@ contract Minter is Manager, IMinter {
         _;
     }
 
-    // Checks if caller is JobsManager
-    modifier onlyJobsManager() {
```
yondonfu: Not used

contracts/token/Minter.sol
```diff
@@ -233,13 +227,6 @@ contract Minter is Manager, IMinter {
         return ILivepeerToken(controller.getContract(keccak256("LivepeerToken")));
     }
 
-    /**
-     * @dev Returns RoundsManager interface
-     */
-    function roundsManager() internal view returns (IRoundsManager) {
```
yondonfu: Not used

contracts/bonding/BondingManager.sol
```diff
@@ -551,6 +548,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
                 TranscoderSlashed(_transcoder, address(0), penalty, 0);
             }
+        } else {
+            TranscoderSlashed(_transcoder, _finder, 0, 0);
```
yondonfu: Still fire an event even if the user being slashed no longer has a bonded amount (the edge case in which the slashing transaction was not processed until the transcoder already unbonded, waited through the unbonding period and withdrew its stake)

contracts/libraries/SortedDoublyLL.sol
```diff
@@ -311,14 +311,14 @@ library SortedDoublyLL {
...
                 // `prevId` does not exist anymore or now has a smaller key than the given key
                 prevId = address(0);
             }
         }
 
         if (nextId != address(0)) {
-            if (!contains(self, nextId) || _key > self.nodes[nextId].key) {
+            if (!contains(self, nextId) || _key < self.nodes[nextId].key) {
```
yondonfu: This inequality comparison for keys was also previously flipped - now it is fixed

PR #205, opened 2018-03-07
Allow job creator to be different from signer for broadcasted segments

### LINKED ISSUE #203: Let job creator be different from signer for broadcasted segments
Currently, the protocol assumes that the creator of a job is the same as the signer for broadcasted segments since the `broadcasterAddress` field in the on-chain `Job` struct is populated by `msg.sender` when `JobsManager.job()` is called. 

This assumption is a blocker for certain use cases where the job creator is a contract account as opposed to a user account. Currently, a contract account can pool funds from other user accounts and call `JobsManager.deposit()` to deposit funds to be used to pay for broadcasting. However, since `JobsManager.job()` assumes that the caller is also the signer for broadcasted segments, the broadcasting workflow does not work since the contract account (the caller of `JobsManager.job()`) cannot sign segments.

The above problem can be fixed if the `job()` function signature changed from:

```
job(string _streamId, string _transcodingOptions, uint256 _maxPricePerSegment, uint256 _endBlock)
```

to

```
job(string _streamId, string _transcodingOptions, address _broadcasterAddress, uint256 _maxPricePerSegment, uint256 _endBlock)
```

such that `_broadcasterAddress` is used to populate the `broadcasterAddress` field in the on-chain `Job` struct. A contract account can then fund the broadcast by depositing funds, but designate an external user account as the signer for the broadcasted segments.

adamsoffer:
Closes #203 

Currently, the protocol assumes the job creator is also the signer for broadcasted segments. This is problematic if the job creator is a contract account since contract accounts cannot sign segments.

This fix changes the job() function signature from:

`job(string _streamId, string _transcodingOptions, uint256 _maxPricePerSegment, uint256 _endBlock)`

to 

`job(string _streamId, string _transcodingOptions, address _broadcasterAddress, uint256 _maxPricePerSegment, uint256 _endBlock)`

such that `_broadcasterAddress` is used to populate the `broadcasterAddress` field in the on-chain `Job` struct. 
 
This unlocks some potentially interesting use cases such as the ability to crowdfund a broadcast.

### DIFF
--- a/contracts/jobs/IJobsManager.sol
+++ b/contracts/jobs/IJobsManager.sol
@@ -8,7 +8,7 @@ pragma solidity ^0.4.17;
 contract IJobsManager {
     event Deposit(address indexed broadcaster, uint256 amount);
     event Withdraw(address indexed broadcaster);
-    event NewJob(address indexed broadcaster, uint256 jobId, string streamId, string transcodingOptions, uint256 maxPricePerSegment, uint256 creationBlock);
+    event NewJob(address indexed broadcaster, uint256 jobId, string streamId, string transcodingOptions, uint256 maxPricePerSegment, uint256 creationBlock, address indexed jobCreator);
     event NewClaim(address indexed transcoder, uint256 indexed jobId, uint256 claimId);
     event Verify(address indexed transcoder, uint256 indexed jobId, uint256 indexed claimId, uint256 segmentNumber);
     event DistributeFees(address indexed transcoder, uint256 indexed jobId, uint256 indexed claimId, uint256 fees);
--- a/contracts/jobs/JobsManager.sol
+++ b/contracts/jobs/JobsManager.sol
@@ -226,15 +226,18 @@ contract JobsManager is ManagerProxyTarget, IVerifiable, IJobsManager {
      * @dev Submit a transcoding job
      * @param _streamId Unique stream identifier
      * @param _transcodingOptions Output bitrates, formats, encodings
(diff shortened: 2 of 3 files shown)


ads1018:
@yondonfu To make the distinction between a broadcaster and a job creator more clear, I think it might also be worth renaming the `Broadcaster` struct to `JobCreator` and the `broadcasters` mapping to `jobCreators`. Thoughts?

yondonfu:
@ads1018 Awesome, thanks for the PR! The contract codebase is currently frozen for an ongoing external audit that should be completed by next week with the plan being to evaluate additional necessary changes at that point. Let's definitely check back in on this PR then - I'll be sure to ping you!

dob:
Yah, thanks for the PR @ads1018! Looking forward to getting this merged after the code freeze.

ads1018:
@dob @yondonfu my pleasure! Had a good time reading through the contract code in the process. Really mindblowing stuff. Looking forward to after the code freeze.

yondonfu:
Hey @ads1018 I left a comment on this issue https://github.com/livepeer/protocol/issues/203 describing some thoughts on why this particular change likely won't be included in the initial version of the protocol (but could definitely be included in the next version of the protocol!). If you have any thoughts/feedback definitely leave them on that issue and we can continue the conversation there!

yondonfu:
Closing this PR since the protocol will no longer involve on-chain jobs in Streamflow - see [this comment](https://github.com/livepeer/protocol/issues/203#issuecomment-481847615) for additional notes on designated signers for contracts.

PR #207, opened 2018-03-13
Deployment and genesis related updates

yondonfu:
Purposes of this PR:
- Edit the mock token distribution contract so it can be used as a dummy token distribution contract that adheres to the interface that the GenesisManager expects, but it does not actually do anything and just serves as a placeholder during genesis. There are no additional contract changes outside of `TokenDistributionMock.sol` which is in the `contracts/test` directory so I do not believe that was in the scope of the audit anyways
- Add a script to validate the contract system state post deployment and genesis that can be run as soon as we finish running our migration scripts as a sanity check that all the parameters are initialized to the correct values
- I created a package [truffle-keystore-provider](https://github.com/yondonfu/truffle-keystore-provider) that allows deployment using a remote ETH node by unlocking a local keystore (does not require a local node - just the keystore file) with a password to sign the deployment transactions. Previously, the private key provider forced the user to set the plaintext private key as an environment variable which did not seem ideal


migrations/4_genesis.js
```diff
@@ -8,20 +8,26 @@ const LivepeerToken = artifacts.require("LivepeerToken")
...
         const token = await LivepeerToken.at(tokenAddr)
 
-        const tokenDistribution = await lpDeployer.deploy(TokenDistributionMock, tokenAddr, faucetAddr, genesis.tokenDistribution.endTime)
-        const genesisManager = await lpDeployer.deploy(GenesisManager, tokenAddr, tokenDistribution.address, genesis.bankMultisig, minterAddr)
+        const dummyTokenDistribution = await lpDeployer.deploy(TokenDistributionMock, genesis.dummyTokenDistribution.timeToEnd)
+
+        let genesisManager
+
+        if (!lpDeployer.isLiveNetwork(network)) {
+            genesisManager = await lpDeployer.deploy(GenesisManager, tokenAddr, dummyTokenDistribution.address, accounts[0], minterAddr)
```
yondonfu: If we're not deploying to a live network, transfer the "company supply" to the account used for the deployment transactions so that account can transfer LPT to other accounts during automated tests

PR #209, opened 2018-03-23
Add script to get vesting contracts

ericxtang:
Should also add `"getVestingContracts": "truffle exec scripts/getVestingContracts.js"` into package.json, but it would cause a merge conflict now.

### DIFF
--- a/scripts/getVestingContracts.js
+++ b/scripts/getVestingContracts.js
@@ -1,0 +1,21 @@
+const GenesisManager = artifacts.require("GenesisManager")
+const genesisConfig = require("../migrations/genesis.config.js")
+
+module.exports = async () => {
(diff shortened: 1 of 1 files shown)


scripts/getVestingContracts.js
```diff
@@ -0,0 +1,21 @@
+const GenesisManager = artifacts.require("GenesisManager")
+const genesisConfig = require("../migrations/genesis.config.js")
+
+module.exports = async () => {
+    const genesisManager = await GenesisManager.at("0x9a4d4cafbb1fc3b00d387095144e34b993d82d9a")
```
yondonfu: The address passed in here will change from deployment to deployment - I think we could change this to `await GenesisManager.deployed()` which will always work right after deployment (from the machine that the deployment account lives on) and if the script ever needs to be run somewhere else where the Truffle artifacts aren't present there can be hand made change to pass in the address

yondonfu:
🚢

PR #210, opened 2018-03-23
Additional deployment related updates

yondonfu:
Deployment related updates
- Make sure the dummy token distribution contract can't be prematurely finalized during deployment
- Fix ethereumjs-abi dependency
- Make sure faucet is deployed and funded in non-production environments
- Make sure assertions in `validateDeploymentAndGenesis.js` use case insensitive comparisons for addresses
- Add some additional visibility for what is happening when running scripts

### DIFF
--- a/contracts/test/TokenDistributionMock.sol
+++ b/contracts/test/TokenDistributionMock.sol
@@ -2,9 +2,10 @@ pragma solidity ^0.4.17;
 
 import "../token/ITokenDistribution.sol";
 import "zeppelin-solidity/contracts/math/SafeMath.sol";
+import "zeppelin-solidity/contracts/ownership/Ownable.sol";
 
 
-contract TokenDistributionMock is ITokenDistribution {
+contract TokenDistributionMock is ITokenDistribution, Ownable {
     using SafeMath for uint256;
 
     // End time of the distribution
@@ -17,7 +18,7 @@ contract TokenDistributionMock is ITokenDistribution {
         over = false;
     }
 
-    function finalize() external {
+    function finalize() external onlyOwner {
         require(!isOver());
 
         over = true;
--- a/migrations/3_deploy_contracts.js
+++ b/migrations/3_deploy_contracts.js
@@ -21,7 +21,8 @@ module.exports = function(deployer, network) {
         await lpDeployer.deployAndRegister(Minter, "Minter", controller.address, config.minter.inflation, config.minter.inflationChange, config.minter.targetBondingRate)
         await lpDeployer.deployAndRegister(LivepeerVerifier, "Verifier", controller.address, config.verifier.solvers, config.verifier.verificationCodeHash)
 
-        if (!lpDeployer.isMainNet(network)) {
+        if (!lpDeployer.isProduction(network)) {
+            // Only deploy a faucet if not in production
             await lpDeployer.deployAndRegister(LivepeerTokenFaucet, "LivepeerTokenFaucet", token.address, config.faucet.requestAmount, config.faucet.requestWait)
         }
 
@@ -31,6 +32,7 @@ module.exports = function(deployer, network) {
         let roundsManager
 
         if (!lpDeployer.isLiveNetwork(network)) {
+            // Only deploy the adjustable rounds manager contract if we are in an isolated testing environment and not a live network
             roundsManager = await lpDeployer.deployProxyAndRegister(AdjustableRoundsManager, "RoundsManager", controller.address)
         } else {
             roundsManager = await lpDeployer.deployProxyAndRegister(RoundsManager, "RoundsManager", controller.address)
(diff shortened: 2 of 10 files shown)


contracts/test/TokenDistributionMock.sol
```diff
@@ -17,7 +18,7 @@ contract TokenDistributionMock is ITokenDistribution {
         over = false;
     }
 
-    function finalize() external {
```
yondonfu: A previous change allowed this function to be called by anyone at anytime. If this function is called before all the grants are added in the genesis migration script then the grants cannot be added afterwords. Adding the `onlyOwner` modifier ensures that the deployment account can only call `finalize` and allows the deployment account to make sure to add all the grants before calling this function

migrations/3_deploy_contracts.js
```diff
@@ -21,7 +21,8 @@ module.exports = function(deployer, network) {
         await lpDeployer.deployAndRegister(Minter, "Minter", controller.address, config.minter.inflation, config.minter.inflationChange, config.minter.targetBondingRate)
         await lpDeployer.deployAndRegister(LivepeerVerifier, "Verifier", controller.address, config.verifier.solvers, config.verifier.verificationCodeHash)
 
-        if (!lpDeployer.isMainNet(network)) {
+        if (!lpDeployer.isProduction(network)) {
```
yondonfu: If we are deploying to main net or doing a dry run on a test network we will not deploy a faucet

migrations/4_genesis.js
```diff
@@ -23,7 +23,8 @@ module.exports = function(deployer, network, accounts) {
 
         let genesisManager
 
-        if (!lpDeployer.isLiveNetwork(network)) {
+        if (!lpDeployer.isProduction(network)) {
```
yondonfu: If we are not deploying to main net and we are not doing a dry run on a test network we send the company supply to the deployment account instead of the bank multisig so the deployment account can transfer tokens wherever necessary - this is primarily to make it easier to automate the process of funding the faucet without having to edit the config file

migrations/4_genesis.js
```diff
@@ -73,5 +74,14 @@ module.exports = function(deployer, network, accounts) {
         deployer.logger.log("Ending genesis and transferring ownership of the LivepeerToken to the protocol Minter...")
 
         await genesisManager.end()
+
+        if (!lpDeployer.isProduction(network)) {
+            // If not in production, the deployment account holds the company supply
+            // and use some of those tokens to fund the faucet supply
+            deployer.logger.log("Not in production - deployment account will fund the deployed faucet")
+
+            const faucetAddr = await controller.getContract(contractId("LivepeerTokenFaucet"))
```
yondonfu: If we are not deploying to main net and we are not doing a dry run on a test network, we have already sent the company supply to the deployment account at this point. So, the deployment account can use those tokens to fund the faucet which will also have been deployed at this point. This automates the process of funding the faucet in non-dry run test networks and isolated testing environments like devenv

package.json
```diff
@@ -48,7 +49,7 @@
     "coveralls": "^3.0.0",
     "eslint": "^4.3.0",
     "eslint-config-google": "^0.9.1",
-    "ethereumjs-abi": "github:yondonfu/ethereumjs-abi#abi-offset-fixed-arrays",
```
yondonfu: We were previously using my fork of this package with a specific fix. The fix was recently merged in, but a new release has not been cut for the NPM package yet, so for now we can install this dependency from Github using the commit hash that includes the mentioned fix

scripts/validateDeploymentAndGenesis.js
```diff
@@ -137,7 +137,7 @@ module.exports = async () => {
     genesisConfig.teamGrants.forEach(async grant => {
         const vestingHolderAddr = await genesisManager.vestingHolders.call(grant.receiver)
         const vestingHolder = await TokenVesting.at(vestingHolderAddr)
-        assert.equal(await vestingHolder.beneficiary.call(), grant.receiver, "should be correct vesting grant receiver")
+        assert.equal((await vestingHolder.beneficiary.call()).toLowerCase(), grant.receiver.toLowerCase(), "should be correct vesting grant receiver")
```
yondonfu: This assertion could fail before if the addresses entered in `genesis.config.js` were checksummed addresses. Now the assertion compares the lower cased representation of the input addresses

truffle.js
```diff
@@ -64,6 +64,20 @@ module.exports = {
             },
             network_id: 4,
             gas: 6600000
+        },
+        rinkebyDryRun: {
```
yondonfu: Added this network config which can be passed in when we specifically want to do a dry run on Rinkeby that mimics what the deployment would look like on main net i.e. no faucet

ericxtang:
LGTM.  🚢

PR #211, opened 2018-03-28
Provider updates and pull Controller ownership transfer into migrations

yondonfu:
- Update truffle-keystore-provider version so checksummed addresses are handled properly if needed
- Pull in Controller ownership transfer into migration script so an additional script does not need to be run post-deployment
- Add a check in the validation script to check that the owner of the Controller is the governance multisig


ericxtang:
🚢

PR #214, opened 2018-04-20
Deployment fixes

yondonfu:
- Remove unneeded library deployments
- Fix some checks in the post deployment validation script
- Bump truffle-keystore-provider version to fix potentially unsynchronized nonce issue when using Infura during deployment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,11 +27,8 @@ jobs:
           name: Run linting
           command: npm run docker:run:lint
       - run:
-          name: Run unit tests with coverage
+          name: Run unit tests
           command: npm run docker:run:test:coverage
-      - run:
-          name: Run unit tests without coverage
-          command: npm run docker:run:test:unit
       - run:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -27,11 +27,8 @@ jobs:
           name: Run linting
           command: npm run docker:run:lint
       - run:
-          name: Run unit tests with coverage
+          name: Run unit tests
```
yondonfu: Before we were running the unit tests twice - once with coverage once without because the coverage tool was using an outdated package that caused some tests to fail. The coverage tool is up to date now and those tests no longer fail so we only need to run unit tests once

yondonfu:
The `SafeMath`, `MathUtils` and `EarningsPool` libraries only contained internal functions so at compile time the Solidity compiler pulls in the library code and directly inlines the library code into the contracts that import the libraries so we don't have to actually deploy those libraries.

PR #222, opened 2018-06-04
Implement LIP-3: Ability to update registered solver in LivepeerVerifier

yondonfu:
Implementation for LIP-3: https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-3.md

Also updated some of the comments in LivepeerVerifier to be more descriptive and to use the NatSpec format.

### DIFF
--- a/contracts/verification/LivepeerVerifier.sol
+++ b/contracts/verification/LivepeerVerifier.sol
@@ -5,13 +5,15 @@ import "./IVerifier.sol";
 import "./IVerifiable.sol";
 
 
+/**
+ * @title LivepeerVerifier
+ * @dev Manages transcoding verification requests that are processed by a trusted solver
+ */
 contract LivepeerVerifier is Manager, IVerifier {
     // IPFS hash of verification computation archive
     string public verificationCodeHash;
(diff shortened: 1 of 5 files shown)


yondonfu:
Yeah `upgradeRound` and `upgradeBlock` are procedural and descriptive values meant to offer a degree of transparency whenever this update happens so users have some understanding of when this feature will be included. Although not relevant for this particular update because clients do not interact with the LivepeerVerifier at all, these values might also be useful if clients want to embed logic to switch to using a new feature after some block or round (because it knows that the update should be live at that point).

The last section of the specification for LIP-3 is basically the checklist for how this particular update should occur. I don't think ensuring that the upgrade happens as prescribed in an automated fashion is really possible since registering the new LivepeerVerifier contract with the Controller is a multisig transaction that needs to be manually executed. Though once the LIP is accepted for inclusion in an upcoming protocol upgrade event we can definitely create a manual checklist for that event

ericxtang:
That makes sense.  In that case, what do you think about adding a `SolverSet(newAddr)` event so clients can subscribe to it and update internal states accordingly?

PR #223, opened 2018-06-05
Implement LIP-9: Service Registry

yondonfu:
Implementation for LIP-9: https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-9.md

### DIFF
--- a/contracts/ServiceRegistry.sol
+++ b/contracts/ServiceRegistry.sol
@@ -1,0 +1,45 @@
+pragma solidity ^0.4.17;
+
+import "./ManagerProxyTarget.sol";
+
+
+/**
+ * @title ServiceRegistry
+ * @dev Maintains a registry of service metadata associated with service provider addresses (transcoders/orchestrators)
+ */
+contract ServiceRegistry is ManagerProxyTarget {
+    // Store service metadata
+    struct Record {
+        string serviceURI;   // Service URI endpoint that can be used to send off-chain requests
+    }
+
+    // Track records for addresses
+    mapping (address => Record) private records;
+
+    // Event fired when a caller updates its service URI endpoint
+    event ServiceURIUpdate(address indexed addr, string serviceURI);
+
+    /**
+     * @dev ServiceRegistry constructor. Only invokes constructor of base Manager contract with provided Controller address
+     * @param _controller Address of a Controller that this contract will be registered with
+     */
+    function ServiceRegistry(address _controller) public Manager(_controller) {}
+
+    /**
+     * @dev Stores service URI endpoint for the caller that can be used to send requests to the caller off-chain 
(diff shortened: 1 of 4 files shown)


contracts/ServiceRegistry.sol
```diff
@@ -0,0 +1,45 @@
...
+     * @dev ServiceRegistry constructor. Only invokes constructor of base Manager contract with provided Controller address
+     * @param _controller Address of a Controller that this contract will be registered with
+     */
+    function ServiceRegistry(address _controller) public Manager(_controller) {}
+
+    /**
+     * @dev Stores service URI endpoint for the caller that can be used to send requests to the caller off-chain 
+     * @param _serviceURI Service URI endpoint for the caller
+     */
+    function setServiceURI(string _serviceURI) external {
```
j0sh: Are there any potential negative externalities for the rest of the network if a bunch of non-transcoders set their service URI records here?

yondonfu: Hm since a broadcaster client would only care about the service URI record for the transcoder address that was assigned to the broadcaster's transcode job if non-transcoder users set their own service URI record they would pay the tx cost, but at least in the current proposed networking architecture those records will never be used in practice unless the non-transcoder user becomes a transcoder in the future. Perhaps something to be wary about is if a transcoder's key is compromised such that an attacker can change the service URI record right before a broadcaster looks it up to send requests to the transcoder

j0sh: Yeah, keeping the transcoder's keys keys cold would entail something like the discussion in https://github.com/livepeer/LIPs/issues/7 which probably should be done separately from this change.

I was mostly wondering if filling the record mapping with unneeded entries could lead to higher costs for the network down the road, eg during a migration or with any enactment of 'data rents', but that's all fairly speculative. LGTM.

contracts/ServiceRegistry.sol
```diff
@@ -0,0 +1,45 @@
...
+import "./ManagerProxyTarget.sol";
+
+
+/**
+ * @title ServiceRegistry
+ * @dev Maintains a registry of service metadata associated with service provider addresses (transcoders/orchestrators)
+ */
+contract ServiceRegistry is ManagerProxyTarget {
+    // Store service metadata
+    struct Record {
```
ericxtang: If we changed the Record struct in the future, the storage layout will be screwy right?  If that's the case, I prefer to have the mapping from `address` to `string`, that way this potential mistake doesn't happen.

yondonfu: Adding additional fields to the `Record` struct in an upgrade is safe as long as 1) we do not remove/replace existing fields 2) the `Record` struct is not used in an array (it is only used in a mapping right now)

PR #225, opened 2018-06-29
Implement LIP-8: Partial Unbonding

yondonfu:
A few notes on storage layout for proxy upgrade compatibility:
- Renamed `withdrawRound` in the `Delegator` struct to `withdrawRoundDEPRECATED` to clearly indicate that this field should never be used within the code. We cannot remove the field because the storage slot could potentially already be filled with old values. However, we can rename the field to make it clear that it shouldn't be used and the corresponding storage slot will just never be used again.
- Removed the `Unbonding` type from the `DelegatorStatus` enum. Since we never explicitly store the `DelegatorStatus` enum and only compute it based on the current state of a delegator's address, it is safe to remove types from the enum.

The spec outlined in LIP-8 will likely need to be updated in the following areas (can be done after some discussion and review):
- Signatures for `Bond`, `Rebond`, and `WithdrawStake` events
- Clarify that a delegator can rebond using an unbonding lock while in the Pending state (defined as when a delegator delegates to an account in the current round, but does not enter the Bonded state until the following round). I don't see any issues with allowing this at the moment.
- More detailed descriptions for the `rebond()` function. I opted to restrict its usage such that it can only be used for an unbonding lock if a delegator is currently delegated to an account - the unbonding lock amount is then applied to the delegator's current delegate. A delegator that is unbonded such that it is not delegated toward anyone cannot use the `rebond()` function even it it has an unbonding lock available until it is actively delegated to an account again.


yondonfu:
Update in 1581f83a4db10cc22b2f76615da5e771c3e89d53:
- Applied the `autoClaimEarnings` modifier to the `rebond()` function. When rebonding using an unbonding lock, all reward shares from past rounds MUST be claimed and credited to the delegator's bonded amount before executing the logic in `rebond()`. Otherwise, `rebond()` would update the delegator's `bondedAmount` such that it would no longer reflect the actual value during the delegator's `lastClaimRound`. As a result, the next time the delegator claims reward shares through past rounds the  calculations would be based off an inaccurate value for the delegator's `bondedAmount`. Applying the `autoClaimEarnings` modifier to the `rebond()` function solves this issue. Generally, any function that updates a delegator's `bondedAmount` should use the `autoClaimEarnings` modifier to ensure that reward shares from past rounds are credited to the delegator's `bondedAmount` first.
- Added an integration test in `tests/integration/Delegation.js` for the scenario in which a delegator partially unbonds, its transcoder distributes rewards in the next round, and then the delegator rebonds using an unbonding lock in the round after that. When rebonding with the unbonding lock, the contract should first credit the delegator with reward shares before executing the rebonding logic.

dob:
> A delegator that is unbonded such that it is not delegated toward anyone cannot use the rebond() function even it it has an unbonding lock available until it is actively delegated to an account again.

What's the reason for this restriction? It just becomes impossible to tell which delegate to rebond to without tracking extra state? 

Not mandatory, but I see it as a nice-to-have to be able to rebond, even if you are unbonding all your token, because I see this happening often. Someone opts to unbond, or does the txn without fully realizing the ramifications, and then decides they want to bond again. 

I guess there's always the workaround of bonding some additional LPT (if you have it), and then you can rebond back to the transcoder.

contracts/bonding/BondingManager.sol
```diff
@@ -332,68 +335,135 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            // transcoder pool so its delegated stake within the pool does not need to be decreased
+            transcoderPool.updateKey(del.delegateAddress, transcoderPool.getKey(del.delegateAddress).sub(_amount), address(0), address(0));
         }
 
-        if (del.delegateAddress != msg.sender && transcoderStatus(del.delegateAddress) == TranscoderStatus.Registered) {
-            // If delegate is not self and is a registered transcoder, decrease its delegated stake
-            // We do not need to decrease delegated stake if delegate is self because we would have already removed self
-            // from the transcoder pool
-            transcoderPool.updateKey(del.delegateAddress, transcoderPool.getKey(del.delegateAddress).sub(del.bondedAmount), address(0), address(0));
+        // Check if delegator is still bonded
```
dob: This comment is slightly confusing to me. We're actually checking if they should no longer be bonded. Maybe:

`// Check if delegator has zero bondedAmount remaining, and if so reset their delegation status`

contracts/bonding/BondingManager.sol
```diff
@@ -785,7 +851,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     )
         public
         view
-        returns (uint256 bondedAmount, uint256 fees, address delegateAddress, uint256 delegatedAmount, uint256 startRound, uint256 withdrawRound, uint256 lastClaimRound)
+        returns (uint256 bondedAmount, uint256 fees, address delegateAddress, uint256 delegatedAmount, uint256 startRound, uint256 lastClaimRound, uint256 nextUnbondingLockId)
```
dob: also need an update to the `getDelegator()` function in the LIP, for reference for client implementations.

contracts/bonding/BondingManager.sol
```diff
@@ -332,68 +335,135 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+    function rebond(
+        uint256 _unbondingLockId
+    ) 
+        external
+        whenSystemNotPaused
+        currentRoundInitialized 
+        autoClaimEarnings
+    {
+        // Caller must not be an unbonded delegator
+        require(delegatorStatus(msg.sender) != DelegatorStatus.Unbonded);
```
dob: What do you think about the addition of a function `rebondFromUnbondingLockWhileUnbonded(uint256 _unbondingLockID)`, which mostly replicates the logic of the bond function, except only does it if the user has an active unbondingLock but is in the Unbonded state?

This would enable rebonding in the case where you're currently Unbonded in the waiting period, but it would keep the logic of the bond() and rebond() methods straightforward without having to handle edge cases?

yondonfu:
> What's the reason for this restriction?

I originally applied the restriction to simplify the implementation and reduce the amount of complexity that might be necessary while rebonding. However, I actually think relative to the case where a delegator rebonds from the `Bonded` or `Pending` states, the only additional logic that needs to be executed when a delegator rebonds from the `Unbonded` state to a delegate of its choice is updating the delegator's `startRound` (which will transition the delegator into the `Pending` state). We would end up with two functions:

- `rebond(uint256 _unbondingLockId)`: only callable when the delegator is in the `Bonded` or `Pending` state. Rebonds tokens for the referenced unbonding lock to the delegator's current delegate.
- `rebondFromUnbonded(address _to, uint256 _unbondingLockId)`: only callable when the delegator is in the `Unbonded` state. Sets the delegator's `startRound` to `currentRound + 1` (which transitions the delegator to the `Pending` state. Then, rebonds tokens for the referenced unbonding lock to the delegator's new delegate `_to`.

contracts/bonding/BondingManager.sol
```diff
@@ -332,68 +335,141 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+    function withdrawStake(uint256 _unbondingLockId)
+        external
+        whenSystemNotPaused
+        currentRoundInitialized
+    {
+        Delegator storage del = delegators[msg.sender];
+        UnbondingLock storage lock = del.unbondingLocks[_unbondingLockId];
+
+        // Unbonding lock must be valid
+        require(lock.withdrawRound > 0);
```
dob: I see two instances of checking if the lock.withdrawRound > 0 determining whether the unbonding lock is valid? Should we move to a convenience method for code readability and reducing the chance of a future error when trying to figure out what it means for a lock to be valid?

yondonfu:
Added public getter `isValidUnbondingLock(address _delegator, uint256 _unbondingLockId)` which is used in `processRebond()` and `withdrawStake()` to determine if a referenced unbonding lock for a delegator is valid (based on its `withdrawRound`).

PR #231, opened 2018-08-29
Update earnings pool accounting 

yondonfu:
- Updates to `EarningsPool.sol`
  - `hasTranscoderRewardFeePool` flag which is set to true by the `init` function. After an upgrade to the BondingManager (which would include these changes since the BondingManager pulls in `EarningsPool.sol` as a dependency), whenever a `EarningsPool.Data` struct is initialized using `init` in `BondingManager.setActiveTranscoders()`, the struct will have this flag enabled. Older `EarningsPool.Data` structs will not have this flag enabled. If this flag is enabled, the earnings pool should maintain separate delegator/transcode reward/fee pools. If this flag is disabled, the earnings pool should use a single reward/fee pool and calculate shares using the old library logic.
  - `addToFeePool()` is used by BondingManager to send transcode job fees to an earnings pool
  - `addToRewardPool()` is used by BondingManager to send rewards to an earnings pool
  - See inline comments for additional information
- Additional unit tests for `EarningsPool.sol`
  - They are separated across `TestEarningsPool.sol` and `TestEarningsPool2.sol`, which set up tests with `hasTranscoderRewardFeePool` enabled, and `TestEarningsPoolNoTranscoderRewardFeePool.sol`, which sets up tests with `hasTranscoderRewardFeePool` disabled.
- Integration test for reward share distributions in `integration/Rewards.js`
- Integration test for upgrading from an old version of BondingManager + EarningsPool to the current version of BondingManager + EarningsPool
  - The old versions are stored in the `previous` folder. There might be a better way to manage this, but using this approach for simplicity for the time being
- Allow `truffle test` to be run with Mocha's `--grep` flag to selectively run specific tests in a file
- Fix up linting

I would like to add a few more tests, but pushing this up for now to get a first pass review of the contract changes.


yondonfu:
a0a606508d225d3d8aeefa73eaf4431bc5dda1ba: Updated the return signature for `BondingManager.getTranscoderEarningsPoolForRound()` and added a clarifying comment in `EarningsPool.sol` about the feature flag

dob:
Ok, the updated changes in a0a6065 look good to me. On with rinkeby testing.

PR #242, opened 2018-10-01
Update getTotalBonded()

yondonfu:
- Updated getTotalBonded() to sum delegated stake of members of registered transcoder pool
- Updated unit tests and integration tests based on updated logic of getTotalBonded()


contracts/bonding/BondingManager.sol
```diff
@@ -71,7 +71,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     mapping (address => Transcoder) private transcoders;
 
     // Keep track of total bonded tokens
```
dob: I see that we changed the variable name, but let's also update the comment to specify why this is deprecated.

contracts/bonding/BondingManager.sol
```diff
@@ -929,6 +924,17 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @dev Return total bonded tokens
      */
     function getTotalBonded() public view returns (uint256) {
+        uint256 totalBonded = 0;
+        uint256 totalTranscoders = transcoderPool.getSize();
+        address currentTranscoder = transcoderPool.getFirst();
+
+        for (uint256 i = 0; i < totalTranscoders; i++) {
+            // Add current transcoder's total delegated stake to total bonded counter
+            totalBonded = totalBonded.add(transcoderPool.getKey(currentTranscoder));
```
dob: I can't remember if we added a well named helper or not to get a transcoder's delegated stake from the transcoder pool, but since it's not used here...

I still think a more readable line of code would be something like `totalBonded = totalBonded.add(delegatedStakeForTranscoder(currentTranscoder))` since I always find it confusing that the value returend by `getKey` is the delegated stake.

yondonfu: Updated to use `transcoderTotalStake()` helper

PR #244, opened 2018-11-19
Fix CircleCI and solidity-coverage

yondonfu:
- Update `circleci/config.yml`. Previously, our CI was setup to compile contracts and run tests within a Docker container. The original intention was for the CI to build a Docker image with contracts already compiled for use. With Truffle v5, we are now using a Dockerized version of solc which enables much faster contract compilation: https://github.com/livepeer/protocol/blob/pm/truffle.js#L103 As a result, if we continued with the current CI setup, we would have to run Docker within a Docker container. We can avoid this by just compiling contracts and running tests outside of a Docker container. If we need to build a Docker image with contracts already compiled for use we can always build the Docker image outside of the CI.
- Update solidity-coverage in order to support newer versions of solc.
- Move Solidity tests into `contracts/test`. solidity-coverage doesn't play well with Solidity tests (perhaps a reason to avoid writing Solidity tests in the future) due to the way the tool instruments contracts in order to generate coverage reports. See https://github.com/sc-forks/solidity-coverage/issues/182 for more on this issue. We can workaround this for now by moving the Solidity tests into the `contracts/test` directory and using the `runSolidityTests.js` (borrowed from Aragon - see comment in file for the original source code) to run the Solidity tests as a part of the JS tests that Truffle runs. 
- Remove unneeded scripts/contracts


contracts/test/helpers/Assert.sol
```diff
@@ -0,0 +1,1403 @@
+// Taken from https://github.com/trufflesuite/truffle/blob/ef03be9e35909e5e7c97a9ecb07ccae4872645ca/packages/truffle-core/lib/testing/Assert.sol
```
yondonfu: Typically, if we ran the Solidity tests with the normal Truffle test workflow, the Solidity tests would be able to access this library via `import truffle/Assert.sol`. However, since we are using a workaround that involves running the Solidity tests as a part of the JS tests in the Truffle workflow the Solidity tests are unable to access this library via `import truffle/Assert.sol`. Instead, we manually include this library so that it can be imported by the Solidity tests

PR #260, opened 2018-11-23
Delayed withdrawal

eladmallel:


test/unit/LivepeerETHTicketBroker.js
```diff
@@ -16,14 +15,16 @@ contract("LivepeerETHTicketBroker", accounts => {
...
+
     before(async () => {
         fixture = new Fixture(web3)
         await fixture.deploy()
     })
 
     beforeEach(async () => {
         await fixture.setUp()
-        broker = await TicketBroker.new(fixture.controller.address, 0)
+        broker = await TicketBroker.new(fixture.controller.address, 0, unlockPeriod)
```
yondonfu: The calls to Fixture in the `beforeEach` and `afterEach` hooks should reset the test chain state to whatever it was post `before` hook so we could move this line into the `before` hook to just deploy TicketBroker once

test/unit/LivepeerETHTicketBroker.js
```diff
@@ -305,5 +306,29 @@ contract("LivepeerETHTicketBroker", accounts => {
                 assert.equal(events.length, 0)
             })
         })
+
+        describe("withdrawTransfer", () => {
```
yondonfu: Should this describe block be nested within a parent describe block for `withdraw` similar to what is done for `redeemWinningTicket` and the nested describe blocks for `winningTicketTransfer` and `penaltyEscrowSlash`?

test/unit/TicketBroker.js
```diff
@@ -5,17 +5,25 @@ import expectThrow from "../helpers/expectThrow"
...
 
+    const unlockPeriod = 20
+
+    before(async () => {
+        fixture = new Fixture(web3)
+    })
+
     beforeEach(async () => {
-        broker = await TicketBroker.new(0)
+        broker = await TicketBroker.new(0, unlockPeriod)
```
yondonfu: Should we move this line into the `before` hook and use `Fixture` to reset the state between each test similar to what we do in `test/LivepeerETHTicketBroker.js`?

test/unit/TicketBroker.js
```diff
@@ -620,4 +648,245 @@ contract("TicketBroker", accounts => {
...
+        })
+
+        it("sets withdrawBlock according to constructor config", async () => {
+            await broker.fundDeposit({from: sender, value: 1000})
+            const fromBlock = (await web3.eth.getBlock("latest")).number
+
+            await broker.unlock({from: sender})
+
+            const withdrawBlock = (await broker.senders.call(sender)).withdrawBlock.toString()
+            const expectedWithdrawBlock = fromBlock + unlockPeriod + 1 // +1 to account for the block created executing unlock
```
yondonfu: We can avoid this +1 by moving the `fromBlock` line after the unlock tx

test/unit/TicketBroker.js
```diff
@@ -620,4 +648,245 @@ contract("TicketBroker", accounts => {
...
+        })
+
+        it("sets isUnlockInProgress to false", async () => {
+            await broker.fundDeposit({from: sender, value: 1000})
+            await broker.unlock()
+
+            await broker.cancelUnlock()
+
+            const isUnlockInProgress = await broker.isUnlockInProgress.call(sender)
+            assert.equal(isUnlockInProgress, false)
```
yondonfu: `assert(!isUnlockInProgress)`?

test/unit/TicketBroker.js
```diff
@@ -620,4 +648,245 @@ contract("TicketBroker", accounts => {
...
+            assert.equal(event.returnValues.sender, sender)
+        })
+    })
+
+    describe("withdraw", () => {
+        it("reverts when both deposit and penaltyEscrow are zero", async () => {
+            await expectRevertWithReason(broker.withdraw(), "sender deposit and penalty escrow are zero")
+        })
+
+        it("reverts when account is locked", async () => {
```
yondonfu: Perhaps we could split this test into 2 tests: the first for when an unlock request was never made and the second for when an unlock request was made, but the sender has not waited through the unlock period yet

contracts/pm/TicketBroker.sol
```diff
@@ -114,6 +127,67 @@ contract TicketBroker {
...
+        emit Withdrawal(msg.sender, withdrawalAmount);
+    }
+
+    function isUnlockInProgress(address _sender) public view returns (bool) {
+        Sender memory sender = senders[_sender];
+        return _isUnlockInProgress(sender);
+    }
+
+    function _cancelUnlock(Sender storage _sender, address _senderAddress) internal {
+        require(_isUnlockInProgress(_sender), "no unlock request in progress");
```
yondonfu: The `processDeposit` and `processPenaltyEscrow` modifiers already have a check using `_isUnlockInProgress()` so we could move this require statement into `cancelUnlock()`

contracts/pm/TicketBroker.sol
```diff
@@ -40,6 +42,9 @@ contract TicketBroker {
     );
     event WinningTicketTransfer(address indexed sender, address indexed recipient, uint256 amount);
     event PenaltyEscrowSlashed(address indexed sender, address indexed recipient, uint256 amount);
+    event Unlock(address indexed sender);
```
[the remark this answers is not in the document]
eladmallel: Great idea, I like erring on the side of extra info. Adding!

contracts/pm/TicketBroker.sol
```diff
@@ -52,15 +61,19 @@ contract TicketBroker {
     modifier processPenaltyEscrow(address _sender, uint256 _amount) {
         require(_amount >= minPenaltyEscrow, "penalty escrow amount must be >= minPenaltyEscrow");
 
-        senders[_sender].penaltyEscrow += _amount;
-
+        Sender storage sender = senders[_sender];
+        sender.penaltyEscrow += _amount;
+        if (_isUnlockInProgress(sender)) {
+            _cancelUnlock(sender, _sender);
+        }
```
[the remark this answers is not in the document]
eladmallel: 👍

contracts/pm/TicketBroker.sol
```diff
@@ -114,6 +120,63 @@ contract TicketBroker {
...
+    }
+
+    function withdraw() public {
+        Sender storage sender = senders[msg.sender];
+
+        require(
+            sender.deposit > 0 || sender.penaltyEscrow > 0,
+            "sender deposit and penalty escrow are zero"
+        );
+        require(
```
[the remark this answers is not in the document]
eladmallel: 👍

contracts/pm/TicketBroker.sol
eladmallel: here's why I thought the current approach makes sense:

- this `require` is here to make sure we don't emit an event unless it's really interesting
- I still want to keep it here in case we use this function in other places and don't wrap it with an if or require statement
- I added the if statements in the modifiers to prevent them from failing because of this error

If I do what you suggest we leave this function open to misuse such that the event might be emitted even if `withdrawBlock` was already set to zero.

wdyt?

contracts/pm/TicketBroker.sol
```diff
@@ -114,6 +120,63 @@ contract TicketBroker {
...
+
+        emit Withdrawal(msg.sender, withdrawalAmount);
+    }
+
+    function isUnlockInProgress(address _sender) public view returns (bool) {
+        Sender memory sender = senders[_sender];
+        return _isUnlockInProgress(sender);
+    }
+
+    function _isUnlockInProgress(Sender sender) internal pure returns (bool) {
```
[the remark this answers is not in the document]
eladmallel: 👍

test/unit/LivepeerETHTicketBroker.js
eladmallel: 👍

test/unit/LivepeerETHTicketBroker.js
eladmallel: 👍

contracts/pm/TicketBroker.sol
yondonfu: My suggestion was originally motivated by a desire to avoid 2 `_isUnlockInProgress()` checks, but your points make sense. I am ok with repeating a check if it means safer usage of `_cancelUnlock()`

test/unit/TicketBroker.js
eladmallel: 👍

test/unit/TicketBroker.js
eladmallel: agreed.. for some reason that wasn't working out for me so I postponed it for later. Fixed as suggested.

test/unit/TicketBroker.js
eladmallel: totally.. I realized this mistake some time after and forgot to fix!

test/unit/TicketBroker.js
eladmallel: 👍

contracts/pm/TicketBroker.sol
eladmallel: awesome!

test/unit/LivepeerETHTicketBroker.js
```diff
@@ -304,6 +306,30 @@ contract("LivepeerETHTicketBroker", accounts => {
 
                 assert.equal(events.length, 0)
             })
+
+            describe("withdrawTransfer", () => {
```
yondonfu: Shouldn't this be nested under a top level describe block for `withdraw` and not under the describe block for `penaltyEscrowSlash` (which itself is nested under `redeemWinningTicket`)?

eladmallel: yes, I think it should be good in the latest commit!

PR #261, opened 2018-11-23
LivepeerETHTicketBroker

yondonfu:
Updated architecture in `contracts/pm`:

- TicketBroker: Base abstract contract with general ticket operations
- ETHTicketBroker: Tickets denominated in ETH
- ERC20TicketBroker: Tickets denominated in a ERC20 token. Meant to serve as an example for the time being
- LivepeerETHTicketBroker: Tickets denominated in ETH and contract meant to be integrated with the rest of the Livepeer contracts


test/unit/Minter.js
```diff
@@ -281,27 +281,52 @@ contract("Minter", accounts => {
...
-        it("should receive ETH from JobsManager", async () => {
-            await fixture.jobsManager.execute(minter.address, functionSig("depositETH()"), {from: accounts[1], value: 100})
+        it("should receive ETH from TicketBroker", async () => {
+            await fixture.ticketBroker.execute(minter.address, functionSig("trustedDepositETH()"), {from: accounts[1], value: 100})
 
             assert.equal(await web3.eth.getBalance(minter.address), 100, "wrong minter balance")
         })
     })
 
+    describe("trustedBurnETH", () => {
```
eladmallel: Perhaps we should add a test that fails when the system is paused, per the modifier on `trustedBurnETH` ?

contracts/verification/LivepeerVerifier.sol
```diff
@@ -135,10 +135,10 @@ contract LivepeerVerifier is Manager, IVerifier {
 
         // Check if transcoded data hash returned by solver matches originally submitted transcoded data hash
         if (q.commitHash == _result) {
-            IVerifiable(controller.getContract(keccak256("JobsManager"))).receiveVerification(q.jobId, q.claimId, q.segmentNumber, true);
+            IVerifiable(controller.getContract(keccak256("TicketBroker"))).receiveVerification(q.jobId, q.claimId, q.segmentNumber, true);
```
eladmallel: @yondonfu can you please remind me the new verification flow and what changes remain for us to make before these lines of code actually work? (since currently `TicketBroker` does not implement the `IVerifiable` interface)

migrations/3_deploy_contracts.js
```diff
@@ -50,5 +55,8 @@ module.exports = function(deployer, network) {
         // Set RoundsManager parameters
         await roundsManager.setRoundLength(config.roundsManager.roundLength)
         await roundsManager.setRoundLockAmount(config.roundsManager.roundLockAmount)
+
+        // Set TicketBroker parameters
+        await broker.setMinPenaltyEscrow(config.broker.minPenaltyEscrow)
```
eladmallel: @yondonfu should I have added a similar call for setting the `unlockPeriod` ?

migrations/migrations.config.js
```diff
@@ -9,21 +8,16 @@ module.exports = {
...
-        verificationRate: 1000,
-        verificationPeriod: 100,
-        verificationSlashingPeriod: 100,
-        failedVerificationSlashAmount: 1,
-        missedVerificationSlashAmount: 5000,
-        doubleClaimSegmentSlashAmount: 30000,
-        finderFee: 50000
+    broker: {
+        minPenaltyEscrow: (new BN(1)).mul(constants.TOKEN_UNIT).div(new BN(2)),
+        unlockPeriod: new BN(40320) // approximately 7 days worth of blocks
```
eladmallel: just want to re-iterate this is an arbitrary number that we should discuss further before we deploy to testnet.

test/integration/TicketFlow.js
```diff
@@ -0,0 +1,119 @@
...
+        const penaltyEscrow = await broker.minPenaltyEscrow.call()
+
+        await expectThrow(broker.fundPenaltyEscrow({from: broadcaster, value: penaltyEscrow.sub(new BN(1))}))
+
+        await broker.fundPenaltyEscrow({from: broadcaster, value: penaltyEscrow})
+
+        assert.equal(await web3.eth.getBalance(minter.address), deposit.add(penaltyEscrow).toString())
+    })
+
+    it("broadcaster sends a winning ticket", async () => {
```
eladmallel: perhaps add to this description that transcoder redeems it?

test/unit/Minter.js
yondonfu: Added a bunch of tests for failure when the Controller is paused in `Minter.js`

test/integration/TicketFlow.js
yondonfu: Updated in dfc5848

migrations/migrations.config.js
yondonfu: Added a TODO in dfc5848

contracts/verification/LivepeerVerifier.sol
yondonfu: At the moment, there are 2 options for implementing the new verification workflow:

1. Implement the `IVerifiable` interface and the `verify()` function in `LivepeerTicketBroker`

2. Make implementations of the `IVerifier` interface standalone and have a broadcaster directly interact with these contracts instead of going through `LivepeerETHTicketBroker`

I am currently leaning toward 2 for the following reasons:

- Separation of concerns such that `LivepeerETHTicketBroker` doesn't need to know about verification
- Less inter-contract calls. With 1 the contract call flow upon completion of the external computation is external process -> `IVerifier` -> `IVerifiable` -> `BondingManager` (if slashing needs to occur). With 2 the contract call flow is external process -> `IVerifier` -> `BondingManager` (if slashing needs to occur)

Up for discussing the best approach here when we actually get to implementing the new verification workflow

yondonfu:
I removed the remaining references to the JobsManager in `OraclizeVerifier.js` and `OraclizeVerifier.sol`. The code in these files also are not used right now so we could also remove them until we're actually implementing the new verification workflow and considering implementing a verifier that uses Oraclize (as opposed to a centralized Livepeer operated verifier), but I left them for now.

contracts/verification/LivepeerVerifier.sol
eladmallel: I'd love to discuss this further f2f to better understand the solution 2!

contracts/pm/LivepeerETHTicketBroker.sol
```diff
@@ -0,0 +1,96 @@
...
+    {
+        minter().trustedDepositETH.value(msg.value)();
+    }
+
+    function withdrawTransfer(address _sender, uint256 _amount) internal {
+        minter().trustedWithdrawETH(_sender, _amount);
+    }
+
+    function winningTicketTransfer(address _recipient, uint256 _amount) internal {
+        // TODO: Consider changing this to the ticket creation round
```
eladmallel: Just noting that our latest decision here is to indeed change this to the ticket creation round, as we are planning to use rounds as the mechanism for ticket expiration as well.

@yondonfu makes sense?

yondonfu: Yep I think we can remove the todo in the PR that implements the updated ticket expiration logic

PR #263, opened 2018-11-26
PM: fundAndApproveSigners() wrapper 

yondonfu:
- Add fundAndApproveSigners() function to wrap funding deposit and penalty escrow and approving signers
- Add test for separation of ticket redeemer and recipient

### DIFF
--- a/contracts/pm/ERC20TicketBroker.sol
+++ b/contracts/pm/ERC20TicketBroker.sol
@@ -21,6 +21,24 @@ contract ERC20TicketBroker is TicketBroker {
         token = ERC20(_token);
     }
 
+    function fundAndApproveSigners(
+        uint256 _depositAmount,
+        uint256 _penaltyEscrowAmount,
+        address[] _signers
+    )
+        external
+        checkDepositPenaltyEscrowSplit(_depositAmount, _penaltyEscrowAmount)
+        processDeposit(msg.sender, _depositAmount)
+        processPenaltyEscrow(msg.sender, _penaltyEscrowAmount)
+    {
+        approveSigners(_signers);
+
+        require(
+            token.transferFrom(msg.sender, address(this), _depositAmount + _penaltyEscrowAmount),
+            "token transfer for deposit and penalty escrow failed"
+        );
+    }
+
     function fundDeposit(uint256 _amount) 
         external
         processDeposit(msg.sender, _amount)
--- a/contracts/pm/ERC20TicketBroker.sol
+++ b/contracts/pm/ERC20TicketBroker.sol
@@ -27,7 +27,6 @@ contract ERC20TicketBroker is TicketBroker {
         address[] _signers
     )
         external
-        checkDepositPenaltyEscrowSplit(_depositAmount, _penaltyEscrowAmount)
         processDeposit(msg.sender, _depositAmount)
         processPenaltyEscrow(msg.sender, _penaltyEscrowAmount)
     {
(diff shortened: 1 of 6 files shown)


contracts/pm/ERC20TicketBroker.sol
```diff
@@ -21,6 +21,24 @@ contract ERC20TicketBroker is TicketBroker {
         token = ERC20(_token);
     }
 
+    function fundAndApproveSigners(
+        uint256 _depositAmount,
+        uint256 _penaltyEscrowAmount,
+        address[] _signers
+    )
+        external
+        checkDepositPenaltyEscrowSplit(_depositAmount, _penaltyEscrowAmount)
```
eladmallel: won't this always fail because it checks against `msg.value` which won't be sufficient in an ERC20 approach?

test/unit/LivepeerETHTicketBroker.js
```diff
@@ -181,6 +181,107 @@ contract("LivepeerETHTicketBroker", accounts => {
         })
     })
 
+    describe("fundAndApproveSigners", () => {
```
eladmallel: should we add tests that ensure events are emitted?

eladmallel:
@yondonfu notice how we're forced to several identical tests to cover both the ETH broker and the LivepeerETH broker. Perhaps we should further discuss how to get rid of this duplication?

contracts/pm/ERC20TicketBroker.sol
yondonfu: Right - we actually don't need to check the deposit/penalty escrow split in ERC20TicketBroker since the contract doesn't need to make sure that the correct value is included in the tx. If the user did not approve a sufficient amount of tokens to be transferred by ERC20TicketBroker to cover both the specified deposit and penalty escrow amounts, the ERC20 `transferFrom` call will fail.

I removed the modifier from ERC20TicketBroker and moved it from TicketBroker into ETHTicketBroker and LivepeerETHTicketBroker in 00f5dfc since the modifier logic does not necessarily apply to all TicketBroker implementations (i.e. ERC20TicketBroker). However, this does introduce duplication of the modifier in LivepeerETHTicketBroker and ETHTicketBroker. Let me know if you have any thoughts on a better approach here.

test/unit/LivepeerETHTicketBroker.js
yondonfu: There are already tests for events emitted from the modifiers and `approveSigners` - since this function just internally calls those modifiers and `approveSigners` I did not include tests for events emitted in the interest of avoiding additional duplication. WDYT?

contracts/pm/ERC20TicketBroker.sol
eladmallel: We could just decide to leave the modifier in `TicketBroker` even though not every broker type needs it, if we choose to prioritize DRY (don't repeat yourself) over maximizing sub-type reuse of super type code. 

I'm leaning in that direction.. to sweeten the deal we could rename the modifier to explicitly mention it's using ETH value, to reduce the odds of misuse.

wdyt?

test/unit/LivepeerETHTicketBroker.js
eladmallel: Well, you did add tests for tracking balances, which is also logic that lives inside the modifiers :) so we're testing one part of the modifiers' behavior but not the other.

I share the sentiment that this is redundant, but find it harder to let go of the inconsistency.

Perhaps we should consider dropping the test for balance tracking?
Or re-discuss how we could design these contracts to achieve more straightforward code reuse that doesn't force us to duplicate a bunch of testing code?

contracts/pm/ERC20TicketBroker.sol
yondonfu: I don't have any strong feelings here, so I am happy to leave the modifier in TicketBroker and update its name to reflect that it deals with ETH value.

PR #264, opened 2018-12-11
PM: TicketBroker inheritance refactor

eladmallel:
So far we've used two different approaches for reusing code from
`TicketBroker` in its subtypes: one where only subtypes implement
functions and use modifiers to reuse `TicketBroker`'s flow of
requirements and accounting, and another approach where `TicketBroker`
implements a public/external function that contains the entire flow,
and calls an abstract method that each subtype implements according to
their needs.

The first approach (using modifiers) resulted in code duplication, i.e.
having to use the exact same modifiers in more than one contract, which
also resulted in duplicate test code.

To get out of this code duplication we're refactoring to the second
approach across the board.

The main cost is the complication introduced in the ERC20 implementation,
but since it's just an example and not a priority, it seems like the
better choice, for the benefit of simplifying the more mainstream
usecase with ETH.

### DIFF
--- a/contracts/pm/ERC20TicketBroker.sol
+++ b/contracts/pm/ERC20TicketBroker.sol
@@ -21,37 +21,45 @@ contract ERC20TicketBroker is TicketBroker {
         token = ERC20(_token);
     }
 
+    // For ETH
+    function fundDeposit() external payable {
+        require(false, "ETH funding not supported. Please use an ERC20 token instead.");
+    }
+
+    function fundPenaltyEscrow() external payable {
+        require(false, "ETH funding not supported. Please use an ERC20 token instead.");
+    }
+
+    // For ERC20
+    function fundDeposit(uint256 _amount) external {
+        _fundDeposit(_amount);
+    }
+
+    function fundPenaltyEscrow(uint256 _amount) external {
+        _fundPenaltyEscrow(_amount);
+    }
+
     function fundAndApproveSigners(
         uint256 _depositAmount,
         uint256 _penaltyEscrowAmount,
         address[] _signers
     )
         external
-        processDeposit(msg.sender, _depositAmount)
-        processPenaltyEscrow(msg.sender, _penaltyEscrowAmount)
+        payable
     {
+        _fundDeposit(_depositAmount);
+        _fundPenaltyEscrow(_penaltyEscrowAmount);
         approveSigners(_signers);
-
-        require(
-            token.transferFrom(msg.sender, address(this), _depositAmount + _penaltyEscrowAmount),
-            "token transfer for deposit and penalty escrow failed"
-        );
     }
 
-    function fundDeposit(uint256 _amount) 
-        external
-        processDeposit(msg.sender, _amount)
-    {
+    function processDeposit(uint256 _amount) internal {
         require(
             token.transferFrom(msg.sender, address(this), _amount),
             "token transfer for deposit failed"
         );
     }
 
-    function fundPenaltyEscrow(uint256 _amount)
-        external
-        processPenaltyEscrow(msg.sender, _amount)
-    {
+    function processPenaltyEscrow(uint256 _amount) internal {
         require(
             token.transferFrom(msg.sender, address(this), _amount),
             "token transfer for penalty escrow failed"
--- a/contracts/pm/ERC20TicketBroker.sol
+++ b/contracts/pm/ERC20TicketBroker.sol
@@ -21,22 +21,19 @@ contract ERC20TicketBroker is TicketBroker {
         token = ERC20(_token);
     }
 
-    // For ETH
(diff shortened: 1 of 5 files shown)


contracts/pm/TicketBroker.sol
```diff
@@ -57,35 +57,31 @@ contract TicketBroker {
...
+    function fundAndApproveSigners(
+        uint256 _depositAmount,
+        uint256 _penaltyEscrowAmount,
+        address[] _signers
+    )
+        external
+        payable
+        checkDepositPenaltyEscrowETHValueSplit(_depositAmount, _penaltyEscrowAmount)
+    {
+        _fundDeposit(_depositAmount);
```
yondonfu: Noting that now that we are not using modifiers, we no longer adhere to the checks -> state updates -> external calls pattern here since `_fundDeposit()` and `_fundPenaltyEscrow()` potentially include external calls depending on the override behavior of `processDeposit()` and `processPenaltyEscrow()`.

Not following the pattern for this wrapper function might be ok as long as we remain cautious of reentrancy attacks. This is not a problem for ETHTicketBroker (which does not make any external calls) and LivepeerETHTicketBroker (since the external call is to a trusted contract - namely the Minter contract that is part of the Livepeer contract system). This might be a problem for ERC20TicketBroker if a developer deployed the contract and pointed it to an untrusted token contract.

In addition to no longer using the checks -> state updates -> external calls pattern, this wrapper function can end up executing 2 external calls when it only needs to execute 1. For example, in LivepeerETHTicketBroker, this function will execute `minter().trustedDeposit.value()` twice, once for `_depositAmount` and once for `_penaltyEscrowAmount`. Instead, this function could execute `minter().trustedDeposit.value()` once with `_depositAmount + _penaltyEscrowAmount` which was the original behavior when we were using modifiers. This is also the case for ERC20TicketBroker which will execute 2 ERC20 `transferFrom()` when it only needs to execute 1.

contracts/pm/ERC20TicketBroker.sol
```diff
@@ -21,37 +21,45 @@ contract ERC20TicketBroker is TicketBroker {
...
+
     function fundAndApproveSigners(
         uint256 _depositAmount,
         uint256 _penaltyEscrowAmount,
         address[] _signers
     )
         external
-        processDeposit(msg.sender, _depositAmount)
-        processPenaltyEscrow(msg.sender, _penaltyEscrowAmount)
+        payable
```
yondonfu: Add `require(msg.value == 0, ERROR_MSG)` here to reject calls that include ETH?

contracts/pm/TicketBroker.sol
```diff
@@ -212,6 +234,12 @@ contract TicketBroker {
         emit UnlockCancelled(_senderAddress);
     }
 
+    // Override
+    function processDeposit(uint256 _amount) internal;
```
yondonfu: Is there a scenario where `processDeposit()` and `processPenaltyEscrow()` would have different behavior? What if we just had a single internal function that subtypes should be override that is called by `_fundDeposit()` and `_fundPenaltyEscrow()`?

contracts/pm/ERC20TicketBroker.sol
```diff
@@ -21,37 +21,45 @@ contract ERC20TicketBroker is TicketBroker {
         token = ERC20(_token);
     }
 
+    // For ETH
+    function fundDeposit() external payable {
+        require(false, "ETH funding not supported. Please use an ERC20 token instead.");
```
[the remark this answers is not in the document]
eladmallel: 👍

contracts/pm/ERC20TicketBroker.sol
```diff
@@ -21,37 +21,45 @@ contract ERC20TicketBroker is TicketBroker {
...
+    }
+
+    function fundPenaltyEscrow() external payable {
+        require(false, "ETH funding not supported. Please use an ERC20 token instead.");
```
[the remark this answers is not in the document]
eladmallel: 👍

contracts/pm/ERC20TicketBroker.sol
eladmallel: 👍

contracts/pm/TicketBroker.sol
eladmallel: I think you're right. At the moment there is absolutely no difference, so yes I will make the change as you suggested. Thanks!

contracts/pm/TicketBroker.sol
eladmallel: Agreed and tried to resolve it all in the latest commit https://github.com/livepeer/protocol/pull/264/commits/4c06692eb7bdb38407f80a4b64a3963bf9293e84

Want to check it out?

PR #265, opened 2018-12-12
PM: senders can revoke signers

eladmallel:
We decided to create a separate param for the signer revocation delay
to decouple this concern from the `unlockPeriod`.
The reason we need a delayed revocation in the first place, is to allow
receivers of tickets to become aware of a pending revocation and make
sure they cash in any tickets signed by the soon-to-be-revoked signer
before those tickets become invalid.

### DIFF
--- a/contracts/pm/TicketBroker.sol
+++ b/contracts/pm/TicketBroker.sol
@@ -7,12 +7,16 @@ import "openzeppelin-solidity/contracts/cryptography/ECDSA.sol";
 
 contract TicketBroker {
 
+    struct Signer {
+        bool approved;
+        uint256 revocationBlock;
+    }
+
     struct Sender {
         uint256 deposit;
         uint256 penaltyEscrow;
         uint256 withdrawBlock;
-        mapping (address => bool) signers;
-        mapping (address => uint256) signersRevocationBlock;
+        mapping (address => Signer) signers;
     }
 
     struct Ticket {
@@ -127,7 +131,9 @@ contract TicketBroker {
         Sender storage sender = senders[msg.sender];
 
         for (uint256 i = 0; i < _signers.length; i++) {
-            sender.signers[_signers[i]] = true;
+            Signer storage signer = sender.signers[_signers[i]];
+            signer.approved = true;
+            signer.revocationBlock = 0;
         }
 
         emit SignersApproved(msg.sender, _signers);
@@ -138,7 +144,7 @@ contract TicketBroker {
         uint256 revocationBlock = block.number + signerRevocationPeriod;
 
         for (uint256 i = 0; i < _signers.length; i++) {
-            sender.signersRevocationBlock[_signers[i]] = revocationBlock;
+            sender.signers[_signers[i]].revocationBlock = revocationBlock;
         }
 
         emit SignersRevocationRequested(msg.sender, _signers, revocationBlock);
@@ -247,8 +253,8 @@ contract TicketBroker {
     }
 
     function isApprovedSigner(address _sender, address _signer) public view returns (bool) {
-        uint256 revocationBlock = senders[_sender].signersRevocationBlock[_signer];
-        return senders[_sender].signers[_signer] && (revocationBlock == 0 || block.number < revocationBlock);
+        Signer memory signer = senders[_sender].signers[_signer];
+        return signer.approved && (signer.revocationBlock == 0 || block.number < signer.revocationBlock);
     }
 
     function _cancelUnlock(Sender storage _sender, address _senderAddress) internal {
--- a/contracts/pm/TicketBroker.sol
+++ b/contracts/pm/TicketBroker.sol
@@ -12,6 +12,7 @@ contract TicketBroker {
         uint256 penaltyEscrow;
         uint256 withdrawBlock;
         mapping (address => bool) signers;
+        mapping (address => uint256) signersRevocationBlock;
     }
 
     struct Ticket {
@@ -26,6 +27,7 @@ contract TicketBroker {
 
     uint256 public minPenaltyEscrow;
     uint256 public unlockPeriod;
+    uint256 public signerRevocationPeriod;
 
     mapping (address => Sender) public senders;
     mapping (bytes32 => bool) public usedTickets;
@@ -33,6 +35,7 @@ contract TicketBroker {
     event DepositFunded(address indexed sender, uint256 amount);
     event PenaltyEscrowFunded(address indexed sender, uint256 amount);
     event SignersApproved(address indexed sender, address[] approvedSigners);
+    event SignersRevocationRequested(address indexed sender, address[] signers, uint256 revocationBlock);
     event WinningTicketRedeemed(
         address indexed sender,
         address indexed recipient,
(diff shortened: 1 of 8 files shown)


test/unit/LivepeerETHTicketBroker.js
```diff
@@ -14,12 +15,13 @@ contract("LivepeerETHTicketBroker", accounts => {
     const recipient = accounts[1]
 
     const unlockPeriod = 20
+    const signerRevocationPeriod = 10
```
yondonfu: Any reason for this value to be different than `unlockPeriod`? (Not a big deal, but if there isn't a reason then perhaps we just use the same constant values when possible to avoid future speculation on whether there is a reason for some value in tests or not)

test/unit/LivepeerETHTicketBroker.js
```diff
@@ -243,4 +245,19 @@ contract("LivepeerETHTicketBroker", accounts => {
             })
         })
     })
+
+    describe("setSignerRevocationPeriod", () => {
+        it("reverts when called by an account that is not the owner", async () => {
```
yondonfu: Small nit: We're really checking that the sender is the Controller's owner rather than the owner of LivepeerETHTicketBroker

contracts/pm/TicketBroker.sol
```diff
@@ -129,6 +133,17 @@ contract TicketBroker {
         emit SignersApproved(msg.sender, _signers);
     }
 
+    function requestSignersRevocation(address[] _signers) public {
+        Sender storage sender = senders[msg.sender];
+        uint256 revocationBlock = block.number + signerRevocationPeriod;
+
+        for (uint256 i = 0; i < _signers.length; i++) {
+            sender.signersRevocationBlock[_signers[i]] = revocationBlock;
```
yondonfu: Should we set the revocation block for a signer address that was never approved? I lean towards no and skipping if the signer address was not previously approved

eladmallel: I had this debate with myself. Here's why I thought it's ok to leave it as-is:

- I expect nearly 100% of the time that senders will only ask to invoke senders they previously approved
- Checking if this sender was approved costs a bit of gas
- To minimize gas costs for the vast majority of use cases I thought to omit this check
- In a small fraction of use cases senders will send signers that were never approved and waste some storage writing gas

So bottom line, I was aiming at the small gas optimization for the main use case.

wdyt?

test/unit/LivepeerETHTicketBroker.js
eladmallel: 👍

test/unit/LivepeerETHTicketBroker.js
eladmallel: No special reason. I set a different value just because I thought to signal to readers that this param won't necessarily be the same duration as `unlockPeriod`, but simply the fact that you didn't like it when you read it is sufficient reason to change it to be the same :)

contracts/pm/TicketBroker.sol
```diff
@@ -129,6 +133,17 @@ contract TicketBroker {
         emit SignersApproved(msg.sender, _signers);
     }
 
+    function requestSignersRevocation(address[] _signers) public {
```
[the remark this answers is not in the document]
eladmallel: of course, totally my bad for submitting this PR when tired!

contracts/pm/TicketBroker.sol
```diff
@@ -12,6 +12,7 @@ contract TicketBroker {
         uint256 penaltyEscrow;
         uint256 withdrawBlock;
         mapping (address => bool) signers;
+        mapping (address => uint256) signersRevocationBlock;
```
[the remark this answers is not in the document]
eladmallel: of course, totally my bad for submitting this PR when tired!

contracts/pm/TicketBroker.sol
yondonfu: Makes sense. I just wanted to make sure that the state transition from not approved to revocation in progress doesn't present any problems.

I don't think it does. The states for signers are:

- `NOT_APPROVED`
- `APPROVED`
- `REVOCATION_IN_PROGRESS`
- `REVOKED`

In these scenarios, the following state transitions for a signer is possible:

`approveSigners()`: `NOT_APPROVED` -> `APPROVED` OR `REVOKED` -> `APPROVED` OR `REVOCATION_IN_PROGRESS` -> `APPROVED`
`requestSignersRevocation()`: `APPROVED` -> `REVOCATION_IN_PROGRESS` OR `NOT_APPROVED` -> `REVOCATION_IN_PROGRESS`
`block.number >= revocationBlock`: `REVOCATION_IN_PROGRESS` -> `REVOKED`

test/unit/LivepeerETHTicketBroker.js
```diff
@@ -243,4 +245,19 @@ contract("LivepeerETHTicketBroker", accounts => {
             })
         })
     })
+
+    describe("setSignerRevocationPeriod", () => {
+        it("reverts when called by an account that is not the Controller owner", async () => {
+            await expectThrow(broker.setSignerRevocationPeriod(1234, {from: accounts[5]}))
+        })
+
+        it("works when called by the owner", async () => {
```
yondonfu: "owner" -> "Controller owner"?

test/unit/TicketBroker.js
```diff
@@ -257,6 +269,80 @@ contract("TicketBroker", accounts => {
         })
     })
 
+    describe("requestSignersRevocation", () => {
```
yondonfu: What do you think about adding a describe block for `isApprovedSigner()` and moving the test cases involving the state transition from `REVOCATION_IN_PROGRESS` -> `REVOKED` in that section (and moving them out of this section)? I mention that because it feels like that particular state transition isn't a direct result of `requestSignersRevocation()` which is just responsible for the state transition to `REVOCATION_IN_PROGRESS`.

I think we could also add two distinct test cases for when the signerRevocationPeriod elapses:
- When `block.number == signer.revocationBlock`
- When `block.number > signer.revocationBlock`

eladmallel: > What do you think about adding a describe block for isApprovedSigner() and moving the test cases involving the state transition

I much prefer to stick with tests that cover the "collaboration" of these two functions. Moving them results in two options that I can see:

1. Simply move all state-transition tests under a different describe block - that's fine, but I don't think we gain much.
2. Split tests up such that under `requestSignersRevocation` we run assertions on `revocationBlock`, and then have similar test cases for `isApprovedSigner` that have the same setup, but then assert in `isApprovedSigner`'s output - this solution seems a bit redundant to me.

So I'm actually inclined to leave the grouping as-is, and perhaps to just rename the describe block to `requestSignersRevocation and isApprovedSigner` since it does indeed encapsulate logic from both functions.

> I think we could also add two distinct test cases for when the signerRevocationPeriod elapses

I agree those are two distinct cases, but the reason I haven't added them so far is because I didn't think this level of granularity mattered much in terms of UX and data integrity. I think it's more important to make sure we cover `block.number > signer.revocationBlock`, so perhaps it's a good idea to change the tests to wait `await fixture.rpc.wait(signerRevocationPeriod + 1)`.

wdyt?

yondonfu: If `isApprovedSigner()` is fully tested (i.e. all code paths are hit) then I am ok with not having an additional describe block for the function if it results in a lot of duplication. I'd suggest keeping the `requestSignersRevocation` name for the describe block though as `requestSignersRevocation and isApprovedSigner` seems (at least to me) to imply that these are all the unit test cases for these 2 functions when in fact `isApprovedSigner()` is also used in the describe block for `approveSigners()` which includes some unit test cases for `isApprovedSigner()` as well. Keeping the name seems ok to me.

> I agree those are two distinct cases, but the reason I haven't added them so far is because I didn't think this level of granularity mattered much in terms of UX and data integrity.

I like covering both distinct cases for the sake of completeness i.e. are we absolutely certain that the same result holds true for both distinct cases?

PR #270, opened 2018-12-13
PM with SafeMath

eladmallel:


eladmallel:
Just note that this is branched off of https://github.com/livepeer/protocol/pull/269

contracts/pm/ERC20TicketBroker.sol
```diff
@@ -2,9 +2,9 @@ pragma solidity ^0.4.25;
...
 pragma experimental ABIEncoderV2;
 
-import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 import "./TicketBroker.sol";
-
+import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
 // JUST AN EXAMPLE
 contract ERC20TicketBroker is TicketBroker {
```
yondonfu: Missing a `using SafeMath for uint256`?

contracts/pm/TicketBroker.sol
```diff
@@ -175,7 +176,7 @@ contract TicketBroker {
             sender.penaltyEscrow = 0;
         } else {
             amountToTransfer = _ticket.faceValue;
-            sender.deposit -= _ticket.faceValue;
+            sender.deposit = sender.deposit.sub(_ticket.faceValue);
```
yondonfu: Due to the conditional on L171 (https://github.com/livepeer/protocol/pull/270/commits/db86d2bfb473523d8ada69ae2297bbe3ecf5df6d#diff-baed5767688fca63ab8b4efd937ce4ccR171) we know that this subtraction could never underflow so using SafeMath here just results in some extra unnecessary checks. I do see an argument for applying SafeMath everywhere for consistency sake. What do you think?

eladmallel: I was prioritizing consistency :) If that's cool with you I'm happy to keep this change as-is. wdyt?

contracts/pm/ERC20TicketBroker.sol
eladmallel: Your EDIT is spot on. I removed the imports and using statements in subtypes.

contracts/pm/TicketBroker.sol
yondonfu: Consistency works for me!

Issue #284, opened 2019-04-22
Single approved signer registry


yondonfu:
Orchestrators and broadcaster can use a single registry to approve/revoke signers that sign off-chain data on behalf of their main ETH account. Any contract that supports approved signers should be aware of this registry and during signature validation should:

- Check if the recovered address from a signature matches the provided address
- If not, check if the recovered address is an approved signer in the registry for the provided address

Issue #246, opened 2018-11-19
Used ticket replay attack defense


yondonfu:
We need to prevent recipients from cashing in the same winning ticket on-chain multiples times - each winning ticket can only be cashed in once.

One way to defend against replay attacks is to have the contract maintain a mapping of ticket IDs corresponding to cashed tickets. We can define a ticket ID as the hash of all the relevant metadata in the ticket including the sender specified random nonce so each ticket ID should be unique. When a recipient cashes in a winning ticket, the contract will check if the ticket ID has already been used and if so, the transaction will revert. 

The downside of this solution is that it adds an additional 20000 gas for an SSTORE whenever a recipient cashes a winning ticket thereby increasing the transaction cost.

Other ways to defend against replay attacks include:

- Instead of storing the ticket ID on-chain the contract just emits an event containing the ticket ID. If we assume that the recipients have some economic value staked either inside or outside (i.e. if the probabilistic payment contract is used as a part of a larger contract system) then anyone that submits a ticket ID that was used twice can penalize the recipient that replayed the ticket ID. This would require either a challenge period after a winning ticket is initially submitted or the sender would just lose funds from the replay attack, but the attacker would be slashed later on once the evidence of a replay is submitted.
- Instead of storing each ticket ID in an on-chain mapping we could use a cryptographic accumulator that supports proofs of non-membership - the accumulator would be represented by a short value stored in the contract and can be used along with a proof of non-membership to demonstrate that a ticket ID has not yet been used. Merkle trees are probably not a good candidate because updating the Merkle tree requires knowledge of all the existing leaves and keeping track of all the leaves on-chain would be more expensive than the single SSTORE required for the original on-chain mapping solution. RSA accumulators might work because they support updates without knowledge of the other members of the set, but these are not as well explored for usage in contracts (research is picking up though with their proposed usage in Plasma Cash). We would need to figure out if using such an accumulator and verifying a proof of non-membership is cheap enough such that it is makes sense to use an accumulator instead of an on-chain mapping.

The original solution should work for an initial implementation and research into optimizations can be left for subsequent iterations.

eladmallel:
@yondonfu very interesting idea with the accumulator. Wondering about the gas costs of this approach - wouldn't it be the same gas cost as our current naive approach?

The way I understand the accumulator approach is that for every submitted winning ticket we need to update the accumulator value for that redeemer. This would result in the same SSTORE cost as adding a ticket hash to a mapping, right?

[...]

PR #286, opened 2019-05-01
pm: Refactor TicketBroker and implement reserve logic

### LINKED ISSUE #277: pm: Implement fundReserve() and fundDepositAndReserve()
The `TicketBroker` contract should implement [fundReserve()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#funddeposit) and [fundDepositAndReserve()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#funddepositandreserve). This will also involve creating the [Reserve](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#reserve) data structure.

### LINKED ISSUE #278: pm: Implement reserve claiming in redeemWinningTicket()
The `TicketBroker` contract's `redeemWinningTicket()` should claim from a sender's reserve according to the [reserve claiming spec](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#reserve-claiming).

### LINKED ISSUE #279: pm: Update relevant functions to handle frozen reserve
Many of the `TicketBroker` contract functions need to execute different logic if a sender's reserve is frozen. These functions include:

- [fundDeposit()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#funddeposit)
- [unlock()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#unlock)
- [cancelUnlock()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#cancelunlock)
- [withdraw()](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#withdraw)

### LINKED ISSUE #280: pm: If old reserve was frozen, migrate remaining funds in fundReserve() and fundReserveAndDeposit()
If a sender's old reserve was frozen and the sender is calling `fundReserve()` or `fundReserveAndDeposit()` after the freeze period, the `TicketBroker` contract should migrate any remaining funds from the sender's old reserve into the sender's new reserve (see the description in the [spec](https://github.com/livepeer/wiki/blob/master/spec/streamflow/pm.md#fundreserve)).

yondonfu:
Closes #277 
Closes #278 
Closes #279 
Closes #280 

- Refactored TicketBroker to use mixins
- Removed non-Livepeer versions of TicketBroker for simplicity
- Removed approved signer logic from TicketBroker in preparation for #284 
- Added reserve logic to TicketBroker

Apologizes for the large commit - I originally split up this PR into different commits for each of the reserve related features, but the refactor resulted in splitting up the original TicketBroker contract into multiple mixin contracts across different files.


dob:
Saw I'm requested as a reviewer here. Thanks for looping me in, but I don't want to be a blocker towards progress. If it's really helpful to get my input at this time, grab me in person for a walkthrough. Thanks!

yondonfu:
> I did some envelope calculation - 10000 concurrent streams at 1/1000 winning percentage over a year yields 2.5G of onchain claimed ticket data.

Is this calculation for keeping track of used tickets on-chain? i.e. marking a winning ticket as used in the contract when it is redeemed

Here is my quick calculation:

```
# Assuming 2 second segments
ticketsInAYear = 10000 streams * ((365 * 24 * 60 * 60) / 2) segments = 157680000000
# 1 out of 1000 tickets win
winningTicketsInAYear = 157680000000 tickets / 1000 = 157680000
# 32 bytes for each ticket
bytesInAYear = 157680000 winning tickets * 32 bytes = 5045760000
gbInAYear = 5045760000 bytes / 1000000000 = 5.04576
```

Storing 32 bytes for each used winning ticket is definitely not great and might actually be problematic in the future if storage rent is introduced (i.e. maintenance costs for a contract based on storage used). While I think this approach to replay attack defense works for now, there is some brainstorming for alternatives in https://github.com/livepeer/protocol/issues/246 that can be a basis for further thinking.

dob:
Pointing out that 1 out of 1000 winning ticks doesn’t make sense to me. That is 2000 seconds or 30 minutes, or something like only 5 cent payout at 10 cent/hr rate. I’m suspecting that to minimize on chain txn costs then winning ticket face values would be like 100-1000x that and therefore 100-1000x less frequent.


________________________________

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -0,0 +1,188 @@
...
+    {
+        ReserveManager storage manager = reserveManagers[_reserveHolder];
+        Reserve storage reserve = manager.reserve;
+
+        // If reserve is not frozen then freeze it
+        if (reserveState(_reserveHolder) == ReserveState.NotFrozen) {
+            uint256 freezeRound = roundsManager().currentRound();
+            // TODO: make sure bondingManager.getTranscoderPoolSize()
+            // returns the locked in # registered for the freeze round
+            uint256 recipientsInFreezeRound = bondingManager().getTranscoderPoolSize();
```
ericxtang: Should we get the size for the entire transcoder pool or the active set here?

yondonfu: We want the size of the entire transcoder pool here because the entire transcoder pool will be the active set once we implement the staking protocol changes. As mentioned in the TODO, we'll need to make sure that there is a way to get the locked in transcoder pool size for a round when we're implementing those changes

PR #288, opened 2019-05-03
pm: Batch ticket redemption

### LINKED ISSUE #274: Batch ticket redemption
The TicketBroker should expose a function for batch ticket redemption i.e. redeem multiple tickets in a single atomic transaction instead of submitting separate transactions for each ticket. Ideally, this function would validate all the tickets, sum the face values of each ticket and finally execute a single value transfer at the end. The simpler more naive approach would be to execute a value transfer for each ticket, but these additional value transfers are unnecessary and just increase gas usage.

yondonfu:
Closes #274 

`batchRedeemWinningTickets()` uses the `redeemWinningTicketNoRevert()` helper to redeem tickets without reverting the entire function when execution hits a revert in the underlying call. 

This PR does not implement the suggestion in #274 of validating each ticket, summing the face values of the tickets and then executing a single batch value transfer which would save some gas. Instead, each ticket redemption is distinct and involves its own associated value transfer. I chose to do this to keep things simple initially. Furthermore, in a Livepeer context, the value transfer involves sending the value to the recipient's (orchestrator) fee pool for the ticket's creation round and in cases where a batch of tickets have different creation rounds, there would need to be a separate value transfer for each creation round anyway. So, I'm leaving an investigation of gas related optimizations for later on.

### DIFF
--- a/contracts/pm/mixins/MixinWrappers.sol
+++ b/contracts/pm/mixins/MixinWrappers.sol
@@ -1,0 +1,72 @@
+pragma solidity ^0.4.25;
+// solium-disable-next-line
+pragma experimental ABIEncoderV2;
+
+import "./interfaces/MTicketBrokerCore.sol";
+
+
+contract MixinWrappers is MTicketBrokerCore {
+    /**
+     * @dev Redeems multiple winning tickets. The function will redeem all of the provided
+     * tickets and handle any failures gracefully without reverting the entire function
+     * @param _tickets Array of winning tickets to be redeemed in order to claim payment
+     * @param _sigs Array of sender signatures over the hash of tickets (`_sigs[i]` corresponds to `_tickets[i]`)
+     * @param _recipientRands Array of preimages for the recipientRandHash included in each ticket (`_recipientRands[i]` corresponds to `_tickets[i]`)
+     */
+    function batchRedeemWinningTickets(
+        Ticket[] memory _tickets,
+        bytes[] _sigs,
+        uint256[] _recipientRands
+    )
+        public
+    {
+        for (uint256 i = 0; i < _tickets.length; i++) {
+            redeemWinningTicketNoRevert(
+                _tickets[i],
+                _sigs[i],
+                _recipientRands[i]
+            );
+        }
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


contracts/pm/mixins/MixinWrappers.sol
```diff
@@ -0,0 +1,72 @@
...
+            "redeemWinningTicket((address,address,uint256,uint256,uint256,bytes32,bytes),bytes,uint256)",
+            _ticket,
+            _sig,
+            _recipientRand
+        );
+
+        // Call `redeemWinningTicket()`
+        assembly {
+            // call will return false upon hitting a revert
+            success := call(
```
ericxtang: A little out of my depth here - we are doing this assembly call because we don't want revert?

yondonfu: Yeah. When you use either a low level call (i.e. invoking a call on an address and generating the calldata yourself) or an assembly based call, if the called function reverts, the call itself will return `false` instead of reverting (see: https://solidity.readthedocs.io/en/v0.4.25/units-and-global-variables.html#address-related). We can use this non-reverting helper function to define a wrapper batch redemption function without changing the actual ticket redemption code.

We could actually use a low level call instead of dropping down to assembly here, but if we end up changing `redeemWinningTicket()` to actually return a value, then we would need to drop down to assembly in order to fetch the return data because the low level call does not give access to the return data. So, I just left the assembly based call which I think is pretty straightforward and self-contained.

ericxtang: I see.  That makes sense.  What do you think about keeping track of which ticket failed or emitting an event for failed tickets?  I think in the current workflow, we'll only see successful ticket redemptions through the `WinningTicketRedeemed` event, but failures will be silent.

yondonfu: I'm not sure that emitting an additional event is worth the gas cost because given a set of tickets that are passed in for batch redemption a client can check the transaction receipt for `WinningTicketRedeemed` events and then any tickets that do not have an associated event can be marked by the client as having failed

PR #291, opened 2019-05-28
Protocol spec broken link

kyriediculous:
Fixed original protocol spec broken link (tributary and snowmelt) in Readme.MD


yondonfu:
Need to get CircleCI running builds for forks...will merge after

kyriediculous:
I'll close this PR considering there's a second one which is not based from a fork that does the same thing.

PR #294, opened 2019-05-30
pm: claimableReserve function & tests

kyriediculous:

Implements a method `claimableReserve(address _claimaint, address _reserveHolder) returns (uint256 claimableReserve)` in `MixinReserve.sol` which returns the amount of funds claimable by a claimant/orchestrator from a reserveHolder/broadcaster's reserve

Added unit tests for the added method 

**Additional Information (TODO)**

* At the moment we do not check if the claimant is actually registered as orchestrator in the current round (if the reserveHolder's reserve is not frozen) or that the claimant was/is registered as orchestrator during the freezeRound in case a reserveHolder's reserve has been frozen. 
This is due to the corresponding functionality not being implemented yet in BondingManager

* Unit tests once that functionality is implemented in the BondingManager.

### DIFF
--- a/contracts/pm/mixins/MixinReserve.sol
+++ b/contracts/pm/mixins/MixinReserve.sol
@@ -43,6 +43,24 @@ contract MixinReserve is MContractRegistry, MReserve {
     }
 
     /**
+     * @dev Returns the amount of funds claimable by a claimant from a reserve
+     * @param _reserveHolder Address of reserve holder
+     * @param _claimant Address of claimant
+     * @return Amount of funds claimable by `_claimant`from the reserve for `_reserveHolder`
+     */
+    function claimableReserve(address _reserveHolder, address _claimant) public view returns (uint256) {
+        ReserveManager storage manager = reserveManagers[_reserveHolder];
+        Reserve storage reserve = manager.reserve;
+        if (reserveState(_reserveHolder) == ReserveState.NotFrozen) {
+            // TODO: Check if claimant is registered in current round
+            return reserve.fundsAdded.div(bondingManager().getTranscoderPoolSize());
+        } else {
+            // TODO: Check if claimant is registered during the freezeRound
+            return reserve.fundsAdded.div(reserve.recipientsInFreezeRound).sub(manager.claimedPerReserve[manager.reserveNonce][_claimant]);
+        }
+    }
+
+    /**
      * @dev Returns the amount of funds claimed by a claimant from a reserve
      * @param _reserveHolder Address of reserve holder
      * @param _claimant Address of claimant
--- a/contracts/pm/mixins/MixinReserve.sol
+++ b/contracts/pm/mixins/MixinReserve.sol
@@ -46,17 +46,20 @@ contract MixinReserve is MContractRegistry, MReserve {
      * @dev Returns the amount of funds claimable by a claimant from a reserve
      * @param _reserveHolder Address of reserve holder
      * @param _claimant Address of claimant
-     * @return Amount of funds claimable by `_claimant`from the reserve for `_reserveHolder`
+     * @return Amount of funds claimable by `_claimant` from the reserve for `_reserveHolder`
      */
     function claimableReserve(address _reserveHolder, address _claimant) public view returns (uint256) {
         ReserveManager storage manager = reserveManagers[_reserveHolder];
         Reserve storage reserve = manager.reserve;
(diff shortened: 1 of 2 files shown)


contracts/pm/mixins/MixinReserve.sol
```diff
@@ -42,6 +42,24 @@ contract MixinReserve is MContractRegistry, MReserve {
         info.thawRound = reserve.freezeRound == 0 ? 0 : reserve.freezeRound.add(freezePeriod);
     }
 
+    /**
+     * @dev Returns the amount of funds claimable by a claimant from a reserve
+     * @param _reserveHolder Address of reserve holder
+     * @param _claimant Address of claimant
+     * @return Amount of funds claimable by `_claimant`from the reserve for `_reserveHolder`
```
yondonfu: Space between `_claimant` and "from"

test/unit/TicketBroker.js
```diff
@@ -1735,4 +1735,93 @@ contract("TicketBroker", accounts => {
             assert.equal(event.returnValues.reserve.toString(), reserve.toString())
         })
     })
+
+    describe("claimableReserve", () => {
+        it("returns 0 when the reserveHolder does not have a reserve", async () => {
+            const numRecipients = 10
+            await fixture.bondingManager.setMockUint256(functionSig("getTranscoderPoolSize()"), numRecipients)
+            assert.equal((await broker.claimableReserve(constants.NULL_ADDRESS, constants.NULL_ADDRESS)).toString(), "0")
+        })
```
yondonfu: Style nit: Let's add a space between the `it` blocks to remain consistent with the style in the rest of the file (we should add this type of check to the JS linter at some point). Comment holds for the rest of the `it` blocks for the changes in this file as well.

test/unit/TicketBroker.js
```diff
@@ -1735,4 +1735,93 @@ contract("TicketBroker", accounts => {
...
+            const ticket = createWinningTicket(recipient, sender, recipientRand, faceValue)
+            const senderSig = await web3.eth.sign(getTicketHash(ticket), sender)
+            // Claim winning ticket - will freeze reserve (deposit = 0)
+            await broker.redeemWinningTicket(ticket, senderSig, recipientRand, {from: recipient})
+            // Assert reserve is frozen
+            await expectRevertWithReason(
+                broker.fundDeposit({from: sender, value: 1000}),
+                "sender's reserve is frozen"
+            )
+            // claimableReserve should be equal to reserve/numRecepients - faceValue
```
yondonfu: `numRecpients` -> `numRecipients`

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -42,6 +42,24 @@ contract MixinReserve is MContractRegistry, MReserve {
...
+     */
+    function claimableReserve(address _reserveHolder, address _claimant) public view returns (uint256) {
+        ReserveManager storage manager = reserveManagers[_reserveHolder];
+        Reserve storage reserve = manager.reserve;
+        if (reserveState(_reserveHolder) == ReserveState.NotFrozen) {
+            // TODO: Check if claimant is registered in current round
+            return reserve.fundsAdded.div(bondingManager().getTranscoderPoolSize());
```
yondonfu: Just for completeness perhaps we should return 0 if the value returned by `bondingManager().getTranscoderPoolSize()` is 0 (and add a test). Otherwise, this line would revert. While reverting wouldn't be that harmful if we are only calling this function using a `eth_call` JSON-RPC request, I think returning 0 in this case would better match expectations (a claimant cannot claim anything if there are no recipients registered) and would be necessary if we ever call this function within a transaction as well.

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -42,6 +42,24 @@ contract MixinReserve is MContractRegistry, MReserve {
...
+        } else {
+            // TODO: Check if claimant is registered during the freezeRound
+            return reserve.fundsAdded.div(reserve.recipientsInFreezeRound).sub(manager.claimedPerReserve[manager.reserveNonce][_claimant]);
```
yondonfu: Same comment as above about returning 0 if `reserve.recipientsInFreezeRound` is 0.

test/unit/TicketBroker.js
kyriediculous: I'll see if there's a spellchecker for vsCode to mitigate these types of mistakes in the future.

PR #307, opened 2019-09-06
bondingmanager: remove pricePerSegment and transcoderElection

### LINKED ISSUE #254: Remove references/usage of pricePerSegment and transcoder election
We will not be using on-chain price discovery and transcoder election in MVS. Thus, we can remove references/usage of pricePerSegment and transcoder election (i.e. [electActiveTranscoder](https://github.com/livepeer/protocol/blob/pm/contracts/bonding/BondingManager.sol#L649)) from BondingManager.

kyriediculous:
_Note: I created a new branch for the project in which we can merge the incoming PRs. The `expand-active-transcoder-set` branch is branched from the `pm` branch._

This PR removes all `pricePerSegment`, `pendingPricePerSegment` and `electTranscoder()` references and usage from the `BondingManager` contract.

Price discovery and transcoder selection now happens off-chain. 

- Deprecated `pricePerSegment` and `pendingPricePerSegment` variables on the `Transcoder` struct to not change the storage layout of the contract
- Removed `electTranscoder()` method from the `BondingManager` contract and interface
- Removed all state updates pertaining to `pricePerSegment` and `pendingPricePerSegment` in the `transcoder()` method.
- The `transcoder()` method will now throw when the current round is in lock period, previously a transcoder could still change its `pricePerSegment` to higher or equal than the price floor for a round, but since we are removing the usage of `pricePerSegment` such functionality is no longer required. 
- Removed `pricePerSegment` from the `TranscoderUpdate` event  

Adjusted & ran unit tests

Fixes #254 

**Checklist:**

Note I still got these two tests failing, but it might be due to a ganache_cli version difference. 
The blocknumber in the contract equals the blocknumber we get through the web3 method. Would need to be confirmed.
https://github.com/livepeer/protocol/blob/49dad4d6434c2835f3ca9e92f9d1f8789b85d942/test/unit/RoundsManager.js#L196
https://github.com/livepeer/protocol/blob/49dad4d6434c2835f3ca9e92f9d1f8789b85d942/test/unit/RoundsManager.js#L212

### DIFF
--- a/test/unit/BondingManager.js
+++ b/test/unit/BondingManager.js
@@ -1,5 +1,6 @@
 import Fixture from "./helpers/Fixture"
 import expectThrow from "../helpers/expectThrow"
+import expectRevertWithReason from "../helpers/expectFail"
 import {contractId, functionSig, functionEncodedABI} from "../../utils/helpers"
 import {constants} from "../../utils/constants"
 import BN from "bn.js"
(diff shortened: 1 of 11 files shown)


test/unit/BondingManager.js
```diff
@@ -123,46 +123,49 @@ contract("BondingManager", accounts => {
...
             await fixture.roundsManager.setMockBool(functionSig("currentRoundInitialized()"), false)
 
-            await expectThrow(bondingManager.transcoder(5, 10, 1))
+            await expectThrow(bondingManager.transcoder(5, 10))
+        })
+
+        it("should fail if the current round is locked", async () => {
+            await fixture.roundsManager.setMockBool(functionSig("currentRoundLocked()"), true)
+
+            await expectThrow(bondingManager.transcoder(5, 10))
```
yondonfu: Since we added a reason string to the require statement for checking if the current round is locked, can we use the [expectRevertWithReason()](https://github.com/livepeer/protocol/blob/pm/test/helpers/expectFail.js) helper to make sure the tx is failing for the correct reason?

yondonfu:
> Note I still got these two tests failing, but it might be due to a ganache_cli version difference.
The blocknumber in the contract equals the blocknumber we get through the web3 method. Would need to be confirmed.

Running a different version of ganache-cli would cause this. The version of ganache-cli that is specified in `package.json` and used in CI tests will return `block.number - 1` in response to a web3 query for the current block. Versions of ganache-cli >= v6.5.1 will return `block.number` in response to a web3 query for the current block - see the [change log](https://github.com/trufflesuite/ganache-cli/releases/tag/v6.5.1#user-content-v6.5.1-Changelog) for ganache-cli v6.5.1. 

This shouldn't be a problem for this PR since the tests comply with the behavior of the older version of ganache-cli specified in `package.json` but when/if we upgrade the ganache-cli version we'll need to update tests that reference the current block number.

yondonfu:
Noticed that Coveralls is complaining about a decrease in test coverage - would be good to look into whether we can fix this to make sure that we're never decreasing test coverage with a merged PR.

test/unit/BondingManager.js
kyriediculous: fixed in d144143

kyriediculous:
The coveralls seems to stem from not checking the default case for `delegatorStatus` anymore, at least that's the line marked red with an exclamation mark. This must have happened somewhere in the transcoderElection tests then, because that's the only one that got removed.

yondonfu:
> at least that's the line marked red with an exclamation mark. This must have happened somewhere in the transcoderElection tests then, because that's the only one that got removed.

That line was previously uncovered already so that wouldn't contribute to a coverage drop. I actually think the coverage drop is due to the branch coverage calculation:

- Prior this PR we had 331 of 334 branches covered resulting in ~99.1 branch coverage.
- After this PR we have 317 of 320 branches covered resulting in ~99.06% branch coverage.

So, while the absolute # of uncovered branches remained the same, the coverage % decreased slightly since the ratio b/w covered branches & total branches changed.

Let's cover those branches in a separate PR and rebase this branch.

kyriediculous:
Rebased

PR #318, opened 2019-09-13
bondingManager: make DLL in place representation of active set

### LINKED ISSUE #300: Make transcoderPool a fluent representation of the active transcoder set and keep track of a transcoders totalStake for a round
After we remove the `activeTranscoderSet` notion as described in #299 we will use the `transcoderPool` doubly linked list as a fluent representation of the active transcoder set. 

This means that we'll have to update a transcoders membership of the activeSet (remove, insert or change position) when a `bond`, `unbond` or `rebond` action occurs. 

To make sure changes to the active set don't propagate in the current round we can keep track of a transcoder's active state for the next round in its `earningsPoolPerRound` by setting a boolean flag based upon insertion or eviction from the `transcoderPool`. When the stake of a transcoder changes during a round we'll keep track of that amount on the `earningsPoolPerRound` for the next round as well. To achieve this we can use the `EarningsPool.Data.activateStake()` and `EarningsPool.Data.deactivateStake()` methods. 

The downside of this approach is that whenever a delegator delegates to a transcoder so that it becomes part of the `transcoderPool`, the delegator will pay a slightly higher gas cost for the transaction because storage operations to the `transcoderPool` need to occur.  Since the delegator will benefit from the action if the transcoder remains in the active set when the new round begins by receiving rewards, this seems like on ok tradeoff to make but we might want to do a gas cost analysis after the update to make sure the cost isn't out of line. 

**Tasks** 
- Update a transcoders membership of the `transcoderPool` and active status (in earningspool) for the next round when a bonding, unbonding or rebonding action to that transcoder occurs
- Update a transcoders `totalStake` on its earningsPool for the next round when a bonding, unbonding or rebonding action to that transcoder occurs
- Adjust unit tests

### LINKED ISSUE #302: Keep track of the total stake of the activeSet for a round
In the current version of the BondingManager we keep track of the total stake by the active set for a round in the `activeTranscoderSet round=>ActiveTranscoderSet` map's `totalStake` field. 
The aggregation of this stake is made when we call `setActiveTranscoders()` and write the current active set for a round to storage.

We aim to delete this function in #299 so we need a different way to keep track of the totalStake for a round so that we can calculate the correct reward distribution when a transcoder calls `BondingManager.Reward()`. 

We can use two new state variables to track this. 
`currentRoundTotalActiveStake` and `nextRoundTotalActiveStake` 

We can update `nextRoundTotalActiveStake` when the active set for the next round changes based on `transcoderPool` insertions/evictions and then "lock in" that amount by setting `currentRoundTotalActiveStake` equal to `nextRoundTotalActiveStake` upon initializing a new round. 

There is a good reason for not choosing a map of `round=>totalStake` here, because that would require setting zeroed storage slots to non-zero, which incurs a higher gas cost (20 000 extra) compared to setting a non-zeroed storage slot to another non-zeroed value.

### LINKED ISSUE #304: Repurpose transcoder registered status to reflect self-delegation as desire to be an transcoder
Since we're removing the notion of an `activeTranscoderSet` will become equal to simply being in the active set or not, as that check is currently whether the transcoder is in `transcoderPool`. 

A transcoder might still however be registered but not part of the active set for a round.  Therefore we need a different way to determine if a transcoder is registered if we are to use `transcoderPool` as the active transcoder set. 

We can treat self-delegation as desire to be a transcoder. This means Registered/Unregistred will be implicit based on a transcoder's own stake but in the current version of the protocol this is also implicit by being part of the `transcoderPool`. 

An alternative would be to add a flag to the `Transcoder` struct indicating registered status but that approach is less favorable.
- It changes the storage layout (even though the changes can be compatible, it seems reasonable to avoid doing it unless necessary). 
- It incurs additional SSTORE operations: registering would cost 20k gas extra (setting a zero storage slot to non-zero) and resigning would cost 5k gas more (setting a non-zero slot to zero)

kyriediculous:
- Make the sorted doubly linked list a fluent representation of the active transcoder set as stake updates happen
- Repurpose transcoder `Registered` status to indicate self-bonding as a willingness to be a transcoder
- Keep track of total stake for the active transcoder set

- Removed `setNumTranscoders()` function 
- Added helpers to increase/decrease stake for active transcoders as well as process total active set stake updates
- Added a helper to join the active set based on current transcoder stake
- Reworked resign method to do the necessary state updates for active transcoders
- Reworked `isRegisteredTranscoder` to indicate self-bonding
- Removed `transcoderStatus()` method, instead we now have `isActiveTranscoder()` and `isRegisteredTranscoder()` that return boolean values
- Deprecated `numActiveTranscoders` variable, instead we can get the size/maxSize of transcoderPool with the public helpers
- added a call to `bondingManager.setCurrentRoundTotalActiveStake()` upon round init to set `nextRoundTotalActiveStake` to the `currentRoundTotalActiveStake` 
- adjusted unit & integration tests
- added a npm run script `npm run test:describe` to run tests for describe blocks (uses mocha's grep feature with `-g` flag)
- added a map `lastActiveStakeUpdateRound` that keeps track of the round in which an active transcoder's stake was last updated

Fixes #300 
Fixes #302 
Fixes #304 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -25,7 +25,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     // Time between unbonding and possible withdrawl in rounds
     uint64 public unbondingPeriod;
     // Number of active transcoders
-    uint256 public numActiveTranscoders;
+    uint256 public numActiveTranscodersDEPRECATED;
     // Max number of rounds that a caller can claim earnings for at once
     uint256 public maxEarningsClaimsRounds;
 
@@ -88,6 +88,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     // Keep track of active transcoder set for each round
     mapping (uint256 => ActiveTranscoderSet) public activeTranscoderSet;
 
+    // The total active stake (sum of the stake of active set members) for the current round
+    uint256 public currentRoundTotalActiveStake;
+    // The total active stake (sum of the stake of active set members) for the next round
+    uint256 public nextRoundTotalActiveStake;
+
     // Check if sender is TicketBroker
     modifier onlyTicketBroker() {
         require(msg.sender == controller.getContract(keccak256("TicketBroker")));
@@ -129,27 +134,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     }
 
     /**
-     * @dev Set max number of registered transcoders. Only callable by Controller owner
-     * @param _numTranscoders Max number of registered transcoders
-     */
-    function setNumTranscoders(uint256 _numTranscoders) external onlyControllerOwner {
-        // Max number of transcoders must be greater than or equal to number of active transcoders
-        require(_numTranscoders >= numActiveTranscoders);
-
-        transcoderPool.setMaxSize(_numTranscoders);
-
-        emit ParameterUpdate("numTranscoders");
-    }
-
-    /**
      * @dev Set number of active transcoders. Only callable by Controller owner
      * @param _numActiveTranscoders Number of active transcoders
      */
     function setNumActiveTranscoders(uint256 _numActiveTranscoders) external onlyControllerOwner {
-        // Number of active transcoders cannot exceed max number of transcoders
-        require(_numActiveTranscoders <= transcoderPool.getMaxSize());
-
-        numActiveTranscoders = _numActiveTranscoders;
+        transcoderPool.setMaxSize(_numActiveTranscoders);
 
         emit ParameterUpdate("numActiveTranscoders");
     }
@@ -177,6 +166,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         Transcoder storage t = transcoders[msg.sender];
         Delegator storage del = delegators[msg.sender];
 
+        uint256 currentRound = roundsManager().currentRound();
+
         // No transcoder operations can take place during the round lock period
         require(
             !roundsManager().currentRoundLocked(),
@@ -194,26 +185,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         t.pendingRewardCut = _rewardCut;
         t.pendingFeeShare = _feeShare;
 
-        uint256 delegatedAmount = del.delegatedAmount;
-
-        // Check if transcoder is not already registered
-        if (transcoderStatus(msg.sender) == TranscoderStatus.NotRegistered) {
-            if (!transcoderPool.isFull()) {
-                // If pool is not full add new transcoder
-                transcoderPool.insert(msg.sender, delegatedAmount, address(0), address(0));
-            } else {
-                address lastTranscoder = transcoderPool.getLast();
-
-                if (delegatedAmount > transcoderTotalStake(lastTranscoder)) {
-                    // If pool is full and caller has more delegated stake than the transcoder in the pool with the least delegated stake:
-                    // - Evict transcoder in pool with least delegated stake
-                    // - Add caller to pool
-                    transcoderPool.remove(lastTranscoder);
-                    transcoderPool.insert(msg.sender, delegatedAmount, address(0), address(0));
-
-                    emit TranscoderEvicted(lastTranscoder);
-                }
-            }
+        if (!transcoderPool.contains(msg.sender)) {
+            tryToJoinActiveSet(msg.sender, del.delegatedAmount, currentRound);
         }
 
         emit TranscoderUpdate(msg.sender, _rewardCut, _feeShare, transcoderPool.contains(msg.sender));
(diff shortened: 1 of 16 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -884,21 +835,78 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            transcoderPool.updateKey(_transcoder, newStake, address(0), address(0));
+            nextRoundTotalActiveStake = nextRoundTotalActiveStake.add(_amount);
+            lastActiveStakeUpdateRound[_transcoder] = nextRound;
+        } else {
+            // Check if the transcoder is eligible to join the active set in the update round
+            tryToJoinActiveSet(_transcoder, newStake, currentRound);
+        }
+    }
+
+    function decreaseTranscoderTotalStake(address _transcoder, uint256 _amount) internal {
```
yondonfu: Let's add a function comment here.

contracts/bonding/libraries/EarningsPool.sol
```diff
@@ -55,6 +55,16 @@ library EarningsPool {
         return earningsPool.claimableStake > 0;
     }
 
+    function activateStake(EarningsPool.Data storage earningsPool, uint256 _stake) internal {
```
yondonfu: Let's add a function comment here.

contracts/bonding/libraries/EarningsPool.sol
```diff
@@ -55,6 +55,16 @@ library EarningsPool {
...
+        earningsPool.totalStake = _stake;
+        earningsPool.claimableStake = _stake;
+    }
+
+    function deactivateStake(EarningsPool.Data storage earningsPool) internal {
```
yondonfu: Let's add a function comment here.

contracts/rounds/RoundsManager.sol
```diff
@@ -93,6 +93,8 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
         _blockHashForRound[currRound] = roundBlockHash;
         // Set active transcoders for the round
         bondingManager().setActiveTranscoders();
+        // set total active stake for the round
```
yondonfu: "set" -> "Set"

contracts/bonding/BondingManager.sol
```diff
@@ -884,21 +835,78 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         return delegators[_delegator].unbondingLocks[_unbondingLockId].withdrawRound > 0;
     }
 
+    function increaseTranscoderTotalStake(address _transcoder, uint256 _amount) internal {
```
yondonfu: Currently, whenever we increase a delegator's `delegatedAmount` field, we also check if the delegate is a registered transcoder and if so we call `increaseTranscoderTotalStake()`. But, given how `increaseTranscoderTotalStake()` is implemented we need to make sure that `delegatedAmount` is increased *after* `increaseTranscoderTotalStake()` or else we will count `_amount` twice when updating active stake.

I see a few things that are not problems right now, but we should be careful about:

- Making sure to always increase `delegatedAmount` *after* calling `increaseTranscoderTotalStake()`
- Making sure to always update `delegatedAmount` when we call `increaseTranscoderTotalStake()` and vice-versa

One way that we can reduce the chance of error is to:

- Change the name of `increaseTranscoderTotalStake()` to `increaseTotalStake()` (I'm going simple with the naming)
- Move the logic for increasing the `delegatedAmount` field on a `Delegator` struct into `increaseTotalStake()`
- Move the logic for checking if an address is a registered transcoder into this function and execute the current function body only if the address is a registered transcoder

WDYT?

contracts/bonding/BondingManager.sol
yondonfu: Currently, whenever we decrease a delegator's `delegatedAmount` field, we also check if the delegate is a registered transcoder and if so we call `decreaseTranscoderTotalStake()`. But, given how `decreaseTranscoderTotalStake()` is implemented we need to make sure that `delegatedAmount` is decreased *after* `increaseTranscoderTotalStake()` or else we will count `_amount` twice when updating active stake.

I see a few things that are not problems right now, but we should be careful about:

- Making sure to always decrease `delegatedAmount` *after* calling `decreaseTranscoderTotalStake()`
- Making sure to always update `delegatedAmount` when we call `decreaseTranscoderTotalStake()` and vice-versa

One way that we can reduce the chance of error is to:

- Change the name of `decreaseTranscoderTotalStake()` to `decreaseTotalStake()` (I'm going simple with the naming)
- Move the logic for increasing the `delegatedAmount` field on a `Delegator` struct into `decreaseTotalStake()`
- Move the logic for checking if an address is a registered transcoder into this function and execute the current function body only if the address is a registered transcoder

WDYT?

contracts/bonding/BondingManager.sol
```diff
@@ -982,13 +990,14 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Increase delegate's delegated amount
         delegators[del.delegateAddress].delegatedAmount = delegators[del.delegateAddress].delegatedAmount.add(amount);
 
-        if (transcoderStatus(del.delegateAddress) == TranscoderStatus.Registered) {
+        // Delete lock
+        delete del.unbondingLocks[_unbondingLockId];
```
yondonfu: Why did this statement move?

contracts/bonding/BondingManager.sol
```diff
@@ -884,21 +835,78 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-            // Decrease total active stake for the round
-            activeTranscoderSet[currentRound].totalStake = activeTranscoderSet[currentRound].totalStake.sub(activeTranscoderTotalStake(_transcoder, currentRound));
-            // Set transcoder as inactive
-            activeTranscoderSet[currentRound].isActive[_transcoder] = false;
-        }
-
-        // Remove transcoder from pools
-        transcoderPool.remove(_transcoder);
-
+        transcoders[_transcoder].earningsPoolPerRound[currentRound].deactivateStake();
```
yondonfu: These statements should be wrapped in a if statement that checks `transcoderPool.contains(_transcoder)`.

Within the if statement, we should also add:

- `transcoderPool.remove(_transcoder)` to actually remove the address from the active set
- Set `lastActiveStakeUpdateRound = 0`

contracts/bonding/BondingManager.sol
```diff
@@ -884,21 +835,78 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        lastActiveStakeUpdateRound[_transcoder] = nextRound;
+        nextRoundTotalActiveStake = pendingNextRoundTotalActiveStake;
+    }
+
     /**
      * @dev Remove transcoder
      */
     function resignTranscoder(address _transcoder) internal {
+        uint256 transcoderStake = transcoderTotalStake(_transcoder);
+        decreaseTranscoderTotalStake(_transcoder, transcoderStake);
```
yondonfu: We don't need this call here since the other statements in this function will remove the address from the active set if needed.

contracts/bonding/BondingManager.sol
```diff
@@ -248,44 +224,46 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        require(delegationAmount > 0);
+        require(delegationAmount > 0, "delegation amount must be greater than 0");
         // Update delegate
         del.delegateAddress = _to;
-        // Update current delegate's delegated amount with delegation amount
-        delegators[_to].delegatedAmount = delegators[_to].delegatedAmount.add(delegationAmount);
-
-        if (transcoderStatus(_to) == TranscoderStatus.Registered) {
+        // Update bonded amount
+        del.bondedAmount = del.bondedAmount.add(_amount);
```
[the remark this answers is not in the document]
kyriediculous: Otherwise if we would bond from being unbonded we would always count as inactive and never try to join the active set because we only update the bonded amount after calling that statement.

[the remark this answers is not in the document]
kyriediculous: Otherwise if we would bond from being unbonded we would always count as inactive and never try to join the active set because we only update the bonded amount after calling that statement.

contracts/bonding/libraries/EarningsPool.sol
kyriediculous: This function is going away in the next PR since we are no longer keeping track of active state on the earningspool.

instead we'll be using `setStake` and `setCommissions`

contracts/bonding/libraries/EarningsPool.sol
kyriediculous: This function is going away in the next PR since we are no longer keeping track of active state on the earningspool.

instead we'll be using `setStake` and `setCommissions`

contracts/bonding/BondingManager.sol
kyriediculous: No particular reason, I just though it would be neater to do the stake update after delegator state changes

contracts/bonding/BondingManager.sol
kyriediculous: I agree that the current flow isn't optimal and could lead to dev errors. 

I contemplated adding increasing delegatedAmount into those methods but didn't add it since all delegator state updates were happening outside of the scope of those increase/decrease helpers.

I'm not against adding it there however, on the contrary. So the feedback is much appreciated.

kyriediculous:
Added a commit that includes the requested changes as well as a new flag on the `Transcoder` struct, `activationRound`. 

We can't do an _active transcoder_ check on `Transcoder.lastActiveStakeUpdateRound >0 && <= currentRound`, because eg. when a delegate bonds to an already registered transcoder then the transcoder's `lastActiveStakeUpdateRound` will be set to `currentRound + 1`, resulting in `isActiveTranscoder` returning `false` when it should be `true`.

yondonfu:
Since #323 is merged and these changes were included in that PR can we close this?

kyriediculous:
Changed included in #323

PR #319, opened 2019-09-13
Remove freeze period

### LINKED ISSUE #306: Consider removing explicit freeze period in TicketBroker
At the moment, the TicketBroker contract explicitly defines a `freezePeriod` parameter. The value of this parameter is the # of rounds that a sender's reserve remains frozen after the sender overspends from its deposit and a recipient claims from the reserve. 

There are a few downsides with the way the explicit multi-round freeze period is currently implemented:

- More state tracking costs. The contract stores the freeze round, the # of recipients in the freeze round and a "reserve nonce" that is used to incremented whenever a reserve is funded after it was previously frozen (the nonce is used to point to a new mapping that tracks amounts claimed by recipients when a reserve is frozen). The main transactions that these costs are passed on to are reserve claiming ticket redemptions and funding a reserve after it was previously frozen
- More client complexity. Recipients might need to handle frozen vs. non-frozen reserves differently. For example, with an explicit multi-round freeze period, if a recipient encounters a frozen reserve it would need to check if it was in the active set in the freeze round as opposed to just checking if it is in the active set in the current round. Furthermore, recipients need to make a decision about whether to work with a sender with a frozen reserve. If the decision is to work with a sender with a frozen reserve then the client needs to also make sure that the reserve is not thawed (check if current round >= the reserve's thaw round)
- Does not allow a sender top off its reserve immediately after the reserve is frozen - the sender needs to wait through the multi-round freeze period first. This UX is a bit rough and treats overspending fairly harshly when in reality it could happen by accident.

An alternative to consider is to remove the explicit freeze period and instead if overspending happens the reserve only stays committed to current round's recipient set. Essentially, the only thing different about the overspending scenario is that funds come from the reserve instead of the deposit and there is a cap on the claimable amount for each recipient. The main downside of this alternative is that if a sender overspends at the very end of a round and a recipient is evicted from the active set as the new round begins, then the recipient will not be able to redeem any received winning tickets to claim from the sender's reserve. The explicit multi-round freeze period helps with this because if someone freezes the sender's reserve at the end of the round, then all recipients are still guaranteed the ability to redeem received winning tickets to claim from the sender's reserve in the following round. However, the benefit of the freeze period here depends on someone actually freezing the sender's reserve before the round ends - if no one freezes the reserve in time then any evicted recipients in the next round would still be out of luck. It is also worth noting that being evicted from the active set in a following round is a general problem even without overspending - if a recipient is evicted in the next round then it will be unable to redeem any received winning tickets.

I think this alternative has a few benefits:

- Less state tracking costs. The contract does not need to store any state pertaining to the freeze round or # of recipients in the freeze round - if the reserve is only frozen for the remainder of the round, then the contract can just rely on contract queries that will be accurate for the round. The "reserve nonce" does not need to be stored either - instead we can use the current round number as a pointer to new mappings that track amounts claimed from a reserve in each round
- Less client complexity. There would be no special casing required for frozen vs. non-frozen reserves. Recipients would always just check if they are active in the current round and the size of the active set in the current round. Recipients would not need to decide whether to work with senders with frozen vs. non-frozen reserves - recipients can work with these senders as long as the allocation from the reserve is sufficient
- Allow a sender to top off its reserve immediately after it is frozen. This treats overspending less harshly - if it happens a sender can quickly top off its funds and continue paying recipients

At a high level, the implementation change for this alternative would:

- Use round numbers to point at mappings for claimed amounts by recipients
- If overspending happens, record the claimable amount at the time of overspending for the round so that value can be used to calculate reserve allocations
- Use a single state variable to track remaining funds in a reserve
- If funds are added to a reserve during a round when overspending happened, update the funds in the reserve, but also the claimable amount for the reserve (to allow recipients in the round to access those new funds if they claim from the reserve)

Most of these changes are outlined in [this spike](https://github.com/livepeer/protocol/compare/pm...yf/freeze-reserve-remainder-of-round-spike). Note: the spike branch still uses the frozen vs. non-frozen terminology for now.

yondonfu:
This PR removes the freeze period from the TicketBroker contract. Instead of freezing the reserve for multiple rounds after the deposit is depleted, recipients in the active set can claim from the reserve up to their max allocation for the remainder of the round. At the beginning of the next round, the max allocation for each recipient in the active set in the round is calculated based on the reserve funds available at the beginning of the round.

Fixes #306 

See the below Discord conversation for additional context:

```
j0sh: The freeze period removal makes sense to me. Just want to clarify my own understanding on one point:

The original concern I had (the reason for re-reviewing things) was due to the fact that tickets could be valid for longer than a round, so if the reserve is "reset" after each round, will the O be assured that it still has its original float available, especially if someone submits a ticket that draws from the reserve at the very end of a round? (The O is still in the pool for the next round.)

There is this item: "record the claimable amount at the time of overspending " - does that mean, if the reserve is drawn, something will be written on-chain which says, "At round X, orchestrators were entitled to $Y" ? How does that work?
```

```
yondon: Ah so this is an additional edge case introduced by the proposed design that "resets" the reserve at the beginning of each round. At the beginning of each round, each O in the active set is allocated X /  Y where X is the reserve amount (before anyone has claimed from it during the round) and Y is the size of the active set. 

If someone claims from B's reserve at the end of round N, then B's reserve will be X' < X. When round N + 1 starts, O's float will be less than X / Y since B's reserve at the beginning of round N + 1 is less than X.

Another situation where O's float can be reduced when a new round begins is if the size of the active set increases. If the size of the active set is Y + 1 in round N + 1, then O's float in round N + 1 is reduced to X / (Y + 1).

In summary, the edge cases that can happen if O receives a winning ticket at the end of a round are:

- O is no longer in the active set in round N + 1 and thus cannot redeem its ticket

- The size of the active set increases to Y + 1 in round N + 1 and thus O might not receive the full ticket faceValue when redeeming if the ticket faceValue > X / (Y + 1)

- Someone claims from the reserve at the end of round N which reduces the reserve to X'. O might not receive the full ticket faceValue when redeeming in round N + 1 if the ticket faceValue > X'
The first 2 edge cases would remain even if a freeze period is used. The 3rd edge case would be removed if a freeze period is used

All of these edge cases are caused by a combination of 2 factors:

1) O receiving a winning ticket very close to the end of a round

2) Some other factor that is out of O's control (i.e. its membership in the active set in the next round, the size of the active set increasing in the next round, someone claiming from the reserve before the current round ends)

It is worth noting that 1) is somewhat controlled by O since O sets the winning probability of tickets - given that tickets should generally have a low winning probability in order to avoid frequent on-chain txs, the probability of receiving a winning ticket close to the end of a round should also be low.

2) is out of O's control, but one strategy that O could use is that if it does receive a winning ticket close to the end of a round, it can pay a slightly higher gas fee to include the tx on-chain faster. This would cut into O's profits a bit though so O might want to use heuristics to determine how likely it is that one of the factors for 2) will occur. For example, O could check its stake ranking in the active set and only pay a slightly higher gas fee if it has a very low ranking with a higher risk of being evicted from the set next round if someone with more stake comes along.
There is this item: "record the claimable amount at the time of overspending " - does that mean, if the reserve is drawn, something will be written on-chain which says, "At round X, orchestrators were entitled to $Y" ? How does that work?

Ah yeah that sentence could've been clearer...let me try to reword...

If overspending happens, the contract will record the reserve amount prior to the first claim that draws from the reserve. The reserve amount recorded is the "claimable amount". During the round, this amount can be divided by the size of the active set to calculate the allocation guaranteed to each member of the active set.

The reason why the contract records the reserve amount prior to the first claim that draws from the reserve when overspending happens is because we want to calculate the allocations based on the total reserve available before it started to decrease due to claims.

Example:

1. reserve = 100 activeSetSize = 10
2. Person A claims 5 so we record claimableAmount = 100
3. reserve = 95
4. Person B claims - we calculate the allocation as 100 / 10 (instead of 95 / 10). So, the max amount that can be claimed is 10
5. Person A claims - we calculate the allocation as (100 / 10) - 5. So, the max amount that can be claimed is 5. We subtract 5 since person A already claimed 5 during that round

However, I ended up moving to a different design mentioned in this comment: https://github.com/livepeer/protocol/issues/306#issuecomment-530104035

Instead of recording the reserve amount prior to the first claim that draws from the reserve when overspending happens, the contract just records the total claimed from the reserve in the round thus far. IMO, this is simpler to reason about.

Example:

1. reserve = 100 activeSetSize = 10 claimed = 0
2. Person A claims 5 so we record claimed = 5
3. reserve = 95 claimed = 5
4. Person B claims - we calculate the allocation as (95 + 5) / 10. So, the max amount that can be claimed is 10
5. Person A claims - we calculate the allocation as (95 + 5) / 10 - 5. So, the max amount that can be claimed is 5.

In a round, we calculate the total reserve available to members of the active set by adding the funds remaining in the reserve with the funds already claimed from the reserve for that round
```

### DIFF
--- a/contracts/pm/mixins/MixinReserve.sol
+++ b/contracts/pm/mixins/MixinReserve.sol
@@ -84,6 +84,19 @@ contract MixinReserve is MContractRegistry, MReserve {
      * @param _reserveHolder Address of reserve holder
      */
     function clearReserve(address _reserveHolder) internal {
+        // This delete operation will only clear reserve.funds and will not clear the storage for reserve.claimedForRound
+        // reserve.claimedByAddress because these fields are mappings and the Solidity `delete` keyword will not modify mappings.
+        // This *could* be a problem in the following scenario:
+        //
+        // 1) In round N, for address A, reserve.claimedForRound[N] > 0 and reserve.claimedByAddress[N][r_i] > 0 where r_i is
+        // a member of the active set in round N
+        // 2) This function is called by MixinTicketBrokerCore.withdraw() in round N
+        // 3) Address A funds its reserve again
+        //
+        // After step 3, A has reserve.funds > 0, reserve.claimedForRound[N] > 0 and reserve.claimedByAddress[N][r_i] > 0
+        // despite having funded a fresh reserve after previously withdrawing all of its funds in the same round.
+        // We prevent this scenario by disallowing reserve claims starting at an address' withdraw round in
+        // MixinTicketBrokerCore.redeemWinningTicket()
         delete reserves[_reserveHolder];
     }
 
--- a/contracts/pm/mixins/MixinReserve.sol
+++ b/contracts/pm/mixins/MixinReserve.sol
@@ -11,23 +11,13 @@ contract MixinReserve is MContractRegistry, MReserve {
     using SafeMath for uint256;
 
     struct Reserve {
-        uint256 fundsAdded;               // Amount of funds added to the reserve
-        uint256 fundsClaimed;             // Amount of funds claimed from the reserve
-        uint256 freezeRound;              // Round that the reserve was frozen
-        uint256 recipientsInFreezeRound;  // Number of recipients registered in BondingManager during freezeRound
+        uint256 funds;                                                        // Amount of funds in the reserve
+        mapping (uint256 => uint256) claimedForRound;                         // Mapping of round => total amount claimed
+        mapping (uint256 => mapping (address => uint256)) claimedByAddress;   // Mapping of round => claimant address => amount claimed
     }
 
-    struct ReserveManager {
-        uint256 reserveNonce;                                                 // Current reserve ID
-        Reserve reserve;                                                      // Storage pointer to a reserve
-        mapping (uint256 => mapping (address => uint256)) claimedPerReserve;  // Mapping of reserve ID => claimant address => amount claimed for reserve ID
-    }
-
-    // Mapping of address => managed reserve for an address
-    mapping (address => ReserveManager) internal reserveManagers;
-
-    // Number of rounds before a frozen reserve thaws
-    uint256 public freezePeriod;
+    // Mapping of address => reserve
+    mapping (address => Reserve) internal reserves;
 
     /**
      * @dev Returns info about a reserve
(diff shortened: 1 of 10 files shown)


contracts/pm/mixins/MixinReserve.sol
```diff
@@ -11,67 +11,61 @@ contract MixinReserve is MContractRegistry, MReserve {
...
-        } else {
-            // TODO: Check if claimant is registered during the freezeRound
-            return reserve.recipientsInFreezeRound > 0 ?
-                reserve.fundsAdded.div(reserve.recipientsInFreezeRound).sub(manager.claimedPerReserve[manager.reserveNonce][_claimant]) :
-                0;
+        Reserve storage reserve = reserves[_reserveHolder];
+
+        uint256 currentRound = roundsManager().currentRound();
+
+        // TODO: Check if claimant is active in the current round
```
kyriediculous: I assume we aren't implementing the active check yet because the function signature changes in #318 ?

yondonfu: Yep I wanted to wait for the BondingManager changes to land first.

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -80,27 +74,8 @@ contract MixinReserve is MContractRegistry, MReserve {
...
-            // value and then setting its new value
-            reserve.fundsClaimed = 0;
-            reserve.freezeRound = 0;
-            reserve.recipientsInFreezeRound = 0;
-            // Increment reserveNonce so the manager can point to a new
-            // mapping for tracking funds claimed from a reserve by
-            // different claimants
-            manager.reserveNonce = manager.reserveNonce.add(1);
-        }
+        Reserve storage reserve = reserves[_reserveHolder];
```
kyriediculous: we could access the field directly since we're no longer accessing multiple fields on the `Reserve` struct within this function. I guess it saves us some stack operations of popping the storage address of the struct onto the stack.

`reserves[_reserveHolder].funds = reserves[_reserveHolder].funds.add(_amount);`

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -133,92 +103,35 @@ contract MixinReserve is MContractRegistry, MReserve {
...
-            reserve.recipientsInFreezeRound = recipientsInFreezeRound;
-
-            emit ReserveFrozen(
-                _reserveHolder,
-                _claimant,
-                freezeRound,
-                recipientsInFreezeRound
-            );
-        }
+        Reserve storage reserve = reserves[_reserveHolder];
```
kyriediculous: I think we can move this inside of the `if claimAmount > 0` block

contracts/pm/mixins/MixinTicketBrokerCore.sol
```diff
@@ -211,7 +198,7 @@ contract MixinTicketBrokerCore is MReserve, MTicketProcessor, MTicketBrokerCore
     /**
      * @dev Withdraws all ETH from the caller's deposit and reserve
      */
-    function withdraw() public reserveNotFrozen(msg.sender) {
+    function withdraw() public {
         Sender storage sender = senders[msg.sender];
 
         uint256 deposit = sender.deposit;
```
kyriediculous: Even though not part of this PR, we don't need this additional MSTORE operation, we can just SLOAD the deposit directly using `sender.deposit` on L208

contracts/pm/mixins/MixinReserve.sol
yondonfu: Fixed in 1723500

contracts/pm/mixins/MixinReserve.sol
yondonfu: Fixed in 1723500

kyriediculous:
Setting the transcoderpoolSize explicitly to 0 instead of implicitly seems like a good idea!

yondonfu:
> Setting the transcoderpoolSize explicitly to 0 instead of implicitly seems like a good idea!

Hm I actually pushed 3c68c6c to try to bring test coverage back up since one of the lines in `claimableReserve()` is uncovered. Need to investigate why that line isn't being hit...

test/unit/TicketBroker.js
```diff
@@ -1737,19 +1524,28 @@ contract("TicketBroker", accounts => {
...
-        it("Returns 0 when the active transcoder pool size is 0 if reserve is not frozen", async () => {
+        it("returns 0 if claimant is not active in the current round", async () => {
             const reserve = 1000
             await broker.fundReserve({from: sender, value: reserve})
             assert.equal((await broker.claimableReserve(sender, constants.NULL_ADDRESS)).toString(10), "0")
         })
 
-        it("Returns claimable reserve for a claimaint if reserve is not frozen", async () => {
+        it("returns 0 when the active transcoder pool size is 0", async () => {
+            await fixture.bondingManager.setMockBool(functionSig("isRegisteredTranscoder()"), true)
```
kyriediculous: for the test-coverage , should this perhaps be 
`"isRegisteredTranscoder(address)"` to match the real function signature?

yondonfu: Aha! Great catch. a43a2b4 should do the trick

kyriediculous: Yep it should, I ran coveralls locally already. 

Let's rebase ? 😄

yondonfu:
Rebased. Coverage is still slightly down, but that should be ok for this PR because the drop is due to a ratio change - the # of missing lines/branches (3) remained constant and all additional lines/branches are covered.

PR #320, opened 2019-09-13
pm: Change unlock period to be round based

yondonfu:
This PR updates the unlock period to be round based which makes the unlock period consistent with other time based variables in the rest of the contracts. Furthermore, the addition of a `withdrawRound` for the sender allows us to prevent reserve claims after the sender's unlock period is over which avoids an edge case with a reserve's `claimedForRound` and `claimedByAddress` mappings not being cleared upon withdrawal round N and being non-zero after a funding operation in that same round.


yondonfu:
Rebased with #319

PR #324, opened 2019-09-18
pm: add whenSystemNotPaused and currentRoundInitialized modifiers

### LINKED ISSUE #308: Add whenSystemNotPaused and currentRoundInitialized modifiers in TicketBroker
External/public state mutating functions in the TicketBroker contract should be protected by the `whenSystemNotPaused` (to prevent state updates if the entire system is paused) modifier and `redeemWinningTicket()` should be protected by the `currentRoundInitialized` modifier (since ticket validity is round based so we want to make sure the current round is initialized).

kyriediculous:
This PR adds a modifier to external/public `TicketBroker` functions and disallows ticket redemption if the round is not initialized.

- Moved all external/public state mutating functions in the `pm` package to `TicketBroker.sol` so we can use `Manager.whenSystemNotPaused`. 
- Added `currentRoundInitialized` modifier to `TicketBroker`
- Add `whenSystemNotPaused` modifier to all external/public functions
- Add `currentRountInitialized` modifier to `redeemWinningTicket` and `batchRedeemWinningTickets`
- Added a revert message for `Manager.whenSystemNotPaused`

Fixes #308 

**Checklist:**

### DIFF
--- a/contracts/pm/mixins/MixinContractRegistry.sol
+++ b/contracts/pm/mixins/MixinContractRegistry.sol
@@ -4,7 +4,15 @@ import "../../ManagerProxyTarget.sol";
 import "./interfaces/MContractRegistry.sol";
 
 
-contract MixinContractRegistry is ManagerProxyTarget, MContractRegistry {
+contract MixinContractRegistry is MContractRegistry, ManagerProxyTarget {
+    /**
+     * @dev Checks if the current round has been initialized
+     */
+    modifier currentRoundInitialized() {
+        require(roundsManager().currentRoundInitialized(), "current round is not initialized");
+        _;
+    }
+
     constructor(address _controller)
         internal
         Manager(_controller)
--- a/contracts/pm/mixins/MixinTicketBrokerCore.sol
+++ b/contracts/pm/mixins/MixinTicketBrokerCore.sol
@@ -67,6 +67,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     function fundDeposit()
         external
         payable
+        whenSystemNotPaused
         processDeposit(msg.sender, msg.value)
     {
         processFunding(msg.value);
@@ -78,6 +79,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     function fundReserve()
         external
         payable
+        whenSystemNotPaused
         processReserve(msg.sender, msg.value)
     {
         processFunding(msg.value);
@@ -94,6 +96,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     )
         external
         payable
+        whenSystemNotPaused
         checkDepositReserveETHValueSplit(_depositAmount, _reserveAmount)
         processDeposit(msg.sender, _depositAmount)
         processReserve(msg.sender, _reserveAmount)
@@ -114,6 +117,8 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
         uint256 _recipientRand
     )
         public
+        whenSystemNotPaused
+        currentRoundInitialized
     {
         bytes32 ticketHash = getTicketHash(_ticket);
 
@@ -178,7 +183,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     /**
      * @dev Initiates the unlock period for the caller
      */
-    function unlock() public {
+    function unlock() public whenSystemNotPaused {
         Sender storage sender = senders[msg.sender];
 
         require(
@@ -196,7 +201,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     /**
      * @dev Cancels the unlock period for the caller
      */
-    function cancelUnlock() public {
+    function cancelUnlock() public whenSystemNotPaused {
         Sender storage sender = senders[msg.sender];
 
         _cancelUnlock(sender, msg.sender);
@@ -205,7 +210,7 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
     /**
      * @dev Withdraws all ETH from the caller's deposit and reserve
      */
-    function withdraw() public {
+    function withdraw() public whenSystemNotPaused {
         Sender storage sender = senders[msg.sender];
 
         uint256 deposit = sender.deposit;
--- a/contracts/pm/mixins/MixinWrappers.sol
+++ b/contracts/pm/mixins/MixinWrappers.sol
@@ -3,9 +3,11 @@ pragma solidity ^0.4.25;
 pragma experimental ABIEncoderV2;
 
 import "./interfaces/MTicketBrokerCore.sol";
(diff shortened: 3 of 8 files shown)


kyriediculous:
fixed some linter issues in 32311ed

contracts/pm/TicketBroker.sol
```diff
@@ -15,7 +15,13 @@ contract TicketBroker is
     MixinTicketBrokerCore,
     MixinTicketProcessor,
     MixinWrappers
-{
+{ 
+    // Check if current round is initialized
+    modifier currentRoundInitialized() {
```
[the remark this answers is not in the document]
kyriediculous: fixed up in 11ecbe1

contracts/pm/TicketBroker.sol
```diff
@@ -15,7 +15,7 @@ contract TicketBroker is
     MixinTicketBrokerCore,
     MixinTicketProcessor,
     MixinWrappers
-{
+{
```
yondonfu: Let's kill this extra whitespace.

contracts/pm/mixins/MixinContractRegistry.sol
```diff
@@ -4,7 +4,15 @@ import "../../ManagerProxyTarget.sol";
 import "./interfaces/MContractRegistry.sol";
 
 
-contract MixinContractRegistry is ManagerProxyTarget, MContractRegistry {
+contract MixinContractRegistry is MContractRegistry, ManagerProxyTarget {
```
yondonfu: Ah so I'm guessing that the inheritance order is switched because we need to override the `whenSystemNotPaused()` implementation in `MContractRegistry` with the implementation in `ManagerProxyTarget`. Let's add a `@dev` for this contract (i.e. in a NatSpec comment right before this line) that mentions the importance of inheriting `MContractRegistry` *before* `ManagerProxyTarget`.

contracts/pm/mixins/MixinTicketBrokerCore.sol
```diff
@@ -61,12 +61,13 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
         _;
     }
 
-    /**
+     /**
```
yondonfu: Extra whitespace here?

contracts/pm/mixins/MixinTicketBrokerCore.sol
```diff
@@ -114,6 +117,8 @@ contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor,
         uint256 _recipientRand
     )
         public
+        currentRoundInitialized
```
yondonfu: Let's run the `whenSystemNotPaused()` check first - this is the ordering we use in the BondingManager and also skips the round initialized check when the system is actually paused.

contracts/pm/mixins/MixinWrappers.sol
```diff
@@ -19,6 +21,8 @@ contract MixinWrappers is MTicketBrokerCore {
         uint256[] _recipientRands
     )
         public
+        currentRoundInitialized
```
yondonfu: Let's run the `whenSystemNotPaused()` check first - this is the ordering we use in the BondingManager and also skips the round initialized check when the system is actually paused.

contracts/pm/mixins/interfaces/MContractRegistry.sol
```diff
@@ -6,6 +6,22 @@ import "../../../rounds/IRoundsManager.sol";
 
 
 contract MContractRegistry {
+    /**
+     * @dev Checks if the system is paused
+     * @notice Executes the 'whenSystemNotPaused' modifier 'MixinContractRegistry' inherits from 'Manager.sol'
```
yondonfu: I think the notice here should actually be a part of the `@dev` tag since it is developer documentation. Basing this off of the [NatSpec tag docs](https://solidity.readthedocs.io/en/v0.5.11/natspec-format.html#tags).

contracts/pm/mixins/interfaces/MContractRegistry.sol
```diff
@@ -6,6 +6,22 @@ import "../../../rounds/IRoundsManager.sol";
...
+     */
+    modifier whenSystemNotPaused() {
+        _;
+    }
+
+    /**
+     * @dev Checks if the current round has been initialized
+     * @notice Executes the 'currentRoundInitialized' modifier in 'MixinContractRegistry'
```
yondonfu: I think the notice here should actually be a part of the `@dev` tag since it is developer documentation. Basing this off of the [NatSpec tag docs](https://solidity.readthedocs.io/en/v0.5.11/natspec-format.html#tags).

contracts/pm/mixins/MixinContractRegistry.sol
kyriediculous: I'm not sure a comment is necessary as the need for the change is part of the solidity documentation on [multiple inheritance and linearization](https://solidity.readthedocs.io/en/v0.5.3/contracts.html#multiple-inheritance-and-linearization)

The order in which the bases are read is right-to-left so if we'd switch this around we'd be calling the modifier in `MContractRegistry` which does nothing but continue the operation.

contracts/pm/mixins/interfaces/MContractRegistry.sol
kyriediculous: fixed in 752bec9

contracts/pm/mixins/interfaces/MContractRegistry.sol
kyriediculous: fixed in 752bec9

contracts/pm/mixins/MixinWrappers.sol
kyriediculous: fixed in 752bec9

contracts/pm/mixins/MixinTicketBrokerCore.sol
kyriediculous: fixed in 752bec9

contracts/pm/mixins/MixinTicketBrokerCore.sol
kyriediculous: fixed in 752bec9

contracts/pm/TicketBroker.sol
kyriediculous: fixed in 752bec9

contracts/pm/mixins/MixinContractRegistry.sol
yondonfu: > I'm not sure a comment is necessary as the need for the change is part of the solidity documentation

Fair enough! Though, I'd note that the real reason for not adding a comment is not that multiple inheritance is described in the Solidity docs, but rather that our unit tests ensure that the correct modifier behavior is exhibited when we test the functions that use the modifier so if we ever accidentally switched the inheritance order the tests would fail. While we do have the right test cases setup in this situation, generally we should be cautious around multiple inheritance because the end behavior may not always be so obvious - see this [blog post](https://pdaian.com/blog/solidity-anti-patterns-fun-with-inheritance-dag-abuse/) for a fun deep dive into a sneaky inheritance based backdoor added to a contract.

kyriediculous:
rebased !

PR #325, opened 2019-09-18
bonding: EarningsClaimed event

### LINKED ISSUE #305: Emit an event for claiming earnings
It would be helpful for the BondingManager contract to emit an event when earnings are claimed. Indexers could use these events to build of a history of earnings over time.

kyriediculous:
This PR adds an `EarningsClaimed` event to the `BondingManager` that will be emitted when a delegator claims earnings. 

```
   event EarningsClaimed(address indexed delegator, address indexed delegate, uint256 rewards, uint256 fees, uint256 startRound, uint256 endRound);
```

- added `EarningsClaimed` event definition to `BondingManager` interface
- emit `EarningsClaimed` event from `updateDelegatorWithEarnings`
- add unit tests to check if the event gets emitted correctly

Fixes #305 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -1029,23 +1029,32 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             // we know they will require too much gas to loop through all the necessary rounds to claim earnings
             // The user should instead manually invoke `claimEarnings` to split up the claiming process
             // across multiple transactions
-            require(_endRound.sub(del.lastClaimRound) <= maxEarningsClaimsRounds);
+            uint256 lastClaimRound = del.lastClaimRound;
+            require(_endRound.sub(lastClaimRound) <= maxEarningsClaimsRounds);
+            uint256 startRound = lastClaimRound.add(1);
 
             uint256 currentBondedAmount = del.bondedAmount;
             uint256 currentFees = del.fees;
 
-            for (uint256 i = del.lastClaimRound + 1; i <= _endRound; i++) {
+            for (uint256 i = startRound; i <= _endRound; i++) {
                 EarningsPool.Data storage earningsPool = transcoders[del.delegateAddress].earningsPoolPerRound[i];
 
                 if (earningsPool.hasClaimableShares()) {
-                    bool isTranscoder = _delegator == del.delegateAddress;
-
-                    (uint256 fees, uint256 rewards) = earningsPool.claimShare(currentBondedAmount, isTranscoder);
+                    (uint256 fees, uint256 rewards) = earningsPool.claimShare(currentBondedAmount, _delegator == del.delegateAddress);
 
                     currentFees = currentFees.add(fees);
                     currentBondedAmount = currentBondedAmount.add(rewards);
                 }
             }
+
+            emit EarningsClaimed(
+                del.delegateAddress,
+                _delegator,
+                currentBondedAmount.sub(del.bondedAmount),
+                currentFees.sub(del.fees),
+                startRound,
+                _endRound
+            );
 
             // Rewards are bonded by default
             del.bondedAmount = currentBondedAmount;
--- a/contracts/bonding/IBondingManager.sol
+++ b/contracts/bonding/IBondingManager.sol
@@ -16,6 +16,7 @@ contract IBondingManager {
     event Rebond(address indexed delegate, address indexed delegator, uint256 unbondingLockId, uint256 amount);
     event WithdrawStake(address indexed delegator, uint256 unbondingLockId, uint256 amount, uint256 withdrawRound);
     event WithdrawFees(address indexed delegator);
+    event EarningsClaimed(address indexed delegate, address indexed delegator, uint256 rewards, uint256 fees, uint256 startRound, uint256 endRound);
 
     // Deprecated events
     // These event signatures can be used to construct the appropriate topic hashes to filter for past logs corresponding
--- a/test/unit/BondingManager.js
+++ b/test/unit/BondingManager.js
@@ -4,6 +4,7 @@ import {contractId, functionSig, functionEncodedABI} from "../../utils/helpers"
 import {constants} from "../../utils/constants"
 import BN from "bn.js"
 import {expectRevertWithReason} from "../helpers/expectFail"
+import truffleAssert from "truffle-assertions"
 
 const BondingManager = artifacts.require("BondingManager")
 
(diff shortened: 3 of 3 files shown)


contracts/bonding/IBondingManager.sol
```diff
@@ -16,6 +16,7 @@ contract IBondingManager {
     event Rebond(address indexed delegate, address indexed delegator, uint256 unbondingLockId, uint256 amount);
     event WithdrawStake(address indexed delegator, uint256 unbondingLockId, uint256 amount, uint256 withdrawRound);
     event WithdrawFees(address indexed delegator);
+    event EarningsClaimed(address indexed delegator, address indexed delegate, uint256 rewards, uint256 fees, uint256 startRound, uint256 endRound);
```
yondonfu: Since the other events that include both a delegator and a delegate address specify the delegator address and then the delegate address what do you think about doing the same in this event?

contracts/bonding/BondingManager.sol
```diff
@@ -1047,9 +1048,19 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+                currentFees.sub(del.fees),
+                currentBondedAmount.sub(del.bondedAmount),
+                lastClaimRound + 1,
+                _endRound
+            );
+
             // Rewards are bonded by default
             del.bondedAmount = currentBondedAmount;
             del.fees = currentFees;
+
```
yondonfu: Extra newline here?

contracts/bonding/BondingManager.sol
```diff
@@ -1029,12 +1029,13 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             // we know they will require too much gas to loop through all the necessary rounds to claim earnings
             // The user should instead manually invoke `claimEarnings` to split up the claiming process
             // across multiple transactions
-            require(_endRound.sub(del.lastClaimRound) <= maxEarningsClaimsRounds);
+            uint256 lastClaimRound = del.lastClaimRound;
```
yondonfu: Can we instead store `uint256 startRound = del.lastClaimRound + 1` so that we can use `startRound` in the for loop conditional as well as when we emit the `EarningsClaimed` event? We would just need to change the below require statement to:

```
require(_endRound.sub(startRound) < maxEarningsClaimsRounds);
```

contracts/bonding/IBondingManager.sol
kyriediculous: Oh yeah sounds good, I didn't notice

contracts/bonding/BondingManager.sol
kyriediculous: We can do that but `_endRound.sub(startRound)` will revert when we invoke `autoClaimEarnings` twice in the same round so we'll still have to use `_endRound.sub(del.lastClaimedRound)`

contracts/bonding/BondingManager.sol
kyriediculous: fixed in d8c8a5f

contracts/bonding/IBondingManager.sol
kyriediculous: fixed in d8c8a5f

contracts/bonding/BondingManager.sol
yondonfu: Great point!

Can we avoid an extra storage read for `del.lastClaimRound` by storing `lastClaimRound` in memory?

Ex.

```
uint256 lastClaimRound = del.lastClaimRound;
require(_endRound.sub(lastClaimRound) <= maxEarningsClaimsRounds);

uint256 startRound = lastClaimRound.add(1);

// Rest of code
```

kyriediculous: With some work-arounds yes, otherwise we'll be running into the stack size limit and we'll pop off the `_delegator` argument , which becomes problematic if we need to access that value on the stack down the line (the notorious "stack too deep" error). 

I think removing `isTranscoder` should work, we can evaluate that expression directly instead.

It's for the same reason (stack size limit) that we are doing arithmetics in the line where the event gets logged. The obvious solution would be to keep track of a local `totalFees` and `totalRewards` variable, but due EVM restrictions it's not possible.

kyriediculous: Fixed in 9f4751a

contracts/bonding/BondingManager.sol
```diff
@@ -1029,24 +1029,33 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
                     currentFees = currentFees.add(fees);
                     currentBondedAmount = currentBondedAmount.add(rewards);
                 }
             }
 
+            emit EarningsClaimed(
+                del.delegateAddress,
+                _delegator,
+                currentFees.sub(del.fees),
```
yondonfu: The arguments here are in the wrong order - `rewards` comes before `fees`. The order should instead be `currentBondedAmount.sub(del.bondedAmount)` followed by `currentFees.sub(del.fees)`. See my other comment in this review for a note on why the tests are still passing.

test/unit/BondingManager.js
```diff
@@ -1841,6 +1841,20 @@ contract("BondingManager", accounts => {
             assert.equal((await bondingManager.getDelegator(delegator1))[5], currentRound + 1, "should update caller's lastClaimRound to the current round")
         })
 
+        it("fires a ClaimEarnings event", async () => {
```
yondonfu: Let's update the event name here to `EarningsClaimed`.

test/unit/BondingManager.js
```diff
@@ -1841,6 +1841,20 @@ contract("BondingManager", accounts => {
...
+            const expRewards = new BN(delegatorRewards * .3) // 30%
+            const expFees = new BN(delegatorFees * .3) // 30%
+            bondingManager.EarningsClaimed({from: delegator1}).on("data", e => {
```
yondonfu: This event emitter based test isn't working properly because the test completes before the assertions in the callback are executed.

If this event emitter based test is replaced with a test that uses `truffleAssert`, the new test will fail because of the incorrect argument ordering when the `EarningsClaimed` event is fired:

```
const txResult = await bondingManager.claimEarnings(currentRound + 1, {from: delegator1})
            
truffleAssert.eventEmitted(txResult, "EarningsClaimed", e => {
    return e.delegate === transcoder &&
               e.delegator == delegator1 &&
               e.fees == expFees.toString() &&
               e.rewards == expRewards.toString() &&
               e.startRound == (currentRound + 1).toString() &&
               e.endRound == (currentRound + 1).toString()
})
```

kyriediculous: There's other tests in BondingManager.js that use the eventEmitter, would this be true for those tests as well then? Should we replace them all with `truffleAssert.eventEmitted` ?

kyriediculous: fixed in 0ee7707

test/unit/BondingManager.js
kyriediculous: fixed in 0ee7707

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 0ee7707

test/unit/BondingManager.js
yondonfu: > There's other tests in BondingManager.js that use the eventEmitter, would this be true for those tests as well then?

Those might be a problem - let's take a look at those separately from this PR.

kyriediculous: I'll create an issue for it in the security audit prep project

kyriediculous: https://github.com/livepeer/protocol/issues/326

Issue #333, opened 2019-09-22
Update GenericMock.sol to return revert reasons when forwarding calls


kyriediculous:
Currently `GenerickMock.sol` uses a high-level version of the `CALL` opcode, which returns only whether the external call succeeded before solidity 0.5.0 , therefore we are currently not returning revert reasons to test for in the unit tests. 

We could use a low-level inline assembly-snippet to make the external call, but since we'll be upgrading to solidity 0.5.0 we can make this change after the compiler upgrade. After that high-level implementation of `CALL` does return the return data as a second return value and we can change the `GenericMock.execute` function to:

```
    function execute(address _target, bytes _data) external payable {
       (bool ok, bytes memory r) = _target.call.value(msg.value)(_data);
        require(ok, r);
    }
```

yondonfu:
Closed by #354

PR #335, opened 2019-09-23
bonding, rounds: revert reasons

### LINKED ISSUE #313: Add revert reasons to all contracts
We should make sure that the require statements in all contracts include revert reasons. Doing so will allow upgraded contracts to start returning revert reasons which could be helpful for debugging failed transactions.

kyriediculous:
This PR adds revert reasons to the require statements for the upgradeable contracts.

_NOTE:_ tests that use a `GenerickMock.sol` contract to execute methods in unit tests don't return revert reasons yet, see #333. 

_NOTE:_ `public view` functions seem to not return revert reasons as they are calls and not transactions.

Fixes #313 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -102,7 +102,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
     // Check if sender is TicketBroker
     modifier onlyTicketBroker() {
-        require(msg.sender == controller.getContract(keccak256("TicketBroker")));
+        require(
+            msg.sender == controller.getContract(keccak256("TicketBroker")),
+            "caller must be TicketBroker"
+        );
         _;
     }
 
@@ -117,13 +120,13 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
     // Check if sender is Verifier
     modifier onlyVerifier() {
-        require(msg.sender == controller.getContract(keccak256("Verifier")));
+        require(msg.sender == controller.getContract(keccak256("Verifier")), "caller must be Verifier");
         _;
     }
 
     // Check if current round is initialized
     modifier currentRoundInitialized() {
-        require(roundsManager().currentRoundInitialized());
+        require(roundsManager().currentRoundInitialized(), "current round is not initialized");
         _;
     }
 
@@ -250,7 +253,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             decreaseTotalStake(currentDelegate, del.bondedAmount);
         }
 
-        // Delegation amount must be > 0 - cannot delegate to someone without having bonded stake
+        // cannot delegate to someone without having bonded stake
         require(delegationAmount > 0, "delegation amount must be greater than 0");
         // Update delegate
         del.delegateAddress = _to;
@@ -329,7 +332,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundInitialized
         autoClaimEarnings
     {
-        // Caller must not be an unbonded delegator
         require(delegatorStatus(msg.sender) != DelegatorStatus.Unbonded, "caller must be bonded");
 
         // Process rebond using unbonding lock
@@ -351,8 +353,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundInitialized
         autoClaimEarnings
     {
-        // Caller must be an unbonded delegator
-        require(delegatorStatus(msg.sender) == DelegatorStatus.Unbonded);
+        require(delegatorStatus(msg.sender) == DelegatorStatus.Unbonded, "caller must be unbonded");
 
         // Set delegator's start round and transition into Pending state
         delegators[msg.sender].startRound = roundsManager().currentRound().add(1);
@@ -374,10 +375,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         Delegator storage del = delegators[msg.sender];
         UnbondingLock storage lock = del.unbondingLocks[_unbondingLockId];
 
-        // Unbonding lock must be valid
-        require(isValidUnbondingLock(msg.sender, _unbondingLockId));
-        // Withdrawal must be valid for the unbonding lock i.e. the withdraw round is now or in the past
-        require(lock.withdrawRound <= roundsManager().currentRound());
+        require(isValidUnbondingLock(msg.sender, _unbondingLockId), "invalid unbonding lock ID");
+        require(lock.withdrawRound <= roundsManager().currentRound(), "withdraw round must be before or equal to the current round");
 
         uint256 amount = lock.amount;
         uint256 withdrawRound = lock.withdrawRound;
@@ -399,8 +398,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundInitialized
         autoClaimEarnings
     {
-        // Delegator must have fees
-        require(delegators[msg.sender].fees > 0);
+        require(delegators[msg.sender].fees > 0, "no fees to withdraw");
 
         uint256 amount = delegators[msg.sender].fees;
         delegators[msg.sender].fees = 0;
@@ -418,10 +416,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function reward() external whenSystemNotPaused currentRoundInitialized {
         uint256 currentRound = roundsManager().currentRound();
 
-        // Sender must be an active transcoder
         require(isActiveTranscoder(msg.sender), "caller must be an active transcoder");
-
-        // Transcoder must not have called reward for this round already
         require(transcoders[msg.sender].lastRewardRound != currentRound, "caller has already called reward for the current round");
 
         Transcoder storage t = transcoders[msg.sender];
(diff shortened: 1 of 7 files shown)


yondonfu:
Mind switching the target branch for this PR to `streamflow`?

kyriediculous:
Done ! Probably overlapped opening this PR with merging `expand-active-transcoder-set` into `streamflow`

contracts/bonding/BondingManager.sol
```diff
@@ -568,7 +571,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         uint256 currentRound = roundsManager().currentRound();
         Delegator storage del = delegators[_delegator];
         
-        require(_endRound <= currentRound && _endRound > del.lastClaimRound, "end round must be before or equal to current round and after lastClaimRound");
+        // End round must be before or equal to current round and after lastClaimRound
```
yondonfu: Let's remove this comment since the revert reasons below already explain what the require statements do.

contracts/bonding/BondingManager.sol
```diff
@@ -593,8 +598,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function pendingFees(address _delegator, uint256 _endRound) public view returns (uint256) {
         uint256 currentRound = roundsManager().currentRound();
         Delegator storage del = delegators[_delegator];
+
         // End round must be before or equal to current round and after lastClaimRound
```
yondonfu: Let's remove this comment since the revert reasons below already explain what the require statements do.

test/unit/BondingManager.js
```diff
@@ -1,6 +1,6 @@
 import Fixture from "./helpers/Fixture"
-import expectThrow from "../helpers/expectThrow"
 import expectRevertWithReason from "../helpers/expectFail"
+import expectThrow from "../helpers/expectThrow"
```
yondonfu: Since we're still keeping this import we can just move it back to its original location above L2 to reduce the size of the diff.

contracts/bonding/BondingManager.sol
```diff
@@ -376,9 +379,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         UnbondingLock storage lock = del.unbondingLocks[_unbondingLockId];
 
         // Unbonding lock must be valid
-        require(isValidUnbondingLock(msg.sender, _unbondingLockId));
+        require(isValidUnbondingLock(msg.sender, _unbondingLockId), "invalid unbonding lock ID");
         // Withdrawal must be valid for the unbonding lock i.e. the withdraw round is now or in the past
-        require(lock.withdrawRound <= roundsManager().currentRound());
+        require(lock.withdrawRound <= roundsManager().currentRound(), "withdraw round must be equal to or smaller than the current round");
```
yondonfu: Let's try to use the same language as the other revert reasons when comparing values (I see that you're using the language "before or equal to"). This reason could be:

"withdraw round must be before or equal to the current round"

test/unit/BondingManager.js
```diff
@@ -1132,14 +1132,15 @@ contract("BondingManager", accounts => {
         beforeEach(async () => {
             await fixture.roundsManager.setMockBool(functionSig("currentRoundInitialized()"), true)
             await fixture.roundsManager.setMockBool(functionSig("currentRoundLocked()"), false)
-            await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound)
+            await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound - 1)
```
yondonfu: What is the reason for these additional test setup changes?

test/unit/RoundsManager.js
```diff
@@ -1,9 +1,9 @@
 import Fixture from "./helpers/Fixture"
 import expectThrow from "../helpers/expectThrow"
```
yondonfu: I think the `blockHash()` tests are the only tests that still use this helper. Can we update the `blockHash()` tests to use `expectRevertWithReason` and then remove this import?

test/unit/BondingManager.js
kyriediculous: otherwise the test for `"should fail for invalid unbonding lock ID"` would revert because the caller isn't bonded, rather than an invalid unbonding lock ID.

test/unit/BondingManager.js
kyriediculous: Yeah I had deleted this first, but then `public view` comments weren't returning revert reasons so I added it again, will fix the position

test/unit/RoundsManager.js
yondonfu: Ah! Experimented locally and interestingly the revert reason *is* there it just isn't used in `err.reason`:

Screen Shot 2019-09-23 at 4 20 47 PM

This [closed Truffle PR](https://github.com/trufflesuite/truffle/pull/2340) provides some background on the motivation behind not propagating the revert reason into `err.reason` for now.

For this PR, let's just stick with using `expectThrow()` then.

kyriediculous: Interesting. 

I very much share this thought and it was on my mind during the weekend as well: 

> **Very unclear to me if require should be used in view methods** and last week I opened an issue at EthPM suggesting we amend EIP1319 to define 'null' return values for its Read API because of the lack of clarity around how this stuff is supposed to work. I'm not sure what's normal here.

I'm okay with sticking to `expectThrow` for now although extracting the reason from the error message wouldn't be too hard.

test/unit/BondingManager.js
yondonfu: Hm that's strange. In the original test setup, `delegator` should still be `Bonded` in `currentRound + 1` after it calls `unbond()` so the test for an invalid unbonding lock ID should pass. Tests pass for me locally when using the old test setup.

test/unit/RoundsManager.js
yondonfu: Yeah I can see the benefits in moving revert logic out of view functions and just wrapping usage of view functions with a require statement if a state mutating function needs to use the return value of the view functions. But, I think its ok to leave the existing view function structure as-is at this stage since we can still access the revert reason if needed.

I think we should just use `expectThrow` in this PR for now since checking the revert reason for these view functions isn't critical at the moment.

test/unit/BondingManager.js
kyriediculous: I'll doublecheck

test/unit/RoundsManager.js
kyriediculous: I had done a quick check already to see if it would pass when checking the error message rather than the reason field

It does, but can always revert the changes back. 

2eb6a0f

test/unit/BondingManager.js
kyriediculous: Yeah works with the old set-up not sure what went wrong then. 

Fixed in 2eb6a0f

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 2eb6a0f

test/unit/BondingManager.js
kyriediculous: fixed in 2eb6a0f

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 2eb6a0f

test/unit/RoundsManager.js
yondonfu: Looks like CI tests are failing for a few cases after this change

kyriediculous: Was a little too fast pushing the commit, fixed now.

kyriediculous: I had to rebase after cherry-picking but tests should pass now

scripts/run_ganache.sh
```diff
@@ -10,7 +10,7 @@ ganache_running() {
 
 start_ganache() {
     if [ "$SOLIDITY_COVERAGE" = true ]; then
-        node_modules/.bin/testrpc-sc -l 0xfffffffffff -p "$ganache_port" -a 100 > /dev/null &
+        node_modules/.bin/testrpc-sc -k "petersburg" -l 0xfffffffffff -p "$ganache_port" -a 100 > /dev/null &
```
yondonfu: Petersburg should be the default right now given the version of `solidity-coverage`/`testrpc-sc` used so we can omit this flag.

kyriediculous:
rebased !

test/unit/RoundsManager.js
```diff
@@ -254,19 +253,22 @@ contract("RoundsManager", accounts => {
     describe("blockHash", () => {
         it("should fail if block is in the future", async () => {
             const latestBlock = await web3.eth.getBlockNumber()
-            await expectThrow(roundsManager.blockHash(latestBlock + 1))
+            // Note that current block = latestBlock + 1, so latestBlock + 2 is in the future
```
yondonfu: Actually, the current block number as returned by `web3.eth.getBlockNumber()` should be equal to `latestBlock` here since `eth_call` by default executes in the context of the latest block and not in the context of the pending block to be mined. So, in this case `latestBlock + 1` would be a block number in the future.

test/unit/RoundsManager.js
```diff
@@ -254,19 +253,22 @@ contract("RoundsManager", accounts => {
...
             const latestBlock = await web3.eth.getBlockNumber()
-            await expectThrow(roundsManager.blockHash(latestBlock - 257))
+            // Note that current block = latestBlock + 1, so latestBlock - 256 = current block - 257
+            await expectRevertWithReason(roundsManager.blockHash(latestBlock - 257), "can only retrieve hashes for last 256 blocks")
         })
 
         it("should fail if block is the current block", async () => {
             const latestBlock = await web3.eth.getBlockNumber()
-            await expectThrow(roundsManager.blockHash(latestBlock))
+            // Note that current block = latestBlock + 1
```
yondonfu: Based on my previous comment, `latestBlock` would actually be the current block here.

kyriediculous:
Fixed it up & rebased

PR #337, opened 2019-09-23
test: fix event tests

### LINKED ISSUE #326: Change tests for events to use truffleAssert instead of eventEmitted
When testing emitted contract events using the `eventEmitter`, the test seems to complete before the callback runs.

We can replace these tests by using `truffle.assert.eventEmitted()`

See [this comment](https://github.com/livepeer/protocol/pull/325#discussion_r326237324) as well

kyriediculous:
Fix up the tests that use the `EventEmitter` pattern, the callback doesn't seem to fire before the tests complete. Now using `truffleAssert.eventEmitted` instead.

Fix up the tests that were emitted by executing a contract call by calling one of the mock contracts, truffle doesn't include event logs if the event isn't included in the base contract's ABI. The logs are however included as raw (not decoded) logs, so a helper has been added to decode and check these 'deep' logs.

Created a separate commit 507bd7f that renames the event params for the `Unbond` event, these still contained `pending` causing the rewritten tests to fail. Not sure if I've overlooked this or if this was accidentally left out during a rebase. 

Adjusted & ran unit and integration tests

Fixes #326 

**Checklist:**


test/unit/BondingManager.js
```diff
@@ -132,14 +133,15 @@ contract("BondingManager", accounts => {
...
-                    })
-
                     await bondingManager.bond(1000, accounts[0])
-                    await bondingManager.transcoder(5, 10)
+                    const txRes = await bondingManager.transcoder(5, 10)
+
+                    truffleAssert.eventEmitted(
+                        txRes,
+                        "TranscoderUpdate",
+                        e => e.transcoder == accounts[0] && e.rewardCut == 5 && e.feeShare == 10 && e.registered == true,
```
yondonfu: Can omit the `true` and just do `e.registered` and if we're doing the opposite check can omit the `false` and just do `!e.registered`. Same comment applies to the rest of the event assertions that include `e.registered`.

test/unit/BondingManager.js
```diff
@@ -954,15 +955,15 @@ contract("BondingManager", accounts => {
...
-                    assert.equal(e.returnValues.amount, 500, "wrong amount in Unbond event")
-                    assert.equal(e.returnValues.withdrawRound, currentRound + 1 + unbondingPeriod, "wrong withdrawRound in Unbond event")
-                })
 
-                await bondingManager.unbond(500, {from: delegator})
+                const txRes = await bondingManager.unbond(500, {from: delegator})
+                truffleAssert.eventEmitted(
+                    txRes,
+                    "Unbond",
+                    e => (e.delegate == transcoder && e.delegator == delegator && e.unbondingLockId == unbondingLockID.toNumber() &&
```
yondonfu: Do we need the parens?

test/unit/BondingManager.js
```diff
@@ -412,15 +417,14 @@ contract("BondingManager", accounts => {
...
-                    assert.equal(e.returnValues.additionalAmount, 1000, "wrong additionalAmount in Bond event")
-                    assert.equal(e.returnValues.bondedAmount, 1000, "wrong bondedAmount in Bond event")
-                })
-
-                await bondingManager.bond(1000, transcoder0, {from: delegator})
+                const txRes = await bondingManager.bond(1000, transcoder0, {from: delegator})
+                truffleAssert.eventEmitted(
+                    txRes,
+                    "Bond",
+                    e => e.newDelegate == transcoder0 && e.oldDelegate == constants.NULL_ADDRESS &&
```
yondonfu: When we have many successive checks like this (i.e. > 3-4) I feel like having each check on a separate line can improve readability. 

Ex.

(Indentation could be updated)

```
truffleAssert.eventEmitted(txResult, "EarningsClaimed", e => {
    return e.delegate === transcoder &&
            e.delegator == delegator1 &&
            e.fees == expFees.toString() &&
            e.rewards == expRewards.toString() &&
            e.startRound == (currentRound + 1).toString() &&
            e.endRound == (currentRound + 1).toString()
})
```

WDYT?

test/unit/ServiceRegistry.js
```diff
@@ -46,12 +48,13 @@ contract("ServiceRegistry", accounts => {
...
-                assert.equal(e.returnValues.serviceURI, "foo", "wrong service URI in ServiceURIUpdate event")
-            })
-
-            await registry.setServiceURI("foo", {from: accounts[0]})
+            const txRes = await registry.setServiceURI("foo", {from: accounts[0]})
+            truffleAssert.eventEmitted(
+                txRes,
+                "ServiceURIUpdate",
+                e => e.addr == accounts[0] && e.serviceURI == "foo",
+                "ServiceURIUpdate event no emitted correctly"
```
yondonfu: "no emitted correctly" -> "not emitted correctly"

test/unit/BondingManager.js
```diff
@@ -1921,20 +1942,25 @@ contract("BondingManager", accounts => {
...
                         "slashTranscoder(address,address,uint256,uint256)",
                         ["address", "uint256", "uint256", "uint256"],
                         [transcoder, constants.NULL_ADDRESS, PERC_DIVISOR / 2, 0]
                     )
                 )
+
+                // 'expectEvent' is used to parse deep events that don't occur in the base contract being called
+                // These events are not included as decoded event logs in a transaction receipt's 'logs' field
+                // Rather they are still encoded as a hex string in the 'rawLogs' field
+                expectEvent(
```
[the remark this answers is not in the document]
kyriediculous: I thought the helper was pretty neat because it simply decodes the raw event logs according to the abi rather than calling `bondigManager.getPastEvents()` but I can restructure it to overwrite the decoded event logs on the transaction receipt , then return the new receipt.

That way we can do

```
truffleAssert.eventEmitted(
    decodeRawLogs(BondingManager.abi, txRes),
    "TranscoderSlashed",
    e => e.transcoder == transcoder && e.finder == constants.NULL_ADDRESS && e.penalty == 0 && e.finderReward == 0
)
```

I'm fine with either option though.

test/unit/BondingManager.js
kyriediculous: Yeah that might be a little neater indeed

test/unit/BondingManager.js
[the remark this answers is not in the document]
yondonfu: > rather than calling bondigManager.getPastEvents()

Ah didn't know that `truffleAssert.createTransactionResult()` uses `getPastEvents()` - thanks for the tip! It uses a filter that targets the specific block number of the provided transaction receipt though so its not as bad.

> it simply decodes the raw event logs according to the abi

I do like this approach! However, I think the helper will run into some trouble if the contract that returns `txRes` also emits an event as a part of the transaction. Take the `GenericMock` example and suppose `GenericMock` emitted an `Executed` event for `execute()`. Then, the receipt's `rawLogs` would include the `TranscoderSlashed` event AND the `Executed` event. Since `expectFail` takes in the ABI for `BondingManager` it will try to decode the `Executed` event and fail since that event is not a part of the provided ABI. This isn't an issue right now because `execute()` doesn't emit any events of its own, but `expectFail()` wouldn't be a generic drop-in unless additional changes are made to ensure that all events in `rawLogs` can be decoded regardless of what contract emitted them.

IMO just using `truffleAssert` for now feels simpler - WDYT?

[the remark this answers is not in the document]
kyriediculous: That's correct that this helper currently doesn't forsee decoding for multiple ABI's.

I'm okay with using `truffleAssert`.

package.json
```diff
@@ -60,7 +61,8 @@
     "solidity-coverage": "^0.5.11",
     "solium": "^1.1.8",
     "truffle": "5.0.0-next.17",
-    "truffle-keystore-provider": "^1.0.2",
-    "truffle-assertions": "0.6.3"
-  }
+    "truffle-assertions": "0.6.3",
+    "truffle-keystore-provider": "^1.0.2"
+  },
+  "dependencies": {}
```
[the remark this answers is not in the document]
kyriediculous: Ah no probably not I first installed `abi-decoder` in deps rather than devDeps , then uninstalled it which is why there's now an empty object there. 

Since we're using `truffleAssert.createTransactionResult()` we no longer need `abi-decoder` either, meaning we can completely revert all changes to `package.json`

test/unit/ServiceRegistry.js
kyriediculous: fixed in ad3cb40

test/unit/BondingManager.js
kyriediculous: fixed in ad3cb40

test/unit/BondingManager.js
kyriediculous: fixed in ad3cb40

test/unit/BondingManager.js
kyriediculous: fixed in ad3cb40

test/unit/BondingManager.js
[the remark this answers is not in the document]
kyriediculous: fixed in ad3cb40

package.json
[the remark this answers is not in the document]
kyriediculous: fixed in ad3cb40

test/unit/BondingManager.js
```diff
@@ -174,17 +178,22 @@ contract("BondingManager", accounts => {
...
                         const totalBonded = (await bondingManager.getTotalBonded()).toNumber()
 
                         // Caller bonds 6000 which is more than transcoder with least delegated stake
                         await bondingManager.bond(6000, newTranscoder, {from: newTranscoder})
-                        await bondingManager.transcoder(5, 10, {from: newTranscoder})
+                        const txRes = await bondingManager.transcoder(5, 10, {from: newTranscoder})
+                        truffleAssert.eventEmitted(
+                            txRes,
+                            "TranscoderUpdate",
+                            e => {
```
yondonfu: Since the rest of this PR omits the brackets and `return`, let's do the same here for consistency?

package.json
```diff
@@ -60,7 +60,7 @@
     "solidity-coverage": "^0.5.11",
     "solium": "^1.1.8",
     "truffle": "5.0.0-next.17",
-    "truffle-keystore-provider": "^1.0.2",
-    "truffle-assertions": "0.6.3"
+    "truffle-assertions": "0.6.3",
```
yondonfu: Can this change in this file and in `package-lock.json` be reverted since we are no longer introducing a new dependency in this PR?

kyriediculous: I think npm did this automatically (alphabetical order).

kyriediculous: fixed package-lock 35f5b36

but I think we can keep this change as it is convention to order dependencies alphabetically ?

test/unit/BondingManager.js
kyriediculous: fixed in 35f5b36

kyriediculous:
Will resolve merge conflicts when rebasing

kyriediculous:
rebased & resolved conflicts

PR #344, opened 2019-09-24
Improved test commands and misc file cleanup

### LINKED ISSUE #316: Clean up helper files
There are a bunch of helper files from V1 that are no longer used.

Examples:

- `utils/batchTranscodeReceipts.js`
- `utils/transcodeReceipt.js`

We can remove these helper files to clean up the codebase.

yondonfu:
This PR updates test commands so that `ganache-cli` is started and cleaned up after each test command. This avoids the problem where whenever we update `ganache-cli`, if two versions are incompatible (EVM version change for example) and we accidentally connect to an old instance from a previous test command we end up running into weird issues. Also:

- Cleans up `package.json` scripts
- Removes some unneeded utility files
- Adds `space-infix-ops` eslint rule since this has come up in some recent PR reviews

Fixes #316


yondonfu:
Changed base branch to `streamflow` and rebased.

yondonfu:
> I'm guessing for the changes to the npm scripts to use package names instead of paths, that npm will first try to resolve those dependencies in node_modules/.bin before checking globally installed modules?

Yep! See https://firstdoit.com/no-need-for-globals-using-npm-dependencies-in-npm-scripts-3dfb478908

kyriediculous:
Looks like the tests pass but coveralls throws an error at the end of the process

yondonfu:
Coveralls fixed!

PR #351, opened 2019-09-30
Various fixes

### LINKED ISSUE #345: Consider using <address>.call() instead of inline assembly in redeemWinningTicketNoRevert()
`TicketBroker.batchRedeemWinningTickets()` uses the `redeemWinningTicketNoRevert()` helper in order to redeem each of the tickets in a batch without reverting if any individual ticket redemption fails. Currently, the helper constructs the ABI encoded calldata for `redeemWinningTicket()` and then uses the calldata in an inline assembly call with the target contracts set to the `TicketBroker` itself. We should be able to accomplish the same thing with `<address>.call()` which requires less configuration. Since there does not seem to be a strong reason to use inline assembly here we should just not use it.

### LINKED ISSUE #346: Check for non-zero input in setTicketValidityPeriod()
`TicketBroker.setTicketValidityPeriod()` can require non-zero input in order to prevent the ticket validity period from ever being set to 0 which would mean that tickets are always invalid. In practice, if this mistake is ever made it could be immediately corrected by calling the function with a non-zero value, but adding a require statement can be an added precaution.

yondonfu:
This PR includes the following commits:

- 8b7797c: Use `<address>.call()` instead of assembly in TicketBroker. Fixes #345 

- d1c52ea: Adds a check for non-zero values in `TicketBroker.setTicketValidityPeriod()` and adds tests. Fixes #346 

- edfd64c: Adds tests for `TicketBroker.setUnlockPeriod()`

- f7e6ac3: Fix a linting error in BondingManager

- 4e64c16: Use an empty constructor in TicketBroker because any state updates in the constructor will not be invoked when interacting with a proxy that uses the TicketBroker implementation. We can just use the available setters instead.

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -561,7 +561,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function pendingStake(address _delegator, uint256 _endRound) public view returns (uint256) {
         uint256 currentRound = roundsManager().currentRound();
         Delegator storage del = delegators[_delegator];
-        
+
         require(_endRound <= currentRound, "end round must be before or equal to current round");
         require(_endRound > del.lastClaimRound, "end round must be after last claim round");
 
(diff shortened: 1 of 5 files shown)


kyriediculous (approved):
LGTM 🚀

PR #352, opened 2019-09-30
Revert Minter changes to match mainnet implementation

### LINKED ISSUE #347: Rollback modifier changes to Minter
The `Minter` contract on the `master` branch uses the `onlyBondingManagerOrJobsManager()` and the `onlyMinterOrJobsManager()` modifiers, but the `Minter` contract on the `streamflow` branch uses the `onlyBondingManagerOrTicketBroker()` and the `onlyMinterOrTicketBroker()` modifiers. Since we will not be upgrading the `Minter`, we should rollback these changes so that the `Minter` uses the `onlyBondingManagerOrJobsManager()` and the `onlyMinterOrJobsManager()` modifiers (the end result will be that the `Minter` implementation will be logically the same and the only changes contained in the repo would be syntactical + for compiler versioning). In practice, the `TicketBroker` can be registered under the contract ID `keccak256("JobsManager")` - the `TicketBroker` will then be able to use the `Minter` to deposit/withdraw escrowed ETH without any modifications to the currently deployed `Minter`.

yondonfu:
This PR reverts changes in the Minter to logically match the mainnet implementation (excluding compiler version, revert reasons, etc.). One thing to note is that contracts will use the JobsManager contract ID to interact with the TicketBroker.

Fixes #347

### DIFF
--- a/test/unit/helpers/Fixture.js
+++ b/test/unit/helpers/Fixture.js
@@ -23,15 +23,23 @@ export default class Fixture {
         this.minter = await this.deployAndRegister(MinterMock, "Minter")
         this.bondingManager = await this.deployAndRegister(BondingManagerMock, "BondingManager")
         this.roundsManager = await this.deployAndRegister(GenericMock, "RoundsManager")
+        this.jobsManager = await this.deployAndRegister(GenericMock, "JobsManager")
         this.ticketBroker = await this.deployAndRegister(GenericMock, "TicketBroker")
+        // Register TicketBroker with JobsManager contract ID because in a production system the Minter likely will not be upgraded to be
+        // aware of the TicketBroker contract ID and it will only be aware of the JobsManager contract ID
+        await this.register("JobsManager", this.ticketBroker.address)
         this.verifier = await this.deployAndRegister(GenericMock, "Verifier")
+    }
+
+    async register(name, addr) {
+        // Use dummy Git commit hash
+        const commitHash = web3.utils.asciiToHex("0x123")
+        await this.controller.setContractInfo(contractId(name), addr, commitHash)
     }
 
     async deployAndRegister(artifact, name, ...args) {
         const contract = await artifact.new(...args)
-        // Use dummy Git commit hash
-        const commitHash = web3.utils.asciiToHex("0x123")
-        await this.controller.setContractInfo(contractId(name), contract.address, commitHash)
+        await this.register(name, contract.address)
         return contract
     }
 
--- a/contracts/pm/mixins/MixinTicketProcessor.sol
+++ b/contracts/pm/mixins/MixinTicketProcessor.sol
@@ -18,7 +18,7 @@ contract MixinTicketProcessor is MContractRegistry, MTicketProcessor {
      */
     function processFunding(uint256 _amount) internal {
         // Send funds to Minter
-        minter().trustedDepositETH.value(_amount)();
+        minter().depositETH.value(_amount)();
     }
 
     /**
--- a/contracts/test/mocks/MinterMock.sol
+++ b/contracts/test/mocks/MinterMock.sol
@@ -4,12 +4,7 @@ import "./GenericMock.sol";
 
 contract MinterMock is GenericMock {
 
-    event TrustedBurnETH(uint256 amount);
     event TrustedWithdrawETH(address to, uint256 amount);
-
-    function trustedBurnETH(uint256 _amount) external {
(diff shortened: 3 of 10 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -101,10 +101,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     uint256 public nextRoundTotalActiveStake;
 
     // Check if sender is TicketBroker
-    modifier onlyTicketBroker() {
+    modifier onlyJobsManager() {
```
kyriediculous: Perhaps we should consider registering `TicketBroker` as both `TicketBroker` and `JobsManager`

This would reduce the amount of possible confusion. Additionally since `updateTranscoderWithFees` where this modifier is used does not involve any direct `Minter` logic, so I would keep the scope of reverting modifier name changes to the `Minter` contract. WDYT?

kyriediculous: Perhaps we should consider registering `TicketBroker` as both `TicketBroker` and `JobsManager` and revert the name changes here.

This would reduce the amount of possible confusion (also with the revert reason). Additionally since `updateTranscoderWithFees` where this modifier is used does not involve any direct `Minter` logic, so I would keep the scope of reverting modifier name changes to the `Minter` contract. WDYT?

test/unit/helpers/Fixture.js
```diff
@@ -23,7 +23,7 @@ export default class Fixture {
         this.minter = await this.deployAndRegister(MinterMock, "Minter")
         this.bondingManager = await this.deployAndRegister(BondingManagerMock, "BondingManager")
         this.roundsManager = await this.deployAndRegister(GenericMock, "RoundsManager")
-        this.ticketBroker = await this.deployAndRegister(GenericMock, "TicketBroker")
+        this.jobsManager = await this.deployAndRegister(GenericMock, "JobsManager")
```
kyriediculous: In continuation of my comment above, to avoid confusion, I would then change this to:

```
this.ticketBroker = await this.deployAndRegister(GenerickMock, "TicketBroker")
// Register TicketBroker as both TicketBroker and JobsManager
await this.controller.setContractInfo(contractId("JobsManager"), this.ticketBroker.address, web3.utils.asciiToHex("0x123"))
```

We can then also update the tests to use `fixture.ticketBroker` so that the semantics reflect the real logic.

contracts/bonding/BondingManager.sol
yondonfu: Agreed that it feels clearer for other contracts besides the Minter that *are* aware of the TicketBroker contract ID to use the `onlyTicketBroker` modifier. Fixed in c7d0ee4

test/unit/helpers/Fixture.js
yondonfu: Fixed in c7d0ee4

yondonfu:
> Out of pure interest, I'm wondering what the design decision was to allow redeploying a new minter rather than making it an upgradeable contract like the BondingManager. Hindsight is 20/20 but it would have allowed an upgrade path where the two concerns above are addressed. Although having an upgradeable contract to be the owner of the token contract would be the new concern then.

The Minter is the single contract that escrows all funds in the system - while other contracts handle the state accounting that dictates how much each account is owed, the Minter is the actual holder of staked LPT and deposited ETH. Furthermore, as you mention in your last statement, the Minter is also the owner of the token contract so it is able to mint LPT based upon the logic of its implementation. Given its mission critical role in the system and the amount of funds it could have in its custody, the implementation should:

- Not be easily changed - since the contract outsources most accounting logic to other contracts, it should be relatively feature bare
- Only be upgraded to fix serious bugs and should not be upgraded as a part of feature iteration

So, initially we decided to use the delegatecall proxy upgrade pattern for contracts that are likely to be upgraded as a part of feature iteration and this migration based upgrade pattern for the Minter which involves some more friction (for the reasons that you mentioned in your comment) to indicate both to ourselves as developers and to others that the upgrade mechanism is not meant to be used regularly and that changes to the contract are discouraged. One upgrade mechanism swaps the implementation of an existing contract and the other migrates to a new contract without changing the functionality of an existing contract. Though I can certainly understand the argument that it ultimately comes down to optics and some people may find the difference to be minute.

kyriediculous:
Thanks for the clarification ! 

Changes look good to me 🚀

yondonfu:
Rebased

PR #369, opened 2020-01-06
Funds migration to Refunder contract from JobsManager

yondonfu:
This PR adds support for funds migration from the `JobsManager` to a refunder address specified by the owner of the `Controller`. In practice, the address specified will be the [Refunder contract](https://github.com/livepeer/protocol/pull/368) which will allow users to withdraw a refund amount that matches the amount they currently have deposited with the `JobsManager`.

449bd36 is mainly test refactoring.

6b3e3c9 contains the actual updates to the `JobsManager`.

- Copy the existing `JobsManager` into `PreMigrationJobsManager` and replace usage of `JobsManager` with `PreMigrationJobsManager` in existing tests. This is done to make the diff for the `JobsManager` easier to read and avoids the need to make substantial changes to the tests to accommodate the changes that will be made to the `JobsManager`
- Add a `migrateFunds()` function to the `JobsManager` that is only callable by the owner of the `Controller`. This function can only be called once and will transfer a specified amount to a specified refunder address from the `Minter`. The amount will be the sum of outstanding deposits managed by the `JobsManager` and will be calculated off-chain using [this script](https://github.com/yondonfu/jobsmanager-deposits)
- Add `revert()` statements to all state mutating functions that can be called by anyone. This is to prevent additional state updates to the `JobsManager` after it is deprecated. 

The updated `JobsManager` code in this PR will be deployed as the target implementation for the current `JobsManager` proxy. After this target implementation upgrade, funds will be migrated to the `Refunder` contract and all state mutating functions on the `JobsManager` proxy will be disabled [1].

See the [upgrade plan](https://gist.github.com/yondonfu/c87756debb08d7839105c9bbece709c7) for more details on how funds migration and the `Refunder` contract will work.

[1] Besides the functions that are only callable by the `Controller` owner - these functions only update global contract params and do not affect user state.


kyriediculous (commented):
This overall looks good to me, do you think it's worth adding a revert reason to prevent the state updates along the lines of "contract deprecated since streamflow" ? 

For code readability I would have perhaps opted for a modifier to be ran before `whenSystemNotPaused` 

```
modifier preventStateUpdates {
 revert("contract deprecated since streamflow update");
_;
}
```
 
```
func deposit() external payable preventStateUpdates whenSystemNotPaused() { ... }
```

Feel free to do with the feedback as you like though, I have no strong preference for using a modifier here as the contract won't be in use anymore essentially.

yondonfu:
> do you think it's worth adding a revert reason to prevent the state updates along the lines of "contract deprecated since streamflow" ?

Unfortunately, revert reason strings were not supported by solc until [v0.4.22](https://github.com/ethereum/solidity/releases/tag/v0.4.22), but the `master` branch is using Truffle v4.0.4 which bundles solc v0.4.18 (older versions of Truffle did not allow you to choose a solc version). If we want to use a revert reason string here we would need to upgrade to a newer version of Truffle to compile the contracts with solc >= v0.4.22 which introduces the potential headache of upgrading Truffle (there may be changes in other areas of the codebase required?). Additionally, using a different compiler version would be one more change (albeit small) to the `JobsManager`. In the interest of keeping the changes required to upgrade the `JobsManager` as small as possible, I'm ok with not using a revert reason string here.

> For code readability I would have perhaps opted for a modifier to be ran before whenSystemNotPaused

Agreed that the modifier improves code readability (as well as the readability of the diff). 73a6190 uses the `noStateUpdates()` modifier.

kyriediculous:
Unfortunate RE revert reasons.

Changes look good, let's rebase !

yondonfu:
Rebased

kyriediculous (approved):
approved pending green CI

PR #383, opened 2020-07-13
LIP-25: Governor contract

kyriediculous:
This PR implements the `Governor` contract according to the spec defined in LIP25 https://github.com/livepeer/LIPs/pull/33

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -1,0 +1,116 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+/**
+ * @title Governor
+ * @dev The Governor holds the rights to stage and execute contract calls i.e. changing Livepeer protocol parameters.
+ */
+contract Governor {
+
+    using SafeMath for uint256;
+
+    address public owner;
+
+    /// @dev mapping of updateHash (keccak256(update) => executeBlock (block.number + delay)
+    mapping(bytes32 => uint256) public updates;
+
+    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
+
+    event UpdateStaged(Update update, uint256 delay);
+
+    event UpdateExecuted(Update update);
+
+    event UpdateCancelled(Update update);
+
+    struct Update {
+        address[] target;
+        uint256[] value;
+        bytes[] data;
+        uint256 nonce;
+    }
+
+   /// @notice Throws if called by any account other than the owner.
+    modifier onlyOwner() {
+        require(msg.sender == owner, "unauthorized: msg.sender not owner");
+        _;
+    }
+
+
+    /// @notice Throws if called by any account other than this contract.
+    /// @dev Forces the `stage/execute` path to be used to call functions with this modifier instead of directly.
+    modifier onlyThis() {
+        require(msg.sender == address(this), "unauthorized: msg.sender not Governor");
+        _;
+    }
+
+
+    /// @dev The Ownable constructor sets the original `owner` of the contract to the sender account.
+    constructor() public {
+        owner = msg.sender;
+        emit OwnershipTransferred(address(0), msg.sender);
+    }
+
+
+    /// @notice Allows the current owner to transfer control of the contract to a newOwner.
+    /// @dev Can only be called through stage/execute, will revert if the caller is not this contract's address.
+    /// @param newOwner The address to transfer ownership to.
+    function transferOwnership(address newOwner) public onlyThis {
+        require(newOwner != address(0), "newOwner is a null address");
+        emit OwnershipTransferred(owner, newOwner);
+        owner = newOwner;
+    }
+
+    /// @notice Stage a batch of updates to be executed.
+    /// @dev Reverts if the 'msg.sender' is not the 'owner'
+    /// @dev Reverts if an update is already staged
+    /// @param _update Update to be staged.
+    /// @param _delay (uint256) Delay (in number of blocks) for the update.
+    function stage(Update memory _update, uint256 _delay) public onlyOwner {
+        bytes32 updateHash = keccak256(abi.encode(_update));
+
+        require(updates[updateHash] == 0, "update already staged");
+
+        updates[updateHash] = block.number.add(_delay);
+
+        emit UpdateStaged(_update, _delay);
+    }
+
+    /// @notice Execute a staged update.
+    /// @dev Updates are authorized during staging.
+    /// @dev Reverts if a transaction can not be executed.
+    /// @param _update  Update to be staged.
+    function execute(Update memory _update) public payable {
+        bytes32 updateHash = keccak256(abi.encode(_update));
+        uint256 executeBlock = updates[updateHash];
+
+        require(executeBlock != 0, "update is not staged");
+        require(block.number >= executeBlock, "delay for update not expired");
+
+        // prevent re-entry and replay
+        delete updates[updateHash];
+        for (uint256 i = 0; i < _update.target.length; i++) {
+            /* solium-disable-next-line */
+            (bool success, bytes memory returnData) = _update.target[i].call.value(_update.value[i])(_update.data[i]);
+            require(success, string(returnData));
+        }
+
+        emit UpdateExecuted(_update);
+    }
+
+    /// @notice Cancel a staged update.
+    /// @dev Reverts if an update does not exist.
+    /// @dev Reverts if the 'msg.sender' is not the 'owner'
+    /// @param _update Update to be cancelled.
+    function cancel(Update memory _update) public onlyOwner {
+        bytes32 updateHash = keccak256(abi.encode(_update));
+        uint256 executeBlock = updates[updateHash];
+
+        require(executeBlock != 0, "update is not staged");
+        delete updates[updateHash];
+
+        emit UpdateCancelled(_update);
+    }
+}
--- a/test/integration/GovernorUpdate.js
+++ b/test/integration/GovernorUpdate.js
@@ -1,0 +1,231 @@
+import truffleAssert from "truffle-assertions"
+import {utils} from "ethers"
+
+import {constants} from "../../utils/constants"
+import {contractId} from "../../utils/helpers"
+
+const Controller = artifacts.require("Controller")
+const BondingManager = artifacts.require("BondingManager")
+const Governor = artifacts.require("Governor")
+const Minter = artifacts.require("Minter")
+
+contract("Governor update", accounts => {
+    let controller
+    let bondingManager
+    let governor
+
+    before(async () => {
+        controller = await Controller.deployed()
+        await controller.unpause()
+
+        const bondingManagerAddr = await controller.getContract(contractId("BondingManager"))
+        bondingManager = await BondingManager.at(bondingManagerAddr)
+
+        // Deploy Governor
+        governor = await Governor.new()
+        // Transfer Controller ownership to Governor
+        await controller.transferOwnership(governor.address)
+    })
+
+    it("controller is now owned by the governor", async () => {
+        assert.equal(await controller.owner(), governor.address)
+    })
+
+    it("governor has the correct owner", async () => {
+        assert.equal(await governor.owner(), accounts[0])
+    })
+
+    describe("single param change", () => {
+        it("reverts when the param change is not initiated through the governor", async () => {
+            await truffleAssert.reverts(
+                bondingManager.setNumActiveTranscoders(20),
+                "caller must be Controller owner"
+            )
+        })
+
+        it("reverts when the delay for the staged update has not expired", async () => {
+            const data = utils.hexlify(utils.arrayify(bondingManager.contract.methods.setNumActiveTranscoders(20).encodeABI()))
+            const update = {
+                target: [bondingManager.address],
+                value: ["0"],
+                data: [data],
+                nonce: 0
+            }
+            await governor.stage(update, 10)
+
+            await truffleAssert.reverts(
+                governor.execute(update),
+                "delay for update not expired"
+            )
+        })
+
+        it("succesfully executes a single param change", async () => {
+            const data = utils.hexlify(utils.arrayify(bondingManager.contract.methods.setNumActiveTranscoders(30).encodeABI()))
+            const update = {
+                target: [bondingManager.address],
+                value: ["0"],
+                data: [data],
+                nonce: 0
+            }
+            await governor.stage(update, 0)
+
+            let tx = await governor.execute(update)
+            assert.equal((await bondingManager.getTranscoderPoolMaxSize()).toNumber(), 30)
+
+            truffleAssert.eventEmitted(
+                tx,
+                "UpdateExecuted",
+                e => e.update.target[0] == update.target[0]
+                    && e.update.value[0] == update.value[0]
+                    && e.update.data[0] == update.data[0],
+                "UpdateStaged event not emitted correctly"
+            )
+        })
+    })
+
+    describe("complex update: migrate to new Minter", () => {
+        // Minter upgrade steps
+        // 1. Pause the protocol
+        // 2. call migrateToNewMinter
+        // 3. register the new Minter
+        // 4. Unpause the protocol
+        let minter
+        let newMinter
+
+        let pauseData
+        let pauseTarget
+        let migrateData
+        let migrateTarget
+        let setInfoData
+        let setInfoTarget
+        let unpauseData
+        let unpauseTarget
+
+        before(async () => {
+            const minterAddr = await controller.getContract(contractId("Minter"))
+            minter = await Minter.at(minterAddr)
+            newMinter = await Minter.new(controller.address, "100", "1", "500000")
+
+            pauseData = utils.hexlify(utils.arrayify(controller.contract.methods.pause().encodeABI()))
+            pauseTarget = controller.address
+
+            migrateData = utils.hexlify(utils.arrayify(minter.contract.methods.migrateToNewMinter(newMinter.address).encodeABI()))
+            migrateTarget = minter.address
+
+            setInfoData = utils.hexlify(utils.arrayify(controller.contract.methods.setContractInfo(contractId("Minter"), newMinter.address, "0x123").encodeABI()))
+            setInfoTarget = controller.address
+
+            unpauseData = utils.hexlify(utils.arrayify(controller.contract.methods.unpause().encodeABI()))
+            unpauseTarget = controller.address
+        })
+
+
+        it("step 1 'pause' fails: the protocol is already paused", async () => {
+            // pause twice
+            const update = {
+                target: [pauseTarget, pauseTarget, migrateTarget, setInfoTarget, unpauseTarget],
+                value: ["0", "0", "0", "0", "0"],
+                data: [pauseData, pauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update))
+        })
+
+        it("step 2 'migrateToNewMinter' fails: the protocol is not paused", async () => {
+            // omit pausing from the update
+            const update = {
+                target: [pauseTarget, unpauseTarget, migrateTarget, setInfoTarget, unpauseTarget],
+                value: ["0", "0", "0", "0", "0"],
+                data: [pauseData, unpauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update), "system is not paused")
+        })
+
+        it("step 2 'migrateToNewMinter' fails: new Minter cannot be current Minter", async () => {
+            // the previous test should have reverted in it's entirety
+            // so we should not run into an "already paused" error
+
+            migrateData = utils.hexlify(utils.arrayify(minter.contract.methods.migrateToNewMinter(minter.address).encodeABI()))
+
+
+            const update = {
+                target: [pauseTarget, migrateTarget, setInfoTarget, unpauseTarget],
+                value: ["0", "0", "0", "0"],
+                data: [pauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update), "new Minter cannot be current Minter")
+        })
+
+        it("step 2 'migrateToNewMinter' fails: new Minter cannot be null address", async () => {
+            migrateData = utils.hexlify(utils.arrayify(minter.contract.methods.migrateToNewMinter(constants.NULL_ADDRESS).encodeABI()))
+
+
+            const update = {
+                target: [pauseTarget, migrateTarget, setInfoTarget, unpauseTarget],
+                value: ["0", "0", "0", "0"],
+                data: [pauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update), "new Minter cannot be null address")
+        })
+
+        it("step 3 'setContractInfo' fails: wrong target", async () => {
+            migrateData = utils.hexlify(utils.arrayify(minter.contract.methods.migrateToNewMinter(newMinter.address).encodeABI()))
+
+            const update = {
+                target: [pauseTarget, migrateTarget, migrateTarget, unpauseTarget],
+                value: ["0", "0", "0", "0"],
+                data: [pauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update))
+        })
+
+        it("step 4 'unpause' fails: system is already unpaused", async () => {
+            // call unpause twice in the update
+            const update = {
+                target: [pauseTarget, migrateTarget, setInfoTarget, unpauseTarget, unpauseTarget],
+                value: ["0", "0", "0", "0", "0"],
+                data: [pauseData, migrateData, setInfoData, unpauseData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await truffleAssert.reverts(governor.execute(update))
+        })
+
+        it("succesfully executes all updates", async () => {
+            const update = {
+                target: [pauseTarget, migrateTarget, setInfoTarget, unpauseTarget],
+                value: ["0", "0", "0", "0"],
+                data: [pauseData, migrateData, setInfoData, unpauseData],
+                nonce: 0
+            }
+
+            // run the migrate to new minter update
+            await governor.stage(update, "0")
+            await governor.execute(update)
+
+            const actualNewMinterAddr = (await controller.getContractInfo(contractId("Minter")))[0]
+            assert.equal(actualNewMinterAddr, newMinter.address)
+        })
+    })
+})
--- a/test/unit/Governor.js
+++ b/test/unit/Governor.js
@@ -1,0 +1,695 @@
+import Fixture from "./helpers/Fixture"
+
+import truffleAssert from "truffle-assertions"
+import {utils, constants} from "ethers"
+import {BN} from "ethereumjs-util"
+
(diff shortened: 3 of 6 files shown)


contracts/zeppelin/ECTools.sol
```diff
@@ -0,0 +1,49 @@
+pragma solidity ^0.5.11;
```
yondonfu: Looks like we can remove this file.

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,102 @@
...
+contract Governor {
+
+    address public owner;
+
+    uint256 public nonce;
+
+    /// @dev mapping of updateHash (keccak256(update) => executeBlock (block.number + delay)
+    mapping(bytes32 => uint256) updates; 
+
+    using SafeMath for uint256;
```
yondonfu: Let's place this directive above the rest of the state variable declarations to be consistent with the style used by the other contracts in this repo. Would be nice to have the Solidity linter check for this...not sure if it supports this style rule though.

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,102 @@
...
+    /// @param newOwner The address to transfer ownership to.
+
+    function transferOwnership(address newOwner) public onlyThis {
+        require(newOwner != address(0));
+        owner = newOwner;
+        emit OwnershipTransferred(owner, newOwner);
+    }
+    
+    /// @notice Stage a batch of updates to be executed
+    /// @dev Anyone can execute this but only transactions signed by the owner are valid
```
yondonfu: Let's update this comment because we're now using msg.sender based authentication instead of signature based authentication for the updates.

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,102 @@
...
+        bytes32 updateHash = keccak256(abi.encode(_update));
+        uint256 executeBlock = updates[updateHash];
+
+        require(executeBlock != 0, "update is not staged");
+        require(block.number >= executeBlock, "delay for update not expired");
+
+        // prevent re-entry and replay
+        delete updates[updateHash];
+        for (uint256 i = 0; i < _update.target.length; i++) {
+            (bool success,) = _update.target[i].call.value(_update.value[i])(_update.data[i]);
```
yondonfu: The linter is [complaining](https://app.circleci.com/pipelines/github/livepeer/protocol/20/workflows/a4dd81b5-ddc3-40a1-935b-453c40ee1fcc/jobs/1410/steps) about this line because the security plugin by default disallows the use of `.call.value()`. I actually now disagree with this style rule (given the need to use it in settings such as this one and given that when just sending ETH `.transfer()`'s gas stipend can introduce issues due to opcode gas cost changes) so I'd be in favor of updating `.soliumrc.json` to turn off the rule altogether.

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,102 @@
...
+        owner = newOwner;
+        emit OwnershipTransferred(owner, newOwner);
+    }
+    
+    /// @notice Stage a batch of updates to be executed
+    /// @dev Anyone can execute this but only transactions signed by the owner are valid
+    /// @param _update Update to be staged
+    /// @param _delay (uint256) Delay (in number of blocks) for the update
+    function stage(Update memory _update, uint256 _delay) public onlyOwner {
+        require(nonce < _update.nonce, "nonce too low");
```
yondonfu: Having the nonce in the update is useful to distinguish between updates that could contain the same tx data, but I'm not sure that the contract needs to keep track of a global nonce. Do you have a scenario that you're thinking of that global nonce tracking would be required for?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,102 @@
...
+    event UpdateExecuted(Update update);
+
+    struct Update {
+        address[] target;
+        uint256[] value;
+        bytes[] data;
+        uint256 nonce;
+    }
+
+/**
```
yondonfu: I see that the comment style here is different from the style used in the rest of the file. Can we be consistent with the comment style we use?

test/unit/Governor.js
```diff
@@ -0,0 +1,444 @@
...
+    let setUint256
+
+    before(() => {
+        fixture = new Fixture(web3)
+    })
+
+    beforeEach(async () => {
+        await fixture.setUp()
+        ecTools = await ECTools.new()
+        await Governor.link("ECTools", ecTools.address)
```
yondonfu: Looks like we can remove all references to ECTools.

test/unit/Governor.js
```diff
@@ -0,0 +1,444 @@
...
+
+            await truffleAssert.reverts(
+                governor.stage(
+                    {
+                        target: [setUint256.address],
+                        value: ["0"],
+                        data: [data],
+                        nonce: "1"
+                    },
+                    "115792089237316195423570985008687907853269984665640564039457584007913129639935"
```
yondonfu: If we're already pulling in `ethers`, it could be clearer to use `ethers.constants.MaxUint256.add(utils.bigNumberify(1)` here.

test/unit/Governor.js
```diff
@@ -0,0 +1,444 @@
...
+                )
+            )
+        })
+
+        it("stages an update: emit UpdateStaged event and increase updateCount and nonce", async () => {
```
yondonfu: "updateCount" seems to imply there is an additional var that is being updated. If that is not the case I suggest removing it.

yondonfu (changes requested):

contracts/governance/Governor.sol
kyriediculous: Right, this is not as useful without signature based auth. 

I guess it's useful as an index for the edge case where there's multiple updates in a single block, otherwise blocknumbers would suffice to have historical ordering.

kyriediculous:
moved solium package to ethlint in 93301cc 

https://github.com/duaraghav8/Ethlint

> If you're currently using the solium package for npm install, it is highly recommended that you move to ethlint. The solium package will not receive updates after December, 2019. There are no differences between the updates pushed to ethlint and solium packages.

test/unit/Governor.js
kyriediculous: fixed in 1de239d

test/unit/Governor.js
kyriediculous: fixed in cd04b83

test/unit/Governor.js
kyriediculous: fixed in cd04b83

contracts/governance/Governor.sol
kyriediculous: fixed in cd04b83

contracts/governance/Governor.sol
kyriediculous: fixed in cd04b83

contracts/governance/Governor.sol
kyriediculous: fixed in cd04b83

contracts/governance/Governor.sol
kyriediculous: fixed in cd04b83

contracts/zeppelin/ECTools.sol
kyriediculous: fixed in cd04b83

contracts/governance/Governor.sol
yondonfu: The `UpdateStaged` and `UpdateExecuted` events within a single block should be emitted in the order that updates are staged or executed. I believe this ordering is achieved by attaching an index to each log within a block (would be good to double check this though).

Furthermore, I think having a nonce included in the update makes sense, but I don't think the contract needs to track a global nonce and do any checks on the nonce included in an update.

yondonfu (changes requested):
Looks like Coveralls is complaining about reduced coverage due to uncovered lines in the Governor - see the [latest report](https://coveralls.io/builds/32156721/source?filename=contracts/governance/Governor.sol).

contracts/governance/Governor.sol
kyriediculous: Ordering of transactions within a single block are determined by the miner based on transaction nonce and price (and can thus vary between implementations) thus I don't think there's guarantees that log index (or nonce tracked client-side by events) ordering is preserved (Unless `owner` is an EOA but in production this will not be the case, right?). 

The nonce in this contract does provide that guarantee, so I think the question is whether the tradeoff of a `SLOAD` and `STORE` against having hard ordering is worth it. 

The way the delay mapping works in this implementation should provide ample replay and re-entry prevention (it would be good to have this as a highlight for the audit) so it's more a nice to have rather than an explicit security requirement.

kyriediculous:
I will add the necessary tests to bring coverage up.

contracts/governance/Governor.sol
yondonfu: Hm I actually think regular transaction nonces already handle the scenario that you're describing.

2 distinct `stage()`/`update()` calls can be included in the same block in the following scenarios:

1. If the contract owner batched the calls together into a single atomic transaction
2. If the contract owner made two separate calls in two separate transactions and those transactions happen to be included in the same block

In scenario 1, the ordering of the calls is defined by the signer of the transaction.

In scenario 2, regardless of whether the contract owner itself is a contract or an EOA, each of the transactions would have a nonce that guarantees execution order - the transaction that was created/signed first should have a lower nonce and should be executed before the second transaction.

yondonfu (commented):
Should there also be a function to cancel staged updates in the scenario that an update is staged, but before it is executed there is a desire to not move forward with it?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,101 @@
...
+    /// @dev The Ownable constructor sets the original `owner` of the contract to the sender account.
+    constructor() public {
+        owner = msg.sender;
+    }
+
+
+    /// @notice Allows the current owner to transfer control of the contract to a newOwner.
+    /// @dev Can only be called through stage/execute, will revert if the caller is not this contract's address.
+    /// @param newOwner The address to transfer ownership to.
+
```
yondonfu: Let's fix this spacing.

contracts/governance/Governor.sol
kyriediculous: In scenario 1 while true, other stakeholders depend on the benevolence of the `owner` to define the right nonces/indexes. There are no barriers of submitting multiple transactions with the same nonce or older nonces if it is not tracked and checked. 

Full transparency does come at a cost, in this case that of a storage read and write.

Scenario 2 does only uphold for EOA's because contract accounts will not increment nonces unless for `CREATE` (and I think `CREATE2` opcodes) 

From the [Ethereum yellow paper](http://gavwood.com/paper.pdf), page 3, section 4.1 'World State' :  

> nonce: A scalar value equal to the number of transactions sent from this address or, in the case
of accounts with associated code, the number of
contract-creations made by this account. For account of address a in state σ, this would be formally denoted σ[a]n.

kyriediculous:
> Should there also be a function to cancel staged updates in the scenario that an update is staged, but before it is executed there is a desire to not move forward with it?

From a feature perspective it's a similar discussion as the nonce in my opinion: 

Is social convention enough for updates cancellation (e.g. staged but never executed upon). 

Or do we want guarantees that a cancelled update can never be executed (explicit cancellation function) 

From a technical perspective having explicit cleanups would be nice for reducing unnecessary state growth. 

But again it comes at a cost. 

Also we'd have to figure out ( and probably put in scope for this LIP) what the cancellation process would look like and how that would work with delays.

contracts/governance/Governor.sol
yondonfu: > There are no barriers of submitting multiple transactions with the same nonce or older nonces if it is not tracked and checked.

Right, but the contract owner can always submit the same update with a nonce that passes the contract nonce check. So, I'm not sure that I see what the contract nonce check protects against in this scenario. Since the contract owner has the authority to call `stage()`, it can stage whatever update that it wants.

> Scenario 2 does only uphold for EOA's because contract accounts will not increment nonces

The scenario I am referring to here involves a contract calling the Governor in two separate transactions (the contract calls against the Governor would be "internal transactions", but those shouldn't be relevant here since we just care about the other transactions that triggered the contract calls in the first place). In order for there to be two separate transactions, an EOA would have to submit two separate transactions to the contract. Each of these transactions will need to have a unique nonce.

yondonfu:
> Or do we want guarantees that a cancelled update can never be executed (explicit cancellation function)

I think this is preferable since after the delay for a staged update any account can call the `execute()` function to  execute the update which means never executing a staged update would be insufficient.

> Also we'd have to figure out ( and probably put in scope for this LIP) what the cancellation process would look like and how that would work with delays.

We could allow a `cancel()` function to be called for a staged update as long as it hasn't been executed yet (in this case the staged update would have been removed from the Governor anyway). One reason why a staged update would be canceled is if during the delay a bug in the implementation was discovered. The social convention would be that the `cancel()` function should only be used for this reason.

kyriediculous:
482bdbb - adds a `cancel(Update memory _update)` function that works similar to the other Governor functions

56c3078 - forwards revert reasons from external calls 

e7c13ef - adds test for our modified `Ownable.sol` implementation (mainly the `onlyThis` modifier) 

5ef3569 - misc fixes

contracts/governance/Governor.sol
kyriediculous: fixed in d700e93

contracts/governance/Governor.sol
kyriediculous: > Right, but the contract owner can always submit the same update with a nonce that passes the contract nonce check.

That's true but the owner can't submit an update with the same nonce regardless, it's about ordering not security at this point.

The owner could theoretically submit a batch of `stage()` calls within the same transaction using a smart contract wallet. 

I'm okay with removing the nonce notion since it's unlikely this scenario will ever take place though, but let's be mindful of it.

kyriediculous: removed notion of a nonce in 9050b40

kyriediculous:
This should be ready for final review now.

once this PR is approved we can move it to last call. I'd like to clean up the commit history a little bit after final review though.

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
+pragma solidity ^0.5.11;
```
kyriediculous: Note to self: fix compiler version, this is considered good practice. Not really super important for our use case but if someone compiles this contract with a different compiler version above 0.5.11 installed there _could_ be differences in behaviour (or exposure to vulnerabilities that could be in a specific compiler version that isn't as widely used).

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+    modifier onlyOwner() {
+        require(msg.sender == owner, "unauthorized");
+        _;
+    }
+
+
+    /// @notice Throws if called by any account other than this contract.
+    /// @dev Forces the `stage/execute` path to be used to call functions with this modifier instead of directly.
+    modifier onlyThis() {
+        require(msg.sender == address(this), "unauthorized");
```
yondonfu: Should the revert reason be more specific to differentiate between other authorization related modifiers (i.e. `onlyOwner`)? i.e. "unauthorized: msg.sender not Governor"

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    struct Update {
+        address[] target;
+        uint256[] value;
+        bytes[] data;
+    }
+
+   /// @notice Throws if called by any account other than the owner.
+    modifier onlyOwner() {
+        require(msg.sender == owner, "unauthorized");
```
yondonfu: Should the revert reason be more specific to differentiate between other authorization related modifiers (i.e. `onlyThis`)?

Ex. "unauthorized: msg.sender not owner"

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+    constructor() public {
+        owner = msg.sender;
+    }
+
+
+    /// @notice Allows the current owner to transfer control of the contract to a newOwner.
+    /// @dev Can only be called through stage/execute, will revert if the caller is not this contract's address.
+    /// @param newOwner The address to transfer ownership to.
+    function transferOwnership(address newOwner) public onlyThis {
+        require(newOwner != address(0), "newOwner is zero");
```
yondonfu: "null address" instead of "zero"?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+        emit OwnershipTransferred(owner, newOwner);
```
yondonfu, line 62: The convention used in this repo is to emit the event last after any effects & interactions - let's continue to follow that convention here?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+        owner = newOwner;
+    }
+    
+    /// @notice Stage a batch of updates to be executed.
+    /// @dev Reverts if the 'msg.sender' is not the 'owner'
+    /// @param _update Update to be staged.
+    /// @param _delay (uint256) Delay (in number of blocks) for the update.
+    function stage(Update memory _update, uint256 _delay) public onlyOwner {
+        bytes32 updateHash = keccak256(abi.encode(_update));
+        updates[updateHash] = block.number.add(_delay);
```
yondonfu, line 76: Since we don't check if the update is already staged, it would be possible for the same update to be staged multiple times which would just set the execute block farther in the future. 

Should we be more strict here and disallow staging an existing update by adding `require(updates[updateHash] == 0, ...)`?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        emit UpdateExecuted(_update);
+    }
+
+    /// @notice Cancel a staged update.
+    /// @dev Reverts if an update does not exist.
+    /// @dev Reverts if the 'msg.sender' is not the 'owner'
+    /// @param _update Update to be cancelled.
+    function cancel(Update memory _update) public {
```
yondonfu: This needs an `onlyOwner`.

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+
+    afterEach(async () => {
+        await fixture.tearDown()
+    })
+
+
+    const setUint256Tx = async (i, sender) => {
+        const txData = utils.arrayify(setUint256.contract.methods.setUint256(utils.bigNumberify(i)).encodeABI())
+        const dataHash = utils.solidityKeccak256(["address", "uint256", "bytes"], [setUint256.address, utils.bigNumberify("0"), txData])
+        const sig = await web3.eth.sign(dataHash, sender)
```
yondonfu: Doesn't seem like we need this anymore since updates don't need to be signed.

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+                        target: [setUint256.address],
+                        value: ["0"],
+                        data: [data]
+                    },
+                    constants.MaxUint256
+                )
+            )
+        })
+
+        it("stage emits an UpdateStaged event", async () => {
```
yondonfu, line 169: Let's also check that the contract storage mapping for updates is updated correctly?

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+                        target: [governor.address],
+                        value: ["0"],
+                        data: [txData]
+                    }
+                ),
+                "newOwner is zero"
+            )
+        })
+
+        it("updates ownership to a new owner", async () => {
```
yondonfu, line 77: Let's also check that the contract storage value for `owner` is updated correctly?

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+                        data: [data0, data1, data2]
+                    },
+                    "0",
+                    {from: accounts[1]}
+                ),
+                "unauthorized"
+            )
+        })
+
+        it("stage emites an UpdateStaged event", async () => {
```
yondonfu: "emites" -> "emits"

yondonfu: Let's also check that the contract storage mapping for updates is updated correctly?

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+            await truffleAssert.reverts(governor.execute(
+                {
+                    target: [setUint256.address],
+                    value: ["0"],
+                    data: [data]
+                }
+            ), "I should fail")
+        })
+
+        it("executes an update: delete the update and emit an UpdateExecuted event", async () => {
```
yondonfu, line 537: Should we also test for successful ETH transfer when the "value" in one of the txs included in an update is non-zero?

test/unit/Governor.js
```diff
@@ -0,0 +1,558 @@
...
+                        target: [setUint256.address, setUint256.address, setUint256.address],
+                        value: ["0", "0", "0"],
+                        data: [data0, data1, data2]
+                    }
+                ),
+                "delay for update not expired"
+            )
+        })
+
+        it("reverts when one of the remote calls in the batch fails", async () => {
```
yondonfu, line 626: Should we add a test when a remote call other than the first one reverts that checks that the revert reason corresponds with the correct remote call (and not just the first one)?

yondonfu (changes requested):

contracts/governance/Governor.sol
kyriediculous: If that convention requires us to define an intermediate variable `oldOwner` I don't see much purpose in adhering to it since the code in this function is executed atomically anyway.

contracts/governance/Governor.sol
kyriediculous: Woops, will add a test for it as well.

contracts/governance/Governor.sol
yondonfu: Ah I see. If emitting the event here avoids introducing an additional var then this looks fine to me.

test/unit/Governor.js
kyriediculous: We could but if this was written the wrong way it wouldn't compile, but we could add it for completeness.

yondonfu: I think we should add a test for ETH transfers if the contract is going to support ETH transfers as a part of updates.

test/unit/Governor.js
kyriediculous: I don't see much value here because 

1. execute already tests that `success == true` when the call doesn't fail and thus no revert is encountered
2. we will only ever get the revert reason for the first failing call

test/unit/Governor.js
kyriediculous: slightly annoying due to not being able to use `ethers.utils.solidityKeccak256`  . It only supports packed data but our data type doesn't support packed data. 

The least annoying solution would be to add a pure function to the governor contract (or a seperate one we just deploy for testing) that can calculate the hash for an update given the pre-image.

kyriediculous:
I believe 21161ac addresses all change requests

test/unit/Governor.js
yondonfu: Bump

test/unit/Governor.js
yondonfu: Fair enough. After some more reflection, I think I'm actually more interested in seeing cases where different the transactions in an update are for different contract and/or different functions - maybe something simple can be included in an integration test.

yondonfu (commented):
The implementation is looking solid! 

WDYT about adding a small integration test to simulate a few batch updates for the other Livepeer contracts via the Governor since that will be how it will be used in practice?

Also, the actual design of the Governor looks pretty good to me at this point so I'd be comfortable with moving it to Last Call whenever you would like to do so.

test/unit/Governor.js
kyriediculous: +1 for some integration test scenario's using our actual contracts

test/unit/Governor.js
kyriediculous: fixed in 9c11912

test/unit/Governor.js
kyriediculous: I added integration tests but I also added some changes for Testability as mentioned in this comment https://github.com/livepeer/protocol/pull/391#issuecomment-674029279 , I also removed the `whenSystemNotPaused` modifier from `Minter.DepositEth` like in #391. 

These Minter changes for testability are a separate commit and can be dropped if we merge #391 first (and the  revert reasons mentioned are added).

kyriediculous: I also decided to not  test for all `Minter.migrateToNewMinter`  revert paths since that is not the purpose of this test, I added two cases here and I think that's sufficient.

kyriediculous:
e4af94a fixes a typo and some descriptions to match revert reasons

test/integration/GovernorUpdate.js
```diff
@@ -0,0 +1,201 @@
+import truffleAssert from "truffle-assertions"
+import {utils} from "ethers"
+
+import {constants} from "../../utils/constants"
+import {contractId} from "../../utils/helpers"
+import {assert} from "chai"
```
yondonfu: Is this needed?

test/integration/GovernorUpdate.js
```diff
@@ -0,0 +1,201 @@
...
+
+    it("controller is now owned by the governor", async () => {
+        assert.equal(await controller.owner(), governor.address)
+    })
+
+    it("governor has the correct owner", async () => {
+        assert.equal(await governor.owner(), accounts[0])
+    })
+
+    describe("single param change", () => {
```
yondonfu, line 38: Is this missing a successful case?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,114 @@
...
+    /// @dev Forces the `stage/execute` path to be used to call functions with this modifier instead of directly.
+    modifier onlyThis() {
+        require(msg.sender == address(this), "unauthorized: msg.sender not Governor");
+        _;
+    }
+
+
+    /// @dev The Ownable constructor sets the original `owner` of the contract to the sender account.
+    constructor() public {
+        owner = msg.sender;
```
yondonfu, line 52: WDYT about emitting an `OwnershipTransferred` (from the null address to msg.sender) event here similar to what the OZ Ownable contract [constructor](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol#L29) does? It's just a single line to emit the event and the benefit is that it preserves the invariant that an event is emitted whenever ownership is updated (including in the constructor).

yondonfu (changes requested):

test/integration/GovernorUpdate.js
kyriediculous: No my vscode autofill keeps adding this , it's already injected by default.

contracts/governance/Governor.sol
kyriediculous: fixed in cf41a3c

test/integration/GovernorUpdate.js
kyriediculous: fixed in db643ac

test/integration/GovernorUpdate.js
kyriediculous: fixed in db643ac

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,115 @@
...
+    /// @dev Forces the `stage/execute` path to be used to call functions with this modifier instead of directly.
+    modifier onlyThis() {
+        require(msg.sender == address(this), "unauthorized: msg.sender not Governor");
+        _;
+    }
+
+
+    /// @dev The Ownable constructor sets the original `owner` of the contract to the sender account.
+    constructor() public {
+        emit OwnershipTransferred(owner, msg.sender);
```
yondonfu: Minor: Think its cheaper to do `emit OwnershipTransferred(address(0), ...)` (as is done in the OZ Ownable constructor) since it avoids reading `owner` and we already know the initial values of `owner` is `address(0)`.

yondonfu (changes requested):

contracts/governance/Governor.sol
kyriediculous: Good point, especially since istanbul.

Fixed in ac98605

kyriediculous:
522ffdb fixes the changes discussed after the security review 

- make `execute` payable instead of `stage`
- introduce a `nonce` to the update struct so that we can differentiate between updates , i.e. unique identifier. Not stored on-chain.

kyriediculous:
> The low-level functions call, delegatecall and staticcall return true as their first return value if the account called is non-existent, as part of the design of the EVM. Account existence must be checked prior to calling if needed.

Should we check the codesize of the target address ?

yondonfu:
> Should we check the codesize of the target address ?

I think its reasonable to just have the update submitter be responsible for ensuring that the contract exists via a check off-chain instead of adding that check in the contract since the update submitter is already responsible for ensuring that the rest of the update (i.e. tx data) is correct.

yondonfu:
Updates look good to me - let's rebase on top of `streamflow` since a few changes have been made there recently?

kyriediculous:
Done !

yondonfu (approved):

PR #385, opened 2020-07-15
ci: Block merges when there are fixup commits on the branch

yondonfu:
This PR adds a GH action to block merges when there are fixup commits on the branch.


kyriediculous (approved):
LGTM :+1:

PR #386, opened 2020-07-16
Fix Solidity unit test runner

yondonfu:
This PR contains a few updates that make running tests locally easier/a bit faster and also fixes an issue with the Solidity unit test runner.

fd7e690: Fix a check in `test.sh` that was preventing the script from starting ganache in the background
ab02f7d: Skip migrations when running contract unit tests because unit tests will handle contract deployment/initialization. Migrations will still be run for contract integration tests.
c4a40d3: Fix the Solidity unit test runner (we use a JS helper from another project to run unit tests written in Solidity due to some quirks with the coverage reporting tool). Previously, failed assertions were not being reported by the test runner.

### DIFF
--- a/migrations/1_initial_migration.js
+++ b/migrations/1_initial_migration.js
@@ -1,5 +1,9 @@
 const Migrations = artifacts.require("./Migrations.sol")
 
-module.exports = function(deployer) {
+module.exports = function(deployer, network) {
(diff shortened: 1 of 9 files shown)


kyriediculous (approved):
LGTM 👍

PR #389, opened 2020-07-29
fix SortedDoublyLL link when using unitTest network

kyriediculous:
#386 skips migrations when using the `unitTest` network, but it also skips linking `DoublySortedLL` to the `BondingManager`. 

This PR fixes the linking when using the `unitTest` network. 

To reproduce the issue check out #386 and run `./node_modules/.bin/truffle test test/unit/BondingManager.js --network=unitTest` 

The following error will be returned 

```
  1) Contract: BondingManager
       "before all" hook in "Contract: BondingManager":
     Error: BondingManager contains unresolved libraries. You must deploy and link the following libraries before you can deploy a new version of BondingManager: SortedDoublyLL
```

### DIFF
--- a/migrations/2_deploy_libraries.js
+++ b/migrations/2_deploy_libraries.js
@@ -2,10 +2,7 @@ const SortedDoublyLL = artifacts.require("SortedDoublyLL")
 const BondingManager = artifacts.require("BondingManager")
 
 module.exports = function(deployer, network) {
-    if (network === "unitTest") {
-        return
-    }
-
+    if (network  === "unitTest") return
     deployer.deploy(SortedDoublyLL)
     deployer.link(SortedDoublyLL, BondingManager)
 }
--- a/test/unit/BondingManager.js
+++ b/test/unit/BondingManager.js
@@ -6,6 +6,7 @@ import BN from "bn.js"
 import truffleAssert from "truffle-assertions"
 
 const BondingManager = artifacts.require("BondingManager")
+const LinkedList = artifacts.require("SortedDoublyLL")
 
 const {DelegatorStatus, TranscoderStatus} = constants
 
@@ -24,6 +25,8 @@ contract("BondingManager", accounts => {
         fixture = new Fixture(web3)
         await fixture.deploy()
 
+        const ll = await LinkedList.new()
+        BondingManager.link("SortedDoublyLL", ll.address)
         bondingManager = await fixture.deployAndRegister(BondingManager, "BondingManager", fixture.controller.address)
 
         await bondingManager.setUnbondingPeriod(UNBONDING_PERIOD)


migrations/2_deploy_libraries.js
```diff
@@ -1,11 +1,7 @@
 const SortedDoublyLL = artifacts.require("SortedDoublyLL")
 const BondingManager = artifacts.require("BondingManager")
 
-module.exports = function(deployer, network) {
-    if (network === "unitTest") {
```
yondonfu, line 5: Removing this check would result in the deployment of `SortedDoublyLL` before each unit test is run (i.e. when the `unitTest` network is used).

Could an alternative be to only deploy and link `SortedDoublyLL` in `test/unit/BondingManager.js` (and generally in any unit test that needs to deploy the BondingManager)?

i.e. In `test/unit/BondingManager.js`

```
before(async () => {
    ...
    const ll = await SortedDoublyLL.new()
    // According to the truffle-contract docs https://github.com/trufflesuite/truffle/tree/develop/packages/contract#mycontractlinkinstance
    // BondingManager.link(ll) should also work, but I encountered an error when I tried
    // Worth double checking if the above works as it would be a bit cleaner since we wouldn't need to also specify the library name
    // Otherwise, the below should be fine
    BondingManager.link("SortedDoublyLL", ll.address)
})
```

kyriediculous: fixed in 168247b

kyriediculous:
So the trufflecontract wrapper does have the notion of linking, but it's impossible to replace an existing library address which would be happening if we run the BondingManager.js tests on a network that isn't "unitTest" which is why CI fails. 

This is due to the fact that the placeholder in the bytecode is replaced by the address; but this can only happen once. 

And as far as I know the truffle wrapper around mocha doesn't give access to the network name used even though it's been requested a couple of times. 

I propose keeping the original solution of ignoring the network name in the library migrations file despite it deploying that contract before every test.

test/unit/BondingManager.js
```diff
@@ -24,6 +25,8 @@ contract("BondingManager", accounts => {
         fixture = new Fixture(web3)
         await fixture.deploy()
 
+        const ll = await LinkedList.new()
+        BondingManager.link("SortedDoublyLL", ll)
```
yondonfu: I think this needs to be:

```
BondingManager.link("SortedDoublyLL", ll.address)
```

See the [docs](https://github.com/trufflesuite/truffle/tree/develop/packages/contract#mycontractlinkname-address) for truffle-contract.

migrations/2_deploy_libraries.js
```diff
@@ -2,10 +2,7 @@ const SortedDoublyLL = artifacts.require("SortedDoublyLL")
 const BondingManager = artifacts.require("BondingManager")
 
 module.exports = function(deployer, network) {
-    if (network === "unitTest") {
```
yondonfu, line 5: Let's keep this i.e. the triple equal check?

yondonfu (commented):
> So the trufflecontract wrapper does have the notion of linking, but it's impossible to replace an existing library address which would be happening if we run the BondingManager.js tests on a network that isn't "unitTest" which is why CI fails.

Really?

I just ran the tests locally successfully (using `npm run test`) after updating the link command in `BondingManager.js` to:

```
BondingManager.link("SortedDoublyLL", ll.address)
```

test/unit/BondingManager.js
kyriediculous: ah yes you're right , shouldn't have done this in between other stuff. fixed now.

yondonfu (approved):
LGTM

PR #390, opened 2020-07-30
LIP-36: Cumulative earnings calculation

kyriediculous:
This PR implements cumulative earnings calculations as described in https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-36.md


test/unit/BondingManager.js
```diff
@@ -4,6 +4,7 @@ import {contractId, functionSig, functionEncodedABI} from "../../utils/helpers"
 import {constants} from "../../utils/constants"
 import BN from "bn.js"
 import truffleAssert from "truffle-assertions"
+import { assert } from "chai"
```
kyriediculous: vscode did this... will remove

kyriediculous:
Considerable changes compared to the original spec: 

1. If an orchestrator redeems ticket 1 with creationRound = N and then ticket 2 with creationRound = N - 1, then the fees for ticket 2 would be lost.

**solution** 
Always claim fees for the current round
https://github.com/livepeer/LIPs/issues/35#issuecomment-673659199

2.  If an orchestrator's cumulativeRewardFactor for round N - 1 is 0 and the orchestrator's lastRewardRound = currentRound then we are unable to initialize the cumulativeFeeFactor for round N since it depends on the last known cumulativeRewardFactor

**solution**
Retroactively calculate the woud be `rewardFactor` for round N-1 and store it on the earningsPool for round N-1 
https://github.com/livepeer/LIPs/issues/35#issuecomment-673659199

3. If an orchestrator calls reward for round N but has missed a reward call for round N-1 we can not correctly calculate rewards 

**solution** 
If `prevEarningsPool.cumulativeRewardFactor == 0` set it to the `cumulativeRewardFactor` of the `earningsPool` for the orchestrator's `lastRewardRound`

4. If the transcoder receives fees for round N But didn't receive fees for round N-1 then `prevEarningsPool.cumulativeFeeFactor` will not be set and we lose fees for all rounds prior to round N-1 

**solution**
Add a new state variable to the `transcoder` struct: `lastFeeRound` 
Set `lastFeeRound` when receiving fees 
If `prevEarningsPool.cumulativeFeeFactor` (for round N-1)  is 0 write the `cumulativeFeeFactor` for the `lastFeeRound` to that pool instead. 

5. If a transcoders last claimed in round N and now tries to claim for round N +1 through N + 1 + Y it will not include fees for round N + 1 

**solution**
When using the cumulative approach for fees and rewards let the `startEarningsPool` be that for round N 


6. Cumulative fee calculation will always returns fees since the inception of cumulative fee calculation regardless of prior fees being claimed and withdrawn or not, which effectively leads to double spending fees 

~**solution**~
~(1)
Track a new state variable on the `delegator` struct: `pastFees`
When withdrawing fees , add the fees being withdrawn to `pastFees` 
When calculating cumulative fees  substract `pastFees`~

~(2) Don't add `delegator.fees` to `cumulativeFees` when calculating fees for earnings when all rounds to claim for are following the cumulative calculation approach~

Update fee earnings calculation to 

`delegator.currentBondedAmount * (endCumulativeFeeFactor - startCumulativeFeeFactor) / startCumulativeRewardFactor`


______

* ~Currently the solution for problem 2 is breaking earnings claiming unit tests probably due to state writes that mess up the calculation~
--> Didn't properly scale the reward factor  (1 instead of PERC_DIVISOR) 

* ~Problem 6 is currently solved through additional state and additional code paths , but there might be a mathematical solution here as well.~

* ~failing LIP-36 transition test whereby we miss the pre-LIP 36 round fees in our calculation after claiming for both a pre-LIP and post LIP 36 round and then progressing rounds, receiving more fees and claiming again~

_______

**TODO**

- Bring Coverage up (2 untested branches in BondingManager and ~some untested codepaths for old methods in EarningsPoolV2~). 

- ~Move LIP_UPGRADE_ROUNDS to roundsmanager~ 

- Add a check to `pendingFees()` and `pendingStake()` to allow a smoother LIP-36 upgrade transition which doesn't force a user to either claim before or after the update. 

(basically add `if earningsPool.hasTranscoderRewardFeePool { break (from loop); }`

- ~Update spec~ & discourse post

- Add an integration test

kyriediculous:
There is a substantial change to the way `updateDelegatorWithEarnings` works. 

In this PR we re-use `pendingFees()` and `pendingStake()` to calculate the amount of fees and rewards owed to the delegator that is claiming earnings. 

This reusability of those two helpers shortens the code for `updateDelegatorWithEarnings` a fair bit. 
________ 

Prior to this PR the only significant difference between using `pendingFees` and `pendingStake` is that `EarningsPool.claimShare()` is called for each earnings pool that earnings are being claimed for in `updateDelegatorWithEarnings()`.

This function subtracts the claimed fees and rewards from the `feePool` and `rewardsPool` as well as subtract the delegators `stake` from `claimableStake` for the earningsPool.

____ 

IIUC correctly however there are no negative drawbacks to NOT doing these state updates as is in this PR. This does not affect earnings calculation. 

ie. consider the following scenario: 

400 in fees in earningsPool
2000 in claimable stake in earningsPool
delegator1 has 500 in stake 
delegator2 has 1500 in stake 

delegator 1 claims and receives 100 in fees ( 400 * (500/2000) ) 
-> remaining feePool = 300 , remaining claimableStake = 1500 

delegator 2 claims and receives 300 in fees ( 300 * (1500/1500) )
-> remaining feePool = 0 , remaining claimableStake = 0_

If we don't do these state updates however the result remains the same 

delegator1 receives 400* (500/2000) = 100 in fees
delegator2 receives 400 * (1500/2000) = 300 in fees 

____ 

Double spending is already avoided by setting a delegator's `lastClaimRound` , the only thing these state updates really accomplish IIUC is to give a historical view of what is left in an `EarningsPool` but this seems rather unnecessary. 

_HOWEVER_ as a result of not doing these state updates we will never have an earningsPool for which `earningsPool.hasClaimableShares()` evaluates to `false`. These in fact are the two branches in coverage that are currently not being hit.

yondonfu:
> IIUC correctly however there are no negative drawbacks to NOT doing these state updates as is in this PR.

The reason the state updates for `feePool`, `rewardsPool` and `claimableStake` were included in the old earnings claiming algo implementation is that they allow additional fees and/or rewards added to the earnings pool for a round after other delegators have already claimed from the earnings pool to be distributed amongst the remaining delegators that have not yet claimed from that earnings pool. For example, in the scenario you outlined, consider the case where more fees are added to the earnings pool after delegator 1 claims. So, I think those state updates should be kept (and if LIP-52 is implemented then those state updates wouldn't need to be executed in practice) so that changes to the old earnings claiming algo can be kept out of this PR even if the behavior of distributing additional rewards/fees to delegators that have not yet claimed is not a part of the new earnings claiming algo.

kyriediculous:
> additional fees and/or rewards added to the earnings pool for a round after other delegators have already claimed from the earnings pool to be distributed amongst the remaining delegators that have not yet claimed from that earnings pool

I don't think this is the right approach either way. This is a UX problem and I've already expressed my opinion that UX should not be handled in the contracts if it involves several state reads and writes. 

This could e.g. be handled by a notification in the explorer stating that you are trying to claim earnings up until a round your orchestrator has not called reward for and you will miss out on those earnings. 

For fees it could be something similar, don't claim up until `currentRound` but `currentRound - 1` unless one really needs all earnings to be claimed right now. 

I also don't think that it's fair to redistribute those fees/rewards to the other delegators, I think it would be fine for those fees/rewards to just be considered lost as well (again, UX problem). 
_____ 

> So, I think those state updates should be kept (and if LIP-52 is implemented then those state updates wouldn't need to be executed in practice) so that changes to the old earnings claiming algo can be kept out of this PR 

I disagree on this point 

The benefit of changing the approach here is that it allows for code readability and reusability with very minimal impact to the current behaviour. The only fees/rewards impacted would be those for the LIP-36 upgrade round _prior_ to the upgrade and _if_ and only if someone claims at the very beginning of that upgrade round, which is quite unlikely to happen. 

After LIP-36 is deployed fees and rewards for those past rounds can't be added either way, so that code really has no use, regardless of LIP-52 being used or not. 

Therefore I would be in favour of allowing for the reusability of `pendingStake()` and `pendingFees()` over an edge case for one round.

yondonfu:
> The benefit of changing the approach here is that it allows for code readability and reusability with very minimal impact to the current behaviour.

I understand. However, my concern is that reviewing these changes might become more challenging because now a third party is not just responsible for reviewing the correctness of the implementation of the new earnings claiming algo, but also the correctness of the changes to the old earnings claiming algo. Contrast this with the scenario where a third party reviewer would only need to be responsible for reviewing the correctness of the implementation of the new earnings claiming algo.

I'd be interested in understanding if there is an alternative way to achieve similar code readability and reusability benefits without expanding the scope of LIP-36 to also include changes to the old earnings claiming algo. Curious if you think that would be possible. I'll also keep this in mind as I jump into the actual code review at the beginning of next week.

kyriediculous:
Okay yeah that's fair enough.

I'll duplicate the code for now then and we can see this as an implementation choice TBD in the review process. 

I guess a middle road is to create helpers for just the cumulative calculations and re-use those.

kyriediculous:
As expected though, duplicating the code puts us over the bytecode limit so we NEED reusable helpers. 

>  but also the correctness of the changes to the old earnings claiming algo.

Since it's for a single round I actually feel like this is okay given the bytecode length hiccup as well. I also had to fight my way around stack depth before that. 

Reusing pendingFees and pendingStake becomes a more reasonable tradeoff the further I dig. 

_________

**UPDATE** 

I wrote two helpers , this puts us right under the bytecode length limit. 

```
build/contracts/BondingManager.json: 24549.5
```

The limit is 24577

Compared to reusing pendingFees and pendingStake however there is still a fair increase in bytecode length and I fear we will not be able to add the necessary logic for the snapshot claiming as is. 

With the PR as is (reusing pendingFees and pendingStake) , the bytecode length is under 24000

______

**UPDATE 2**

I managed to get the bytecode length down a bit more 

```
build/contracts/BondingManager.json: 24386.5
```

kyriediculous:
>  Contrast this with the scenario where a third party reviewer would only need to be responsible for reviewing the correctness of the implementation of the new earnings claiming algo.

After sleeping on it for the weekend I don't think we should let that drive our decision here. Highlighting the change should be quite sufficient in my opinion because it would only matter for the LIP-36  upgrade round itself. Any side effects are thus negligible.

kyriediculous:
For reference here is the bytecode length reusing `pendingFees` and `pendingStake`

`build/contracts/BondingManager.json: 23785.5`

It's significantly smaller and gives us way more wiggle room to add logic for i.e. LIP-52

contracts/bonding/BondingManager.sol
```diff
@@ -306,21 +309,63 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        // for cumulative fee calculation this would result in division by zero. 
+        if (currentRound > lastRewardRound) {
             earningsPool.setCommission(
                 t.rewardCut,
                 t.feeShare
             );
+            
+            earningsPool.setStake(t.earningsPoolPerRound[t.lastActiveStakeUpdateRound].totalStake);
+
+            t.activeCumulativeRewards = t.cumulativeRewards;
```
yondonfu: Is this storage update necessary? If `currentRound > lastRewardRound` then reward has not been called for the current round which means `t.cumulativeRewards` has not been updated with additional rewards in the current round yet and the rewards tracked by `t.cumulativeRewards` would actually all be considered active at this point. So, the value of `t.cumulativeRewards` could be used in all calculations that require the transcoder's active cumulative rewards. Something like the following might work:

```solidity
uint256 activeCumulativeRewards = 0;

if (currentRound > lastRewardRound) {
    // Other code
    ...

    activeCumulativeRewards = t.cumulativeRewards;
} else {
    // Other code
    ...

    activeCumulativeRewards = t.activeCumulativeRewards;
}
```

contracts/bonding/BondingManager.sol
```diff
@@ -306,21 +309,63 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
         }
 
+        // LIP-36: Add fees for the current round instead of '_round'
+        // https://github.com/livepeer/LIPs/issues/35#issuecomment-673659199
+        uint256 totalStake = earningsPool.totalStake;
+        if ( prevEarningsPool.cumulativeRewardFactor == 0 && lastRewardRound > roundsManager().LIPUpgradeRounds(36)) {
+            // if transcoder didn't call reward for 'currentRound - 1' nor 'currentRound', use the cumulativeRewardFactor for 'lastRewardRound'
+            if (lastRewardRound < currentRound) {
+                prevEarningsPool.cumulativeRewardFactor = t.earningsPoolPerRound[lastRewardRound].cumulativeRewardFactor;
```
yondonfu: Could we avoid some of these `prevEarningsPool` storage updates (i.e. cumulativeRewardFactor on this line, cumulativeRewardFactor on L348 and cumulativeFeeFactor on L355) if we declared the `_prevEarningsPool` param for `addToFeePool()` as a memory variable instead of a storage pointer? We can still pass the correct required values via the `EarningsPool.Data` struct into `addToFeePool()`, but these values do not necessarily need to be persisted to storage.

The signature for `addToFeePool()` (the signature for `addToRewardPool()` would look similar) would be:

```solidity
function addToFeePool(
    EarningsPool.Data storage earningsPool,
    EarningsPool.Data memory _prevEarningsPool,
    uint256 _fees
);
```

A sketch of what initialization of the in-memory struct might look like:

```solidity
...

EarningsPool.Data storage earningsPool = t.earningsPoolPerRound[currentRound];

uint256 prevRewardRound = lastRewardRound;
if (prevRewardRound == currentRound) {
    prevRewardRound = currentRound.sub(1);
}
uint256 prevFeeRound = lastFeeRound;
if (prevFeeFound == currentRound) {
    prevFeeRound = currentRound.sub(1);
}

// Create an earnings pool in-memory
EarningsPool.Data memory prevEarningsPool = EarningsPool.Data({
    cumulativeRewardFactor: t.earningsPoolPerRound[prevRewardRound].cumulativeRewardFactor,
    cumulativeFeeFactor: t.earningsPoolPerRound[prevFeeRound].cumulativeFeeFactor
});

if (prevEarningsPool.cumulativeRewardFactor == 0 && prevRewardRound == currentRound) {
    // Calculate the cumulativeRewardFactor for currentRound - 1 and update
    // prevEarningsPool.cumulativeRewardFactor
}

// Fees calculation
...

earningsPool.addToFeePool(prevEarningsPool, fees);
```

contracts/bonding/BondingManager.sol
```diff
@@ -306,21 +309,63 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        uint256 delegatorsFees = MathUtils.percOf(_fees, earningsPool.transcoderFeeShare);
+        uint256 transcoderCommissionFees = _fees.sub(delegatorsFees);
+        uint256 transcoderRewardStakeFees = MathUtils.percOf(delegatorsFees, t.activeCumulativeRewards, totalStake);
+
+        t.cumulativeFees = t.cumulativeFees.add(transcoderRewardStakeFees).add(transcoderCommissionFees);
         // Add fees to fee pool
-        earningsPool.addToFeePool(_fees);
+        earningsPool.addToFeePool(prevEarningsPool, delegatorsFees);
+
+        if (lastFeeRound < _round) {
```
yondonfu: IIUC we're ignoring the `_round` param in the rest of this function and instead updating the earnings pool using the `_fees` param for the current round. But, it could still be the case that `_round < currentRound` since the TicketBroker will continue to call `updateTranscoderWithFees()` using a ticket's creation round. In this case, `lastFeeRound` could start at N, but then be updated to M s.t. M < N. I think we should always be setting `lastFeeRound` to the current round to avoid this?

contracts/rounds/RoundsManager.sol
```diff
@@ -33,6 +33,11 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
     // Mapping round number => block hash for the round
     mapping (uint256 => bytes32) internal _blockHashForRound;
 
+    // LIP Upgrade Rounds 
+    // These can be used as conditionals to ensure backwards compatibility or skip such backwards compatibility logic
```
yondonfu: "as conditionals" -> "in conditionals"

contracts/rounds/RoundsManager.sol
```diff
@@ -101,6 +106,17 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
         emit NewRound(currRound, roundBlockHash);
     }
 
+    /**
+    * @notice setLIPUpgradeRound sets the round an LIP upgrade would become active. 
+    * @dev if no '_round' is provided '_round' will be set to 'currentRound + 1'
+    * @param _LIP the LIP number.
+    * @param _round (optional) the round in which the LIP becomes active
```
yondonfu, line 112: As mentioned in [this comment](https://github.com/livepeer/LIPs/pull/51#discussion_r478669005), I'm not sure how useful allowing the `_round` param to be optional is. Open to hearing any particular use cases you had in mind.

contracts/bonding/BondingManager.sol
```diff
@@ -306,21 +309,63 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            if (lastRewardRound < currentRound) {
+                prevEarningsPool.cumulativeRewardFactor = t.earningsPoolPerRound[lastRewardRound].cumulativeRewardFactor;
+            } else {
+                // if transcoder called reward for 'currentRound' but not for 'currentRound - 1' (missed reward call)
+                // retroactively calculate what its cumulativeRewardFactor would have been for 'currentRound - 1' (cfr. previous lastRewardRound for transcoder)  
+                // based on rewards for currentRound
+                uint256 rewards = MathUtils.percOf(minter().currentMintableTokens(), totalStake, currentRoundTotalActiveStake);
+                uint256 transcoderCommissionRewards = MathUtils.percOf(rewards, earningsPool.transcoderRewardCut);
+                uint256 delegatorRewards = rewards.sub(transcoderCommissionRewards);
+                prevEarningsPool.cumulativeRewardFactor = MathUtils.percOf(earningsPool.cumulativeRewardFactor, MathUtils.percOf(delegatorRewards, totalStake)).add(MathUtils.percPoints(1,1));
```
yondonfu: I don't think this line result is in the correct calculation. In particular, consider this portion of the line:

```solidity
MathUtils.percOf(delegatorRewards, totalStake)
```

The above will calculate `delegatorRewards * totalStake / PERC_DIVISOR` and treats `totalStake` as a percentage value out of `PERC_DIVISOR`.

I believe the desired calculation is actually:

`cumulativeRewardFactor / (1 + delegatorRewards / totalStake)`

which can be expressed as

`cumulativeRewardFactor * (1 / (1 + delegatorRewards / totalStake))`

which can be written in Solidity as

```
MathUtils.percOf(
    earningsPool.cumulativeRewardFactor,
    MathUtils.percPoints(1, MathUtils.percPoints(1, 1).add(MathUtils.percPoints(delegatorsRewards, totalStake)))
)
```

Note that the the second param of the outer `MathUtils.percOf()` call is expressed as a value scaled by `PERC_DIVISOR`.

contracts/bonding/BondingManager.sol
```diff
@@ -676,63 +723,89 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        uint256 lastRewardRound = transcoder.lastRewardRound;
+        if (lastRewardRound < startRound) {
+            return currentBondedAmount;
+        }
+
+        bool isTranscoder = _delegator == del.delegateAddress;
+
+        uint256 LIP_36_ROUND = roundsManager().LIPUpgradeRounds(36);
+        while (startRound <= _endRound && startRound <= LIP_36_ROUND ) {
+            EarningsPool.Data storage earningsPool = transcoders[del.delegateAddress].earningsPoolPerRound[startRound];
```
yondonfu: `transcoders[del.delegateAddress]` can be shortened to `transcoder` since we already declare that storage pointer on L736.

contracts/bonding/BondingManager.sol
```diff
@@ -676,63 +723,89 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        for (uint256 i = del.lastClaimRound + 1; i <= _endRound; i++) {
-            EarningsPool.Data storage earningsPool = transcoders[del.delegateAddress].earningsPoolPerRound[i];
+        uint256 lastFeeRound = transcoder.lastFeeRound;
+        if (lastFeeRound < startRound) {
+            return currentFees;
+        }
+
+        uint256 LIP_36_ROUND = roundsManager().LIPUpgradeRounds(36);
+        while (startRound <= _endRound && startRound <= LIP_36_ROUND ) {
+            EarningsPool.Data storage earningsPool = transcoders[del.delegateAddress].earningsPoolPerRound[startRound];
```
yondonfu: `transcoders[del.delegateAddress]` can be shortened to `transcoder` since that storage pointer is already declared on L779.

contracts/bonding/BondingManager.sol
```diff
@@ -46,6 +46,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         uint256 lastActiveStakeUpdateRound;                             // Round for which the stake was last updated while the transcoder is active
         uint256 activationRound;                                        // Round in which the transcoder became active - 0 if inactive
         uint256 deactivationRound;                                      // Round in which the transcoder will become inactive
+        uint256 activeCumulativeRewards;                                // The orchestrator's cumulative rewards that are active in the current round
```
yondonfu: Since we're continuing to use the term "transcoder" in the contracts let's remain consistent and use transcoder instead of orchestrator in these comments.

contracts/bonding/BondingManager.sol
```diff
@@ -306,21 +309,63 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        // if transcoder hasn't called 'reward()' for '_round' its 'transcoderFeeShare', 'transcoderRewardCut' and 'totalStake'
+        // on the 'EarningsPoolV2' for '_round' would not be initialized and the fee distribution wouldn't happen as expected
+        // for cumulative fee calculation this would result in division by zero. 
+        if (currentRound > lastRewardRound) {
             earningsPool.setCommission(
                 t.rewardCut,
                 t.feeShare
             );
+            
+            earningsPool.setStake(t.earningsPoolPerRound[t.lastActiveStakeUpdateRound].totalStake);
```
yondonfu: When `lastActiveStakeUpdateRound == currentRound` (i.e. if reward was called in currentRound - 1, but not currentRound) is the compiler smart enough to make this a no-op under the hood since `earningsPool.totalStake` is equal to `t.earningsPooPerRound[t.lastActiveStakeUpdateRound].totalStake`?

contracts/bonding/BondingManager.sol
```diff
@@ -676,63 +723,89 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
                 // Calculate and add reward pool share from this round
                 currentBondedAmount = currentBondedAmount.add(earningsPool.rewardPoolShare(currentBondedAmount, isTranscoder));
             }
+            startRound = startRound.add(1);
         }
 
-        return currentBondedAmount;
+        if (_endRound <= LIP_36_ROUND) { return currentBondedAmount; }
+
+        return cumulativeRewards(transcoder, currentBondedAmount, startRound.sub(1), _endRound, isTranscoder);
```
yondonfu: I think we need to add this to `currentBondedAmount` since `cumulativeRewards()` will just return the rewards and not include the current stake?

contracts/bonding/BondingManager.sol
```diff
@@ -1110,28 +1245,66 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            // If it is not initialised set it's cumulative earnings factors (fees and rewards)
+            // This serves two purposes:
+            // - Initialise the correct values for the current claimEarnings 'epoch' to calculate cumulative earnings
+            // - Initialise the correct startEarningsPool for the next claimEarnings 'epoch' to calculate cumulative earnings
+            EarningsPool.Data storage endEarningsPool = transcoder.earningsPoolPerRound[_endRound];
+            if (endEarningsPool.cumulativeRewardFactor == 0) {
+                endEarningsPool.cumulativeRewardFactor = transcoder.earningsPoolPerRound[transcoder.lastRewardRound].cumulativeRewardFactor;
+            }
+            if (endEarningsPool.cumulativeFeeFactor == 0) {
+                endEarningsPool.cumulativeFeeFactor == transcoder.earningsPoolPerRound[transcoder.lastFeeRound].cumulativeFeeFactor;
```
yondonfu: == -> =

contracts/bonding/BondingManager.sol
```diff
@@ -1110,28 +1245,66 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        }
 
+            // Check whether the endEarningsPool is initialised
+            // If it is not initialised set it's cumulative earnings factors (fees and rewards)
+            // This serves two purposes:
+            // - Initialise the correct values for the current claimEarnings 'epoch' to calculate cumulative earnings
+            // - Initialise the correct startEarningsPool for the next claimEarnings 'epoch' to calculate cumulative earnings
+            EarningsPool.Data storage endEarningsPool = transcoder.earningsPoolPerRound[_endRound];
+            if (endEarningsPool.cumulativeRewardFactor == 0) {
+                endEarningsPool.cumulativeRewardFactor = transcoder.earningsPoolPerRound[transcoder.lastRewardRound].cumulativeRewardFactor;
```
yondonfu: Do we need these storage updates if we can always just read the appropriate values for a past earnings pool as mentioned in my earlier comment for `updateTranscoderWithFees()`?

contracts/bonding/libraries/EarningsPoolV2.sol
```diff
@@ -0,0 +1,306 @@
...
+            );
+    }
+
+    /**
+     * @dev Add rewards to the earnings pool
+     * @param earningsPool Storage pointer to EarningsPool struct
+     * @param _rewards Amount of rewards to add
+     */
+    function addToRewardPool(EarningsPool.Data storage earningsPool, EarningsPool.Data storage _prevEarningsPool, uint256 _rewards) internal {
+        uint256 prevCumulativeRewardFactor = _prevEarningsPool.cumulativeRewardFactor != 0 ? _prevEarningsPool.cumulativeRewardFactor : MathUtils.percPoints(1,1);
```
yondonfu: When this is equal to `MathUtils.percPoints(1, 1)`, we would be adding a value scaled by `PERC_DIVISOR` with a value that is not scaled by `PERC_DIVISOR` on L96 (i.e. `MathUtils.percOf(prevCumulativeRewardFactor, ...)`). Wouldn't this lead to a problem? Shouldn't we be making sure that it is the case that either all terms in the calculation are scaled by `PERC_DIVISOR` or none of them are?

contracts/bonding/libraries/EarningsPoolV2.sol
```diff
@@ -0,0 +1,306 @@
...
+                MathUtils.percOf(prevCumulativeRewardFactor, _fees, earningsPool.totalStake)
+            );
+    }
+
+    /**
+     * @dev Add rewards to the earnings pool
+     * @param earningsPool Storage pointer to EarningsPool struct
+     * @param _rewards Amount of rewards to add
+     */
+    function addToRewardPool(EarningsPool.Data storage earningsPool, EarningsPool.Data storage _prevEarningsPool, uint256 _rewards) internal {
```
yondonfu: We may be able to reduce the diff of this PR and more cleanly separate the LIP-36 code by doing something like the following:

- Keep the existing `EarningsPool.sol` file
    - Add the new fields to the `EarningsPool` struct
    - Remove the `addToRewardPool()` and `addToFeePool()` functions from the file
- Create a new file `EarningsPoolLIP36.sol` which would look like:

```solidity
import "./EarningsPool.sol";

library EarningsPoolLIP36 {
    function addToRewardPool(EarningsPool.Data storage earningsPool, ...) { ... }
    function addToFeePool(EarningsPool.Data storage earningsPool, ...) { ... }
}
```

Then, using the new library in the BondingManager would look something like:

```solidity
import "./libraries/EarningsPool.sol";
import "./libraries/EarningsPoolLIP36.sol";

contract BondingManager {
    ...
    using EarningsPool for EarningsPool.Data;
    using EarningsPoolLIP36 for EarningsPool.Data;
}
```

WDYT?

contracts/bonding/BondingManager.sol
```diff
@@ -1110,28 +1245,66 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-            // across multiple transactions
-            require(_endRound.sub(_lastClaimRound) <= maxEarningsClaimsRounds, "too many rounds to claim through");
+            bool isTranscoder = _delegator == del.delegateAddress;
+
+            Transcoder storage transcoder = transcoders[del.delegateAddress];
 
-            for (uint256 i = startRound; i <= _endRound; i++) {
-                EarningsPool.Data storage earningsPool = transcoders[del.delegateAddress].earningsPoolPerRound[i];
+            // 'pendingFees()' and 'pendingStake()' are O(1) operations after LIP-36
+            if (startRound <= LIP_36_ROUND) {
```
yondonfu: Do we need this outer conditional if the inner loop will exit immediately when this is not the case since it would be the case that `startRound > endLoopRound`?

yondonfu (changes requested):

yondonfu:
> Compared to reusing pendingFees and pendingStake however there is still a fair increase in bytecode length and I fear we will not be able to add the necessary logic for the snapshot claiming as is.

I'm open to considering this change if necessary, although I'm interested in seeing where the bytecode size of the BondingManager stands after the requested changes are addressed first. Also, if we do make that change, using `pendingFees()` and `pendingStake()` as-is in `updateDelegatorWithEarnings()` will result in looping from startRound through min(endRound, LIP_36_ROUND) twice so there would likely need to be some refactoring to avoid that.

contracts/bonding/BondingManager.sol
kyriediculous: I tried this in a unit test, this doesn't seem quite correct either ?

yondonfu: Not sure if this is the cause, but I also noticed this line:

```solidity
uint256 rewards = MathUtils.percOf(minter().currentMintableTokens(), totalStake, currentRoundTotalActiveStake);
```

`currentMintableTokens()` will decrease whenever reward is called (the rewards minted is subtracted). I think the value we want to use here is the total mintable tokens at the beginning of the round which can be expressed as:

```solidity
minter().currentMintableTokens().add(minter().currentMintedTokens())
```

which adds the remaining mintable tokens in the round with the tokens already minted in the round.

kyriediculous: We should add that definitely but it's unrelated to the `claimEarnings` tests in the BondingManager unit tests.

contracts/bonding/BondingManager.sol
kyriediculous: yes because we 

1. need a scoped block to avoid stack depth
2. otherwise the require line that checks max claim rounds could underflow.

contracts/bonding/libraries/EarningsPoolV2.sol
kyriediculous: IIUC the factors are scaled by `PERC_DIVISOR` (otherwise they would be floats) so if it is not initialised we need to initialise it with 1*`PERC_DIVISOR` ; i.e. `MathUtils.percPoints(1,1)`

contracts/bonding/BondingManager.sol
kyriediculous: The formula you portray seems correct but I don't think the solidity 'translation' is correct, could you confirm? 

It yields wildly inaccurate results for previously passing unit tests.

contracts/bonding/BondingManager.sol
kyriediculous: Yeah it should be smart enough 

consider the following contract: 

```
pragma solidity ^0.5.11;
contract A {
    uint256 public a = 5; 
    
    function setA(uint256 _a) public { a = _a; }
}
```

Setting `a` to `5` through the setter is an execution costs of ~500 gas 

Setting `a` to `10` results in an execution costs of ~5500 gas

contracts/bonding/BondingManager.sol
kyriediculous: Yeah that's true and the compiler is smart enough to no-op on equal values (as mentioned somewhere else)

contracts/bonding/BondingManager.sol
kyriediculous: Only if there are pre-LIP36 rewards. Otherwise the rewardFactor will scale the currentBondedAmount up to what the total stake should be IIUC

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 40574ad

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 40574ad

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 40574ad

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 97980f7

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 97980f7

But I had to temporarily comment `slashTranscoder` and bump gas limit for ganache from 8M to 10M to get it to deploy.

contracts/bonding/BondingManager.sol
kyriediculous: I think we are also missing the `transcoderRewardStakeRewards` here to subtract from `delegatorsFees` , no ?

contracts/bonding/BondingManager.sol
yondonfu: This storage update code looks unchanged?

contracts/bonding/BondingManager.sol
yondonfu: > The formula you portray seems correct but I don't think the solidity 'translation' is correct, could you confirm?

Ah oof not sure what I was thinking there...

I think the following should be correct:

```solidity
MathUtils.percOf(
    earningsPool.cumulativeRewardFactor,
    totalStake,
    delegatorRewards.add(totalStake)
);
```

This should be equivalent to:

```
cumulativeRewardFactor * (1 / (1 + delegatorRewards / totalStake)) = ...
cumulativeRewardFactor * (1 / (delegatorRewards + totalStake / totalStake)) = ...
cumulativeRewardFactor * (totalStake / delegatorRewards + totalStake)
```

> I think we are also missing the transcoderRewardStakeRewards here to subtract from delegatorsFees , no ?

I don't think `transcoderRewardStakeRewards` is relevant here because those aren't subtracted from `delegatorRewards` in `updateTranscoderWithRewards()`. `transcoderRewardStakeRewards` represents the rewards earned by the cumulative rewards that the transcoder earned based on the portion of its self-delegated stake that comes from its `rewardCut`. The rest of the transcoder's self-delegated stake can earn a portion of `delegatorRewards`.

> but it's unrelated to the claimEarnings tests in the BondingManager unit tests.

The failing claimEarnings tests seem to be due to the actual values for calculated rewards and fees being slightly off from expected values for rewards and/or fees. For example, the acceptable delta between expected and actual rewards/fees in one of the tests is 2, but the actual delta is 4. I'm not sure if this is ok or not yet - will have to look into it more. It might be the case that the difference is the result of using a different method to calculate the expected result than the method used to calculate the actual result.

I also noticed that one of the pendingFees tests was failing [here](https://app.circleci.com/pipelines/github/livepeer/protocol/89/workflows/a1a9bbae-4ce8-41e0-9ce4-43f6cef4f2e2/jobs/1636). The failing test case:

```
5) Contract: BondingManager
       pendingFees
         should return pending fees when transcoder hasn't called reward for the previous round but has for the current round:

      AssertionError: expected '382' to equal '411'
      + expected - actual

      -382
      +411
      
      at _callee209$ (test/unit/BondingManager.js:7077:50)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:65:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:303:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:117:21)
      at step (test/unit/BondingManager.js:25:191)
      at /home/circleci/protocol/test/unit/BondingManager.js:25:361
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

I was able to fix that failing test case locally with the following changes:

```diff
diff --git a/contracts/bonding/BondingManager.sol b/contracts/bonding/BondingManager.sol
index 8273ba6..a11a3c7 100644
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -349,7 +349,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // LIP-36: Add fees for the current round instead of '_round'
         // https://github.com/livepeer/LIPs/issues/35#issuecomment-673659199
         uint256 totalStake = earningsPool.totalStake;
-        if (prevEarningsPool.cumulativeRewardFactor == 0 && prevRewardRound == currentRound && prevRewardRound > roundsManager().LIPUpgradeRounds(36)) {
+        if (prevEarningsPool.cumulativeRewardFactor == 0 && lastRewardRound == currentRound) {
             // if transcoder called reward for 'currentRound' but not for 'currentRound - 1' (missed reward call)
             // retroactively calculate what its cumulativeRewardFactor would have been for 'currentRound - 1' (cfr. previous lastRewardRound for transcoder)
             // based on rewards for currentRound
@@ -358,7 +358,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             uint256 delegatorRewards = rewards.sub(transcoderCommissionRewards);
             prevEarningsPool.cumulativeRewardFactor = MathUtils.percOf(
                 earningsPool.cumulativeRewardFactor,
-                MathUtils.percPoints(1, MathUtils.percPoints(1, 1).add(MathUtils.percPoints(delegatorRewards, totalStake)))
+                totalStake,
+                delegatorRewards.add(totalStake)
             );
         }

diff --git a/test/unit/BondingManager.js b/test/unit/BondingManager.js
index afb7483..8db67e7 100644
--- a/test/unit/BondingManager.js
+++ b/test/unit/BondingManager.js
@@ -2597,6 +2597,9 @@ contract("BondingManager", accounts => {
                 (pendingFees0 + pendingFees1 + pendingFees2).toString()
             )

+            await fixture.minter.setMockUint256(functionSig("currentMintableTokens()"), 0)
+            await fixture.minter.setMockUint256(functionSig("currentMintedTokens()"), 1000)
+            await fixture.roundsManager.execute(bondingManager.address, functionSig("setCurrentRoundTotalActiveStake()"))
             await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound + 4)
             await bondingManager.reward({from: transcoder})
```

Notes on the above diff:

- The conditional for determining whether we need to retroactively calculate the cumulativeRewardFactor for the previous round is updated
    - `prevRewardRound == currentRound` -> `lastRewardRound == currentRound`
        - This is necessary because otherwise `prevRewardRound = currentRound - 1` so the previous check would always be false
    - The LIP upgrade round check is removed as well because I don't think it's necessary? If the previous round was not after the LIP upgrade round then I think the cumulativeRewardFactor for the previous round would just evaluate to `MathUtils(1, 1)` which should be fine...worth double checking this though
- The formula for retroactively calculating the cumulativeRewardFactor in the previous round is updated
- The test is updated with appropriate mock statements. Without those mock statements, incorrect values would be used for the retroactive calculation

contracts/bonding/BondingManager.sol
kyriediculous: Oh sorry my bad.

These storage updates are necessary because they are not for a previous earnings pool but the start earnings pool when claiming. 

Essentially we copy over the values from the `lastRewardRound` and `lastFeeRound` to the earnings pool for the `lastClaimRound` because `lastRewardRound` and `lastFeeRound` is likely subject to change down the line so reading from those would yield different values.

contracts/rounds/RoundsManager.sol
kyriediculous: fixed in c4e3b96ec6a2f864bd8acd8ebe23203e3a76a815, will update the spec as well

contracts/bonding/BondingManager.sol
yondonfu: Addressed in 5c4b93e

contracts/bonding/BondingManager.sol
yondonfu: Ah yes right.

contracts/bonding/libraries/EarningsPoolV2.sol
yondonfu: Addressed in dff6dd2

yondonfu (approved):
LGTM.

Coverage is down a bit because there is a code path that cannot be triggered with the new version of the BondingManager, but can be triggered on mainnet since the old BondingManager version will contain the relevant state to trigger that code path.

Issue #387, opened 2020-07-20
Implement LIP-36


yondonfu:
[Spec](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-36.md)

yondonfu:
Closed by #390

PR #391, opened 2020-07-31
Implement LIP-40

yondonfu:
This PR implements [LIP-40](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-40.md).

TODO:
- [x] Add an integration test for the Minter upgrade as described in the deployment procedure for LIP-40

### DIFF
--- a/contracts/libraries/MathUtilsV2.sol
+++ b/contracts/libraries/MathUtilsV2.sol
@@ -1,0 +1,47 @@
+pragma solidity ^0.5.11;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+library MathUtils {
+    using SafeMath for uint256;
+
+    // Divisor used for representing percentages
+    uint256 public constant PERC_DIVISOR = 1000000000;
+
+    /**
+     * @dev Returns whether an amount is a valid percentage out of PERC_DIVISOR
+     * @param _amount Amount that is supposed to be a percentage
+     */
+    function validPerc(uint256 _amount) internal pure returns (bool) {
+        return _amount <= PERC_DIVISOR;
+    }
+
+    /**
+     * @dev Compute percentage of a value with the percentage represented by a fraction
+     * @param _amount Amount to take the percentage of
+     * @param _fracNum Numerator of fraction representing the percentage
+     * @param _fracDenom Denominator of fraction representing the percentage
+     */
+    function percOf(uint256 _amount, uint256 _fracNum, uint256 _fracDenom) internal pure returns (uint256) {
+        return _amount.mul(percPoints(_fracNum, _fracDenom)).div(PERC_DIVISOR);
+    }
+
+    /**
+     * @dev Compute percentage of a value with the percentage represented by a fraction over PERC_DIVISOR
+     * @param _amount Amount to take the percentage of
+     * @param _fracNum Numerator of fraction representing the percentage with PERC_DIVISOR as the denominator
+     */
+    function percOf(uint256 _amount, uint256 _fracNum) internal pure returns (uint256) {
+        return _amount.mul(_fracNum).div(PERC_DIVISOR);
+    }
+
+    /**
+     * @dev Compute percentage representation of a fraction
+     * @param _fracNum Numerator of fraction represeting the percentage
+     * @param _fracDenom Denominator of fraction represeting the percentage
+     */
+    function percPoints(uint256 _fracNum, uint256 _fracDenom) internal pure returns (uint256) {
+        return _fracNum.mul(PERC_DIVISOR).div(_fracDenom);
+    }
+}
--- a/contracts/token/Minter.sol
+++ b/contracts/token/Minter.sol
@@ -5,7 +5,7 @@ import "./IMinter.sol";
 import "./ILivepeerToken.sol";
 import "../rounds/IRoundsManager.sol";
 import "../bonding/IBondingManager.sol";
-import "../libraries/MathUtils.sol";
+import "../libraries/MathUtilsV2.sol";
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
--- a/contracts/token/Minter.sol
+++ b/contracts/token/Minter.sol
@@ -31,25 +31,31 @@ contract Minter is Manager, IMinter {
 
     // Checks if caller is BondingManager
     modifier onlyBondingManager() {
-        require(msg.sender == controller.getContract(keccak256("BondingManager")));
+        require(msg.sender == controller.getContract(keccak256("BondingManager")), "msg.sender not BondingManager");
         _;
     }
 
     // Checks if caller is RoundsManager
     modifier onlyRoundsManager() {
-        require(msg.sender == controller.getContract(keccak256("RoundsManager")));
+        require(msg.sender == controller.getContract(keccak256("RoundsManager")), "msg.sender not RoundsManager");
         _;
     }
 
     // Checks if caller is either BondingManager or JobsManager
     modifier onlyBondingManagerOrJobsManager() {
-        require(msg.sender == controller.getContract(keccak256("BondingManager")) || msg.sender == controller.getContract(keccak256("JobsManager")));
+        require(
+            msg.sender == controller.getContract(keccak256("BondingManager")) || msg.sender == controller.getContract(keccak256("JobsManager")),
+            "msg.sender not BondingManager or JobsManager"
+        );
         _;
     }
 
     // Checks if caller is either the currently registered Minter or JobsManager
     modifier onlyMinterOrJobsManager() {
-        require(msg.sender == controller.getContract(keccak256("Minter")) || msg.sender == controller.getContract(keccak256("JobsManager")));
+        require(
+            msg.sender == controller.getContract(keccak256("Minter")) || msg.sender == controller.getContract(keccak256("JobsManager")),
+            "msg.sender not Minter or JobsManager"
+        );
         _;
     }
 
@@ -61,11 +67,11 @@ contract Minter is Manager, IMinter {
      */
     constructor(address _controller, uint256 _inflation, uint256 _inflationChange, uint256 _targetBondingRate) public Manager(_controller) {
         // Inflation must be valid percentage
-        require(MathUtils.validPerc(_inflation));
+        require(MathUtils.validPerc(_inflation), "_inflation is invalid percentage");
         // Inflation change must be valid percentage
-        require(MathUtils.validPerc(_inflationChange));
+        require(MathUtils.validPerc(_inflationChange), "_inflationChange is invalid percentage");
         // Target bonding rate must be valid percentage
-        require(MathUtils.validPerc(_targetBondingRate));
+        require(MathUtils.validPerc(_targetBondingRate), "_targetBondingRate is invalid percentage");
 
         inflation = _inflation;
         inflationChange = _inflationChange;
@@ -78,7 +84,7 @@ contract Minter is Manager, IMinter {
      */
     function setTargetBondingRate(uint256 _targetBondingRate) external onlyControllerOwner {
         // Must be valid percentage
-        require(MathUtils.validPerc(_targetBondingRate));
+        require(MathUtils.validPerc(_targetBondingRate), "_targetBondingRate is invalid percentage");
 
         targetBondingRate = _targetBondingRate;
 
@@ -91,7 +97,7 @@ contract Minter is Manager, IMinter {
      */
     function setInflationChange(uint256 _inflationChange) external onlyControllerOwner {
         // Must be valid percentage
-        require(MathUtils.validPerc(_inflationChange));
+        require(MathUtils.validPerc(_inflationChange), "_inflationChange is invalid percentage");
 
         inflationChange = _inflationChange;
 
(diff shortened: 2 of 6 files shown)


contracts/libraries/MathUtilsV2.sol
```diff
@@ -0,0 +1,47 @@
+pragma solidity ^0.5.11;
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+library MathUtils {
```
kyriediculous, line 6: Curious as to why the decision was made to create a V2 file rather than adjusting the existing MathUtils contract. 

No state is affected by this change as these are all pure functions. Existing state can just as well use this contract and the outcome of calculations will be equal, no ? 

Is it purely for historical reference of code changes to the protocol?

yondonfu: LIP-40 only updates the Minter contract to use a new MathUtils library and to accomplish this the Minter is updated to import `MathUtilsV2.sol`. Other contracts such as the BondingManager will continue to use the old MathUtils library by continuing to import `MathUtils.sol`.

kyriediculous: I see. I feel like the BondingManager/EarningsPool could also benefit from more precision post LIP-36 but that's out of scope for this PR then.

contracts/libraries/MathUtilsV2.sol
```diff
@@ -0,0 +1,47 @@
...
+    using SafeMath for uint256;
+
+    // Divisor used for representing percentages
+    uint256 public constant PERC_DIVISOR = 1000000000;
```
kyriediculous, line 10: If we were to create a constant getter for this we could use that getter in the solidity unit tests instead of hardcoding the value but this is quite a trivial change so leaving that up to you.

e.g. 

```
function percDivisor() view returns (uint256) { return PERC_DIVISOR; }
```

```
    function test_validPerc() public {
        Assert.equal(MathUtils.validPerc(50), true, "50 should be a valid percentage");
        Assert.equal(MathUtils.validPerc(0), true, "0 should be a valid percentage");
        Assert.equal(MathUtils.validPerc(MathUtils.percDivisor()), true, "the max should be a valid percentage");
        Assert.equal(MathUtils.validPerc(MathUtils.percDivisor()+1), false, "1 more than the max should not be valid percentage");
    }
```

kyriediculous:
Deleted my previous comment about creating a getter for the constant PERC_DIVISOR and then using that getter in the solidity unit tests instead of hardcoded values because I thought it wouldn't work but it actually does. But leaving it up to you since it's quite trivial.

kyriediculous:
This looks good to me so far but I'll be awaiting the integration test to give my final approval.

contracts/token/Minter.sol
```diff
@@ -171,7 +171,7 @@ contract Minter is Manager, IMinter {
     /**
      * @notice Deposit ETH to this contract. Only callable by the currently registered Minter or JobsManager
      */
-    function depositETH() external payable onlyMinterOrJobsManager whenSystemNotPaused returns (bool) {
+    function depositETH() external payable onlyMinterOrJobsManager returns (bool) {
```
yondonfu, line 180: This change is necessary since `migrateToNewMinter()` can only be called when the Controller is paused. Without this change, the call to `depositETH()` in `migrateToNewMinter()` would revert.

Worth noting that this requirement was not obvious (at least I missed it) until the integration test in 78d22e2 was added. The Minter unit test cases for `migrateToNewMinter()` deploy a [GenericMock](https://github.com/livepeer/protocol/blob/826af1124f645a819b54eb0500ddc31f0d2af755/test/unit/Minter.js#L141) as the new Minter which mocks the `depositETH()` implementation such that it would never revert. However, the `depositETH()` implementation prior to this change would revert if the Controller was paused.

Removing the `whenSystemNotPaused` modifier here should be safe because the only two contracts that can call `depositETH()` are the Minter and the TicketBroker (registered under the JobsManger contract ID). If the caller is the Minter, then we would be running a migration meaning that the Controller needs to be paused. The TicketBroker already has a `whenSystemNotPaused` modifier applied to the functions that invoke `depositETH()` which would prevent more funds from being sent to the Minter from users when the Controller is paused.

yondonfu:
> Deleted my previous comment about creating a getter for the constant PERC_DIVISOR and then using that getter in the solidity unit tests instead of hardcoded values because I thought it wouldn't work but it actually does. But leaving it up to you since it's quite trivial.

I have a slight preference to continue defining the constant in the tests because:

- It avoids adding an additional getter that would only be used for a Solidity test
- The `validPerc()` tests check that the library exhibits the right behavior given an expected constant value
- Even if the constant is defined in the library I think we'd still need to define the constant in the JS tests for contracts that use the library internally (i.e. BondingManager) since the getter wouldn't be exposed on the contract artifact used in the tests

yondonfu:
> This looks good to me so far but I'll be awaiting the integration test to give my final approval.

Integration test added in 78d22e2

9208057 contains a required update that was surfaced by the integration test. See [this comment](https://github.com/livepeer/protocol/pull/391#discussion_r464683531) for more details.

I'll update the LIP-40 spec after this change is reviewed.

contracts/token/Minter.sol
kyriediculous: For the record, this problem would have already existed prior to this upgrade right ? 

Would an alternative be removing the paused requirement for calling `migrateToNewMinter()` ? 

The pausing requirement seems like a weird thing in a case where we would have LIP-25, but I can see how it could still be preferred.

yondonfu: > For the record, this problem would have already existed prior to this upgrade right ?

Yep.

> Would an alternative be removing the paused requirement for calling migrateToNewMinter() ?

That doesn't help here because we would be calling `migrateToNewMinter()` on the existing Minter contract which already has a paused requirement.

test/integration/MinterUpgrade.js
```diff
@@ -0,0 +1,210 @@
...
+        await token.approve(bondingManager.address, amount, {from: transcoder1})
+        await bondingManager.bond(amount, transcoder1, {from: transcoder1})
+
+        // Register transcoder 2
+        await token.approve(bondingManager.address, amount, {from: transcoder2})
+        await bondingManager.bond(amount, transcoder2, {from: transcoder2})
+
+        const deposit = new BN(web3.utils.toWei("1", "ether"))
+        await web3.eth.sendTransaction({from: accounts[0], to: broadcaster1, value: deposit})
+        await web3.eth.sendTransaction({from: accounts[0], to: broadcaster2, value: deposit})
```
kyriediculous: This seems rather unnecessary considering all accounts in ganache should be prefunded with 100 ETH ?

kyriediculous (changes requested):

test/integration/MinterUpgrade.js
yondonfu: Fixed 6ac5977

kyriediculous:
Let's rebase !

yondonfu:
Rebased

contracts/token/Minter.sol
yondonfu: LIP-40 has been updated in livepeer/LIPs#47 to reflect these changes.

yondonfu:
Force pushed to update a Minter unit test to use a realistic mock value for `BondingManager.getTotalBonded()` (the previous value exceeded the total supply). See [this line](https://github.com/livepeer/protocol/pull/391/files#diff-17eb918a92dc9f678996ed5c37d52643R466). No other changes were made.

contracts/token/Minter.sol
kyriediculous: The boolean return value seems unnecessary but since this existed in the previous Minter contract as well I don't think we should change it. Just an observation, but it would reduce the bytecode size by a tiny amount.

kyriediculous:
While the added integration test checks for the correctness of the upgrade path it provides no guarantees about the actual execution of that upgrade. 

What did you have in mind for actually deploying those changes to production? I Don't see a migration that would currently handle this upgrade. Were you planning on adding that? The current `3_deploy_contracts.js` migration file would not work for this because we're actually changing params and not just adding contracts to the protocol. 

Do you think it would be valuable to write a separate upgrade script for this using common web3 libraries so that we can actually integration test for that script instead of the separate steps we would execute manually ? We can re-use that script in a migration. 

On a sidenote I'm actually wondering again how we executed the streamflow update because the `Migrations` contract stores the last number of the executed migration, so AFAIK we would have explicitly told truffle to re-execute `3_deploy_contracts` for the streamflow update? I don't think this is what would be convention though. 

Edit: The upgrade path checks out code-wise to me and the integration test surely helps in confirming that but what I'm a little bit more worried about is the actual upgrade execution  itself.

yondonfu:
> What did you have in mind for actually deploying those changes to production? I Don't see a migration that would currently handle this upgrade.

Upgrades on mainnet do not use migration scripts because all the upgrade operations (minus the deployment of a new contract) are currently multisig transactions that need to be manually submitted. The steps that will be manually executed are described in the [LIP-40 spec](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-40.md) and outlined in the [MinterUpgrade integration test](https://github.com/livepeer/protocol/pull/391/files?file-filters%5B%5D=.sol#diff-7431d83a81135311ac3e787d42b56dd3R115) which might be the best that we can do for now given that we need to manually execute the steps of an upgrade. In the future, if LIP-25 is deployed, we could construct the update (with all the relevant transaction data), test it and then use that same update data when executing the upgrade (a manual action would still be required, but we would always only need one instead of potentially many).

> On a sidenote I'm actually wondering again how we executed the streamflow update because the Migrations contract stores the last number of the executed migration, so AFAIK we would have explicitly told truffle to re-execute 3_deploy_contracts for the streamflow update?

The Streamflow upgrade was executed manually due to the reasons mentioned above. We currently do not rely on migration scripts outside of development and deployment to new networks.

kyriediculous:
>  are currently multisig transactions that need to be manually submitted

I see and the implementation doesn't allow for automation. Unfortunate but okay, let's be very thorough then with the actual upgrade execution itself.

kyriediculous:
I'm adding integration tests for #383 and you mentioned wanting to test for revert reasons of calls. However many of our current contracts do not have revert reasons. 

I added them to the `Minter.migrateToNewMinter` function in #383 for testability but since we are deploying  a new minter anyway what do you think of adding them for the new deployment of the Minter as well ? 

```solidity
    function migrateToNewMinter(IMinter _newMinter) external onlyControllerOwner whenSystemPaused {
        // New Minter cannot be the current Minter
        require(_newMinter != this, "New minter cannot be the current minter");
        // Check for null address
        require(address(_newMinter) != address(0), "New minter is null address");

        IController newMinterController = _newMinter.getController();
        // New Minter must have same Controller as current Minter
        require(newMinterController == controller, "New minter has a different Controller");
        // New Minter's Controller must have the current Minter registered
        require(newMinterController.getContract(keccak256("Minter")) == address(this), "Current Minter not registered with Controller");

        // Transfer ownership of token to new Minter
        livepeerToken().transferOwnership(address(_newMinter));
        // Transfer current Minter's token balance to new Minter
        livepeerToken().transfer(address(_newMinter), livepeerToken().balanceOf(address(this)));
        // Transfer current Minter's ETH balance to new Minter
        _newMinter.depositETH.value(address(this).balance)();
    }
```

yondonfu:
> since we are deploying a new minter anyway what do you think of adding them for the new deployment of the Minter as well

I think this is reasonable as revert reasons add information to error paths that already exist in the contract and do not impact contract behavior in any other way.

Added revert reasons in d9272c8

kyriediculous:
In relation to LIP-36 (#387 ): 

Currently our contract explicitly implement their interfaces  (i.e `contract Minter is IMinter`) but this seems rather unnecessary and even weird (and it increases the bytecode size slightly). 

Even for ERC20 it is weird because the contract doesn't need to implement the interface for itself but for external callers that require it to implement the interface. The only thing this explicitness really gives us is compile-time errors when a contract doesn't implement its interface correctly. 

Furthermore it restricts us from adding public variable getters to the interface as it results in multiple declarations with the same name. 

e.g. `uint public a` would implicitly result in a getter `function a() public view returns(uint)` 

The workaround is very obvious , create an additional getter `function getA() public view returns(uint)` , but that again seems rather unnecessary to have duplicate getters when it's avoidable. 

Therefore I would suggest for the `Minter` to no longer explicitly inherit its interface. This would allow us to add the related getter to the interface for external contracts to use without having to add additional getters to the implementation (and interface) itself. 

___ 

There are no behavioural changes that result from this. The only real drawback is that explicit type conversions from `Minter` to `IMinter` are not allowed as it would no longer inherit its interface. That only results in us having to remove the following check in `migrateToNewMinter()`: `require(_newMinter != this);`. Seeing we already check for the addresses to not be equal I think we can safely  remove this as well as explicit interface inheritance.

yondonfu:
@kyriediculous and I discussed the above comment in Discord. The conclusion was that if we define the following in the IMinter abstract contract:

```solidity
function currentMintableTokens() external view returns (uint256);
```

then the Minter could override the external `currentMintableTokens()` function with the public getter for the `currentMintableTokens` state variable. Given that this workaround is available, the decision was made to leave the code in this PR as-is and to freeze it for deployment.

kyriediculous (approved):

PR #396, opened 2020-09-03
LIP-56 Implementation

yondonfu:

This PR implemented [LIP-56](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-56.md).

- 
- 
- 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -400,6 +400,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         currentRoundTotalActiveStake = nextRoundTotalActiveStake;
     }
 
+    function mintRewards() external onlyRoundsManager returns (uint256) {
+        return minter().createReward(1, 1);
+    }
+
     /**
      * @notice Sets commission rates as a transcoder and if the caller is not in the transcoder pool tries to add it using an optional list hint
      * @dev Percentages are represented as numerators of fractions over MathUtils.PERC_DIVISOR. If the caller is going to be added to the pool, the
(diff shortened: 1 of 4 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -660,9 +664,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             earningsPool.setStake(t.earningsPoolPerRound[lastUpdateRound].totalStake);
         }
 
+        uint256 rewardTokens = 0;
         // Create reward based on active transcoder's stake relative to the total active stake
-        // rewardTokens = (current mintable tokens for the round * active transcoder stake) / total active stake
-        uint256 rewardTokens = minter().createReward(earningsPool.totalStake, currentRoundTotalActiveStake);
+        if (currentRound >= roundsManager().lipUpgradeRound(56)) {
```
kyriediculous, line 669: Do we need this if-else statement ? Reward can't be called for a previous round so the code path for the old approach wouldn't be available after the upgrade anyway (i.e. `currentRound >= roundsManager().lipUpgradeRound(56)` will always evaluate to true after the upgrade took place, no ?)  

The only real reason I can see this being useful is for testing purposes and private deployments?

contracts/rounds/RoundsManager.sol
```diff
@@ -163,6 +204,10 @@ contract RoundsManager is ManagerProxyTarget, IRoundsManager {
         return blockNum().sub(currentRoundStartBlock()) >= roundLength.sub(lockedBlocks);
     }
 
+    function currentRoundIsRewardRound() public view returns (bool) {
+        return nextRewardRound.sub(1) == currentRound();
```
kyriediculous, line 208: Shouldn't this be `nextRewardRound == currentRound` ?  Otherwise L134 would have to be `if (currRound >= nextRewardRound` , no?

PR #397, opened 2020-09-05
LIP-52: claimEarnings snapshot

kyriediculous:

- add a `MerkleSnapshot` contract 
- add unit  tests for `MerkleSnapshot` contract
- add `claimSnapshotEarnings` method to `BondingManager`
- add `claimSnapshotEarnings` unit tests
- bump `ethereumjs-util` library dependency from `5.1.2` to `6.0.0` and resolve incompatibilities
- `MerkleTree` JS class based on // https://github.com/OpenZeppelin/zeppelin-solidity/blob/master/test/helpers/merkleTree.js

Added unit tests and integration tests 

https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-52.md

**Checklist:**
- [] README and other documentation updated


contracts/bonding/BondingManager.sol
```diff
@@ -393,6 +394,61 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        require(lastClaimRound <= LIP52Round, "delegator has already claimed earnings past the LIP-52 round");
+
+        bytes32 leaf = keccak256(abi.encode(msg.sender, _pendingStake, _pendingFees));
+
+        require(
+            IMerkleSnapshot(controller.getContract(keccak256("MerkleSnapshot"))).verify(keccak256("LIP-52"), _earningsProof, leaf)
+        );
+
+        emit EarningsClaimed(
+            delegators[msg.sender].delegateAddress,
```
yondonfu: Could this be replaced with `del` since we already define the storage pointer on L415?

contracts/snapshots/MerkleSnapshot.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.11;
+
+import "../zeppelin//MerkleProof.sol";
```
yondonfu: Looks like there is an extra forward slash before "MerkleProof.sol". Interesting that this import still resolves correctly though...

contracts/snapshots/MerkleSnapshot.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    constructor(address _controller) public Manager(_controller) {}
+
+    function setSnapshot(bytes32 _id, bytes32 _root) external onlyControllerOwner {
+        require(snapshot[_id] == bytes32(0), "Snapshot root for ID already exists");
+        snapshot[_id] = _root;
+    }
+
+    function verify(bytes32 _id, bytes32[] calldata _proof, bytes32 _leaf) external view returns (bool) {
+        require(
```
yondonfu: Since this is defined as a view function that returns a boolean could we just let the caller deal with handling the boolean return value? 

i.e.

```solidity
function verify(bytes32 _id, bytes32[] calldata _proof, bytes32 _leaf) external view returns (bool) {
    return MerkleProof.verify(_proof, snapshot[_id], _leaf);
}
```

I think an additional benefit of doing this is that if when this function is called off-chain the caller just has to handle the returned boolean (I'm not sure what the current behavior is when the function reverts for an `eth_call` - I recall there being some funky behavior in the past).

contracts/snapshots/MerkleSnapshot.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.11;
+
+import "../zeppelin//MerkleProof.sol";
+import "../ManagerProxyTarget.sol";
+
+
+contract MerkleSnapshot is ManagerProxyTarget{
```
yondonfu: Could this contract inherit from `Manager` instead of `ManagerProxyTarget`?

The contract would only need to inherit from `ManagerProxyTarget` if it needs to be upgradeable as a part of a proxy setup (so that the proxy and the implementation share the same exact storage layout). Given that this contract is relatively simple and the state would be easily migrated if needed, I don't think it needs to be upgradeable as a part of a proxy setup. WDYT?

contracts/snapshots/MerkleSnapshot.sol
```diff
@@ -0,0 +1,25 @@
...
+    mapping (bytes32 => bytes32) public snapshot;
+
+    constructor(address _controller) public Manager(_controller) {}
+
+    function setSnapshot(bytes32 _id, bytes32 _root) external onlyControllerOwner {
+        require(snapshot[_id] == bytes32(0), "Snapshot root for ID already exists");
```
yondonfu: I wonder if it could make sense to not have this check here because:

- Not having this check here could make it easier to recover from the hopefully highly unlikely case where the initial snapshot is incorrect (i.e. an address is missing from the snapshot) such that it needs to be updated. This would be especially helpful if this contract is not upgradeable as a part of a proxy setup
- The Controller owner can still swap out snapshots if needed regardless of whether this check is here or not

WDYT?

yondonfu (changes requested):
Will take another pass at review after the rebase on top of #390 and the target branch change tp that PR branch.

kyriediculous:
5d171f1 moves the logic from the modifiers into internal functions to reduce the bytecode length. 

Force-pushed to rebase on top of LIP-36

contracts/snapshots/MerkleSnapshot.sol
kyriediculous: fixed in 2839cec

contracts/snapshots/MerkleSnapshot.sol
kyriediculous: fixed in 2839cec

contracts/snapshots/MerkleSnapshot.sol
kyriediculous: fixed in 2839cec

contracts/snapshots/MerkleSnapshot.sol
kyriediculous: fixed in 2839cec

contracts/bonding/BondingManager.sol
kyriediculous: fixed in c471e04

contracts/bonding/BondingManager.sol
```diff
@@ -449,6 +440,59 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        whenSystemNotPaused
+        currentRoundInitialized
+    {
+        Delegator storage del = delegators[msg.sender];
+
+        uint256 lip52Round = roundsManager().lipUpgradeRound(52);
+
+        uint256 lastClaimRound = del.lastClaimRound;
+
+        require(lastClaimRound <= lip52Round, "delegator has already claimed earnings past the LIP-52 round");
```
yondonfu: Should we make this check `lastClaimRound < lip52Round` to guarantee that `claimSnapshotEarnings()` can only ever be called once? At the moment, since `lastClaimRound` is set to `lip52Round` on L485, it is possible for a delegator to call `claimSnapshotEarnings()` more than once. In practice, this shouldn't have any major adverse effects on contract state, but it does mean that an `EarningsClaimed` event with the same startRound and endRound values could be emitted multiple times. It'd be nice to preserve the invariant where an `EarningsClaimed` event can only be emitted for a round range once.

contracts/bonding/BondingManager.sol
```diff
@@ -1319,8 +1363,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function updateDelegatorWithEarnings(address _delegator, uint256 _endRound, uint256 _lastClaimRound) internal {
         Delegator storage del = delegators[_delegator];
         uint256 startRound = _lastClaimRound.add(1);
-        uint256 currentBondedAmount = del.bondedAmount;
-        uint256 currentFees = del.fees;
+        uint256 currentBondedAmount;
```
yondonfu: We should continue to initialize these variables. If a delegator for some reason has a null delegate and then invokes a staking function then its bondedAmount and fees would be set to 0.

test/integration/LIPs/LIP52_ClaimEarningsSnapshot.js
```diff
@@ -0,0 +1,573 @@
...
+const BondingManager = artifacts.require("BondingManager")
+const AdjustableRoundsManager = artifacts.require("AdjustableRoundsManager")
+const LivepeerToken = artifacts.require("LivepeerToken")
+const TicketBroker = artifacts.require("TicketBroker")
+const MerkleSnapshot = artifacts.require("MerkleSnapshot")
+const BondingManagerPreLIP36 = artifacts.require("BondingManagerPreLIP36")
+const LinkedList = artifacts.require("SortedDoublyLL")
+const ManagerProxy = artifacts.require("ManagerProxy")
+
+async function executeLIP36Upgrade(controller, roundsManager, bondingAddress) {
```
yondonfu: Might make sense to extract this into a test helper file as the canonical LIP-36 upgrade function since it is used here as well as in `test/integration/Earnings.js`.

test/integration/LIPs/LIP52_ClaimEarningsSnapshot.js
```diff
@@ -0,0 +1,573 @@
...
+
+        it("there should be residual rewards (this is a bug)", async () => {
+            const currentRound = await roundsManager.currentRound()
+
+            const pendingStake = await bondingManager.pendingStake(transcoder, currentRound)
+            const data = bondingManager.contract.methods.unbond(pendingStake.toString()).encodeABI()
+            await bondingManager.claimSnapshotEarnings(pendingStake, new BN(0), proof, data)
+
+            const delegatorAfter = await bondingManager.getDelegator(transcoder)
+            console.log(delegatorAfter.bondedAmount.toString())
```
yondonfu: Can remove?

yondonfu (changes requested):

test/integration/LIPs/LIP52_ClaimEarningsSnapshot.js
kyriediculous: fixed in b05f3fd

test/integration/LIPs/LIP52_ClaimEarningsSnapshot.js
kyriediculous: fixed in b05f3fd

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 06e8d46

contracts/bonding/BondingManager.sol
kyriediculous: fixed in 06e8d46

contracts/bonding/BondingManager.sol
```diff
@@ -1425,4 +1469,34 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     function roundsManager() internal view returns (IRoundsManager) {
         return IRoundsManager(controller.getContract(keccak256("RoundsManager")));
     }
+
+    function _onlyTicketBroker() internal {
```
yondonfu: All of these internal functions besides `_autoClaimEarnings()` can be marked with `view`.

migrations/3_deploy_contracts.js
```diff
@@ -4,13 +4,14 @@ const ContractDeployer = require("../utils/contractDeployer")
 const Controller = artifacts.require("Controller")
 const Minter = artifacts.require("Minter")
 const ServiceRegistry = artifacts.require("ServiceRegistry")
-const BondingManager = artifacts.require("BondingManager")
+const BondingManager = artifacts.require("BondingManager.sol")
```
yondonfu: Don't think we need the ".sol"?

test/helpers/executeLIP36Upgrade.js
```diff
@@ -0,0 +1,35 @@
+const BondingManager = artifacts.require("BondingManager")
+const LinkedList = artifacts.require("SortedDoublyLL")
+const AdjustableRoundsManager = artifacts.require("AdjustableRoundsManager")
+
+const {contractId} = require("../../utils/helpers")
+import BN from "bn.js"
+
+module.exports = async function(controller, roundsManager, bondingAddress) {
```
yondonfu: Worth considering a more descriptive var name than "bondingAddress" - maybe "newBondingManagerAddr"?

yondonfu (changes requested):

test/helpers/executeLIP36Upgrade.js
kyriediculous: let's roll with `bondingManagerProxyAddress` then that's most accurate.

kyriediculous: fixed in e3284a4

migrations/3_deploy_contracts.js
kyriediculous: fixed in e3284a4

contracts/bonding/BondingManager.sol
kyriediculous: fixed in e3284a4

yondonfu:
Changed the base branch to `streamflow` since the LIP-36 branch has already been merged there.

@kyriediculous Can you resolve the conflicts and rebase to squash the fixup commits on this branch? PR looks ready to merge after that.

yondonfu:
Just FYI looks like a few of the changes in fixup commits were dropped during the rebase. Examples:

- https://github.com/livepeer/protocol/pull/397/files#r496970771
- https://github.com/livepeer/protocol/pull/397/files#r501957068

yondonfu (approved):
LGTM 🚢

Issue #393, opened 2020-08-12
Flaky test for 300 transcoders


kyriediculous:
CI Often fails on this test for the gas report for 300 transcoders due to the RPC connection no longer being available. 

 ```
 652 passing (23m)
  1 failing

  1) Contract: transcoder pool size gas report
       300 transcoders
         "before all" hook in "300 transcoders":
```

yondonfu:
Doesn't seem to be an issue anymore.

PR #399, opened 2020-09-09
Updates to LIP-36 PR

yondonfu:

[Overall diff](https://github.com/livepeer/protocol/compare/yf/cumulative-earnings-changes) against `streamflow` branch.

The BondingManager bytecode size is now 23280. The size can be calculated with the following steps:

```
npm run compile
cat build/contracts/BondingManager.json | jq '.deployedBytecode' | wc -m | xargs -I {} echo "{} / 2 - 1" | bc
```

- Cherry-picked a75e2b8 from `streamflow` to avoid dealing with rebase conflicts for now
- Fixed failing test cases in #390 
- Re-use ` pendingStakeAndFees()` helper in `pendingStake()`, `pendingFees()` and `updateDelegatorWithEarnings()`
   - This also means that the pre-LIP36 earnings claiming algorithm will no longer update the reward/fee pool and claimable stake for each round and that rewards/fees in a pool that are unaccounted for due to a delegator claiming before those rewards/fees are added to the pool will no longer be re-distributed to other delegators that have not claimed from the pool yet. After looking at the code some more, I agree with @kyriediculous 's point that the advantages of this change outweigh the disadvantages
- Reduce the diff for EarningsPool
    - Use separate `updateCumulativeRewardFactor()` and `updateCumulativeFeeFactor()` functions from EarningsPoolLIP36
    - Remove deprecated functions `addToRewardPool()`, `addToFeePool()` and `claimShare()` from EarningsPool
- Remove unneeded storage update for activeCumulativeRewards in `updateTranscoderWithFees()`
- Updated ethlint config and fixed all warnings/errors
- Reverted a few temporary changes

May still be useful to add additional tests, but left that out for now.

Ran test suite.

N/A

**Checklist:**


contracts/bonding/BondingManager.sol
```diff
@@ -869,24 +792,27 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+     * @return cumulativeFees Transcoder's cumulative fees (earned via its active staked rewards and its fee share)
+     * @return lastFeeRound Latest round that the transcoder received fees
      */
     function getTranscoder(
         address _transcoder
     )
         public
         view
-        returns (uint256 lastRewardRound, uint256 lastFeeRound, uint256 rewardCut, uint256 feeShare, uint256 lastActiveStakeUpdateRound, uint256 activationRound, uint256 deactivationRound, uint256 activeCumulativeRewards, uint256 cumulativeRewards, uint256 cumulativeFees)
+        returns (uint256 lastRewardRound, uint256 rewardCut, uint256 feeShare, uint256 lastActiveStakeUpdateRound, uint256 activationRound, uint256 deactivationRound, uint256 activeCumulativeRewards, uint256 cumulativeRewards, uint256 cumulativeFees, uint256 lastFeeRound)
```
kyriediculous, line 814: I understand why this was added to the end, it breaks clients still accessing tuples using numeric indexes. But most web3 libraries now also convert the tuples to objects with named so index based access is no longer needed. 

I felt like grouping lastRewardRound and lastFeeRound was a little clearer in terms of readability. 

Just to clarify why I added it in between, the tests that still used index based access were changed to used object keys instead. But I'll leave it up to you in terms what you think is the best tradeoff here (readability or old client compatibility)

test/integration/Earnings.js
```diff
@@ -334,7 +317,6 @@ contract("Earnings", accounts => {
 
             console.log("last transcoder stake and fees", (await bondingManager.pendingStake(transcoder, currentRound)).toString(), (await bondingManager.pendingFees(transcoder, currentRound)).toString())
             console.log("last delegator stake and fees", (await bondingManager.pendingStake(delegator, currentRound)).toString(), (await bondingManager.pendingFees(delegator, currentRound)).toString())
```
kyriediculous: Can remove this

kyriediculous (commented):
Did a first pass, going to pull in the branch to do a second pass in the vscode UI and run some testing on the actual state transitions. Looks good so far though !

yondonfu:
f53c96a Fixes a failing test case related to updating activeCumulativeRewards in updateTranscoderWithFees
ab87bf6 Always sets lastFeeRound to the current round since its possible (although unlikely in practice due to the way the client is implemented) for a ticket to be redeemed with round N and then a ticket to be redeemed with round N - 1.

test/gas-report/poolsize.js
```diff
@@ -395,5 +395,5 @@ contract("transcoder pool size gas report", accounts => {
 
     testWithPoolSize(100)
     testWithPoolSize(200)
-    testWithPoolSize(300)
+    // testWithPoolSize(300)
```
kyriediculous: This was re-commented (it wasn't prior to this commit) due to coverage being flaky on this ? https://github.com/livepeer/protocol/issues/393

test/unit/BondingManager.js
```diff
@@ -4,6 +4,7 @@ import {contractId, functionSig, functionEncodedABI} from "../../utils/helpers"
 import {constants} from "../../utils/constants"
 import BN from "bn.js"
 import truffleAssert from "truffle-assertions"
+import {assert} from "chai"
```
kyriediculous, line 7: Looks like this was re-added by vscode plugins upon saving?  I don't mind leaving it there tbh as it's the actual dependency that's being injected anyway.

contracts/bonding/BondingManager.sol
```diff
@@ -1263,72 +1297,47 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
 
+            (
+                currentBondedAmount,
+                currentFees
+            ) = pendingStakeAndFees(_delegator, _endRound);
+
             // Check whether the endEarningsPool is initialised
             // If it is not initialised set it's cumulative earnings factors (fees and rewards)
             // This serves two purposes:
             // - Initialise the correct values for the current claimEarnings 'epoch' to calculate cumulative earnings
```
kyriediculous: I believe this is no longer correct after  the changes, it only serves the single purpose of initialising the correct `lastClaimRound` earningsPool for the next "epoch" (for lack of better terminology)

kyriediculous:
Ran some tests and changes look good to me otherwise , just needs a `npm run eslint:fix` it seems.

test/integration/Earnings.js
yondonfu: Fixed in fcedeac

contracts/bonding/BondingManager.sol
yondonfu: Fixed in 654de5b

test/gas-report/poolsize.js
yondonfu: Fixed in 654de5b

kyriediculous (approved):
LGTM pending green CI

PR #439, opened 2021-08-03
Move CI to github actions

kyriediculous:
Use github actions for CI and fix coveralls run

**Checklist:**


yondonfu:
The CircleCI build is still being triggered even though the config file is removed because this repo is still configured to send a request to a CircleCI webhook for PRs which then causes CircleCI to complain about a missing config file when the build is started. We can remove the webhook from the repo, but that would also disable CircleCI builds for the `streamflow` branch. 

Since we're not merging the `next` branch into the `streamflow` branch for awhile and in order to support CI builds on `streamflow` if anything needs to be merged there before the `next` branch is merged in, let's set the target branch for this PR to `streamflow` so we can continue to support CI builds there with GH actions. Then, the `next` branch can just include the relevant commits with the GH action build as well.

After the GH action build + test is passing here we can remove the CircleCI webhook from the repo to get rid of the failing CircleCI build.

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: Build
+on: [push]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "🔎 The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
```
yondonfu: Can remove?

.github/workflows/build.yml
```diff
@@ -0,0 +1,36 @@
+name: Build
+on: [push]
```
yondonfu:
```suggestion
on:
    pull_request:
    push:
        branches:
            - streamflow
            - next
```

^ So we can run CI off any PR including those from forks while continuing to run for pushes to the `streamflow` and `next` branches. This would be similar to how CI works for [go-livepeer](https://github.com/livepeer/go-livepeer/blob/849c8bf43a73ce8dcc530f044cb8642192ff1331/.github/workflows/linux.yml#L3).

yondonfu (changes requested):

yondonfu:
We should be able to remove the CircleCI config file right? It's ok if CircleCI still fails due to the missing config file since we'll disable it after this PR is merged anyway as long as the GH action passes.

kyriediculous:
Removed circleCI config

yondonfu:
Looks like the Coveralls step of the GH action is [failing](https://github.com/livepeer/protocol/pull/439/checks?check_run_id=3295506144) with:

```
Using lcov file: ./coverage/lcov.info
Error:  "2021-08-11T00:33:49.819Z"  'error from getOptions'
Error: Error: Command failed: git cat-file -p a807a1a3983eb6dd6c1fa0f07c5c5da90af36095
fatal: Not a valid object name a807a1a3983eb6dd6c1fa0f07c5c5da90af36095
```

yondonfu (approved):
LGTM

I'll remove the CircleCI webhook now so it stops getting triggered!

We should make sure to pick up this commit on `next` as well so that it can exclusively use GH actions as well. See [this other comment](https://github.com/livepeer/protocol/pull/446#discussion_r685544094) for additional commentary on that topic.

kyriediculous:
Are we going to remove circleCI before merging this ?

yondonfu:
> Are we going to remove circleCI before merging this ?

The CircleCI webhook has already been removed for this repo so future builds shouldn't trigger anything on CircleCI's end.

PR #440, opened 2021-08-03
removed rpc class which used web3 to save checkpoints

kautukkundan:
removed rpc class which used web3 to save checkpoints


kyriediculous:
Removing the snapshotting and reverting from integration tests seems to do the trick. Since hardhat-deploy's fixtures already call these hooks as well. https://hardhat.org/plugins/hardhat-deploy.html#testing-deployed-contracts

kyriediculous (approved):

PR #446, opened 2021-08-09
removed npm and added yarn

### LINKED ISSUE #419: Use Yarn instead of NPM
There's several benefits in using Yarn over NPM, most notably: 
- Yarn strongly binds package versions
- Yarn can install packages from a local cache, not requiring network config
- Yarn installs packages in parallel, greatly improving performance
- Monorepo support (multiple packages) without the requirement for Lerna

kautukkundan:

removed npm and added yarn

- removed package-lock.json
- updated package.json to use yarn
- updated docker configs to use yarn
- updated documentation

ran `yarn test`

closes #419 

**Checklist:**


hardhat.config.ts
```diff
@@ -37,6 +37,7 @@ const config: HardhatUserConfig = {
     networks: {
         hardhat: {
             gas: 12000000,
+            hardfork: "berlin",
```
kautukkundan, line 40: solidity-coverage now uses [hardhat release v2.6.0](https://github.com/nomiclabs/hardhat/releases/tag/hardhat-core-v2.6.0) by default which uses the London hardfork and enables eip-1559. This was causing the original tests to fail. The workaround was to explicitly set the hardfork version to berlin to fall back to the old behaviour. However, I think the test suite/config should be updated going forward. what do you think? @kyriediculous @yondonfu 

Screenshot 2021-08-10 at 12 54 22 AM

it also drops support for node v10 which is used in github actions in PR #439

yondonfu: > This was causing the original tests to fail. The workaround was to explicitly set the hardfork version to berlin to fall back to the old behaviour. However, I think the test suite/config should be updated going forward. what do you think?

I think pinning the hardfork value to `berlin` in this PR is ok, but agreed that separately we should a) switch the hardfork value to `london` i.e. the default and b) make sure the tests pass when the hardfork value is set to `london`.

@kautukkundan Can you create an issue tracking the above and take a look at that?

> it also drops support for node v10 which is used in github actions in PR

Let's make note of the required Node version in the README.

Regarding #439 - I think we can do the following:

- Keep the Node version used in that PR since it is being merged into `streamflow`
- Rebase `next` on top of `streamflow` and modify the GH action to use a newer Node version supported by hardhat [1]

[1] Or cherry pick the commit. Whichever is easier.

yondonfu: > Let's make note of the required Node version in the README.

Bump on the above. After this should be ready to merge!

PR #456, opened 2021-08-13
removed npm and added yarn: re

kautukkundan:
**What does this pull request do? Explain your changes. (required)**

removed npm and added yarn

**Specific updates (required)**

- removed package-lock.json
- updated package.json to use yarn
- updated docker configs to use yarn
- updated documentation

**How did you test each of these updates (required)**

ran `yarn test`

**Does this pull request close any open issues?**

closes #419 

**Checklist:**

- [x] README and other documentation updated
- [x] All tests using `yarn test` pass

PR #457, opened 2021-08-15
Added temporary fix for london hardfork compatibility

### LINKED ISSUE #447: modify test suite to work with London upgrade
solidity-coverage now uses [hardhat v2.6.0](https://github.com/nomiclabs/hardhat/releases/tag/hardhat-core-v2.6.0) by default which uses the London hardfork and enables eip-1559. This is causing the original tests to fail. The current workaround is to explicitly set the hardfork version to `berlin` to fall back to the old behaviour. However this needs to be corrected such that the test suite can work properly with the latest release.

- [ ] update hardhat version to v2.6.0 for local test environment
- [ ] remove hardfork = `berlin` in the config
- [ ] modify tests to pass in accordance with the new config

kautukkundan:

The current `solidity-coverage` version overrides the gasPrice during execution and sets it equal to 1. This creates an issue with the EIP-1559 "baseFeePerGas". There is an open issue sc-forks/solidity-coverage/issues/652 to deal with this. This PR includes a temporary and "hacky" solution to overcome the issue by setting the `initialBaseFeePerGas` equal to 0 in the hardhat network config. However, moving forward this should not be the final solution. 

- update hardhat version to v2.6.0
- set `initialBaseFeePerGas` equal to 0 in the hardhat config

ran `yarn test:coverage`

closes #447 

**Checklist:**

- [x] All tests using `yarn run test` pass
- [x] need to merge #456 before merging this

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -7,7 +7,7 @@ import "hardhat-abi-exporter"
 
 // deployment plugins
 import "hardhat-deploy"
-import "hardhat-deploy-ethers"
+import "@nomiclabs/hardhat-ethers"
 
 import "solidity-coverage"
 
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -37,7 +37,7 @@ const config: HardhatUserConfig = {
     networks: {
         hardhat: {
             gas: 12000000,
-            hardfork: "berlin",
+            initialBaseFeePerGas: 0,
             allowUnlimitedContractSize: true,
             blockGasLimit: 12000000,
             accounts: {
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "homepage": "https://github.com/livepeer/protocol#readme",
   "devDependencies": {
-    "@nomiclabs/hardhat-ethers": "npm:hardhat-deploy-ethers@^0.3.0-beta.10",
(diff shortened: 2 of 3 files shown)


hardhat.config.ts
```diff
@@ -7,7 +7,7 @@ import "hardhat-abi-exporter"
 
 // deployment plugins
 import "hardhat-deploy"
-import "hardhat-deploy-ethers"
+import "@nomiclabs/hardhat-ethers"
```
kautukkundan, line 10: based on https://github.com/wighawag/hardhat-deploy-ethers#installation
the compile step was failing due to yarn.lock hence had to change this

yondonfu (commented):
This generally looks good to me, but I'll let @kyriediculous give the approval so he has a chance to skim the hardhat related dependency updates.

kyriediculous:
RE the dependency issue this looks good to me, I think I installed it the wrong way (outdated documentation from a blog rather than the official one) which then persisted itself in the lock file. Not sure why it wasn't caught earlier either, maybe with the move to yarn. 

Test update also looks good to me, let's change the base to `nv/next` maybe since that has the conflicts resolved with `streamflow` and then we can merge `nv/next` into `next` and then `next` into `streamflow`. WDYT ?

kyriediculous (approved):

PR #465, opened 2021-08-18
Bootstrap the skeleton for share-based accounting

### LINKED ISSUE #449: Upgrade Solidity compiler version

### LINKED ISSUE #450: Remove SafeMath
Because newer compiler versions have built-in support for this with arithmetic operators.

### LINKED ISSUE #451: Upgrade OZ deps

### LINKED ISSUE #452: Add Confluence staking PoC as skeleton in repo
https://github.com/kyriediculous/protocol-v2/tree/poc

kyriediculous:
Bootstrap the skeleton for share-based accounting

- Upgrade compiler version and fix it to 0.8.4
- Update openzeppelin/contracts
- Remove hardcoded openzeppelin dependencies
- Update `LivepeerToken` implementation with a new underlying `ERC20` implementation , this affects some of the constructor initialization
- Rework inheritance structure of the `pm` contracts (removed the interfaces, these are not very useful given that all contracts are just inherited in the end, kept some of those files purely for event and data type definitions)
- Remove all the different `MathUtils` libs and go for a singular one (`PreciseMathUtils`)

Ran `yarn compile`

Fixes #449
Fixes #450 
Fixes #451 
Fixes #452 

**Checklist:**


contracts/pm/TicketBroker.sol
```diff
@@ -11,8 +9,8 @@ import "./mixins/MixinWrappers.sol";
 contract TicketBroker is
     MixinContractRegistry,
     MixinReserve,
-    MixinTicketBrokerCore,
     MixinTicketProcessor,
+    MixinTicketBrokerCore,
```
kautukkundan: What importance does the order of inheritance have here?

contracts/rounds/AdjustableRoundsManager.sol
```diff
@@ -20,11 +24,11 @@ contract AdjustableRoundsManager is RoundsManager {
         num += _blocks;
     }
 
-    function blockNum() public view returns (uint256) {
+    function blockNum() public view override(RoundsManager) returns (uint256) {
```
kautukkundan, line 27: since this contract has a single parent, can't we just use the override keyword without mentioning the parent contract?

contracts/Manager.sol
```diff
@@ -31,15 +34,15 @@ contract Manager is IManager {
...
+    constructor(address _controller) {
         controller = IController(_controller);
     }
 
     /**
      * @notice Set controller. Only callable by current controller
      * @param _controller Controller contract address
      */
-    function setController(address _controller) external onlyController {
+    function setController(address _controller) external override onlyController {
```
kautukkundan, line 45: we don't need to override

contracts/refund/Refunder.sol
```diff
@@ -1,47 +0,0 @@
-pragma solidity ^0.5.11;
```
kautukkundan, line 1: This and `AlphaJobsManager` files should be restored (for streamflow). Maybe we can remove them in another PR

contracts/test/TestMathUtils.sol
```diff
@@ -17,12 +17,17 @@ contract TestMathUtils {
...
-        Assert.equal(MathUtils.percOf(100, 100000), 10, "10% of 100 is 10");
+        Assert.equal(MathUtils.percOf(100, 3), 0, ".0000000000000000000000003% of 100 is 0");
+        Assert.equal(MathUtils.percOf(10**27, 1), 1, ".0000000000000000000000001% of 1000000000 is 1");
+        Assert.equal(MathUtils.percOf(100, 10**27 / 10), 10, "10% of 100 is 10");
     }
 
     function test_percPoints() public {
-        Assert.equal(MathUtils.percPoints(3, 4), 750000, "3/4 should convert to valid percentage");
-        Assert.equal(MathUtils.percPoints(100, 300), 333333, "100/300 should convert to valid percentage");
+        Assert.equal(MathUtils.percPoints(3, 4), 750000000000000000000000000, "3/4 should convert to valid percentage");
```
kautukkundan, line 26: scary numbers

kautukkundan (commented):
The failing CI error is due to the hardhat-ethers lib which we discussed earlier
#457 fixes that, can rebase from that once it is merged to solve the error

contracts/rounds/AdjustableRoundsManager.sol
kyriediculous: RoundsManager inherits from `IRoundsManager` which is an interface, therefore `RoundsManager` has to `override` the interface definitions that are implicitly defined as `virtual`. 

Since `AdjustableRoundsManager` then inherits `RoundsManager` but `AdjustableRoundsManager` itself also overrides both the functions from `RoundsManager` and `IRoundsManager`. Therefore we need the `override` modifier here and the `virtual` modifier in the respective `RoundsManager` functions as well.

contracts/pm/TicketBroker.sol
kyriediculous: This is simply a more explicit way of writing `TicketBroker is MixinWrappers` since `MixinWrappers` already inherits all of the underlying contracts, whereby the order is `most "base-like" contract > most derived contract` or `low level > top level`

```solidity
     contract X {}
     contract A is X {}
     contract C is X, A {}
```

contracts/Manager.sol
kyriediculous: Since this function is also defined in the `IManager` interface, which this contract inherits we need to override

contracts/refund/Refunder.sol
kyriediculous: We will not be merging this into the `streamflow` branch , rather keep this as the `next` branch and eventually the `confluence` upgrade.

contracts/test/TestMathUtils.sol
kyriediculous: Open to suggestions, but feels out of scope for this PR

kautukkundan (changes requested):
look great 👍🏼

kautukkundan (dismissed):

contracts/IController.sol
```diff
@@ -1,17 +1,21 @@
-pragma solidity ^0.5.11;
+// SPDX-FileCopyrightText: 2021 Livepeer <info@livepeer.org>
 
-import "./zeppelin/Pausable.sol";
+// SPDX-License-Identifier: GPL-3.0
```
yondonfu: We use the MIT license for Livepeer contracts right now so unless the code in a file is from a project that is using a different license, we should specify MIT for now.

kyriediculous: fixed in 1954d82

kautukkundan (approved):

PR #468, opened 2021-08-20
Staking support on behalf of another address

### LINKED ISSUE #453: Support for staking on behalf of another address

kautukkundan:

Modified bonding manager contract to extend "stake for" feature. This allows an address to stake LPT on behalf of another address.

- modified `bondWithHint` function to accept `onBehalfOf` address
- modified `bond` function to pass `msg.sender` as `onBehalfOf`
- added a new function `bondOnBehalf`

closes #453 

**Checklist:**


kyriediculous:
Cool that you opened a draft PR for this, I have some ideas that can be useful to improve the API a bit that I didn't consider before.

My current thinking is that we can remove the functions that don't accept hints and just have those that do. If hints then become the default we can just drop the `withHint` suffix. 

Then I think it would be good to separate the two functions: 
- `bond(uint256 _amount, address _orchestrator, ... hints)` which calls the internal function and supplies `msg.sender` as `_for` argument
- `bondFor(uint256 _amount, address _orchestrator, address _for, ... hints)` which calls the internal function with the same arguments as provided in the external call

and an internal function 
- `bond(uint256 _amount, address _orchestrator, address _for, ... hints)`

.eslintrc.js
```diff
@@ -19,6 +19,8 @@ module.exports = {
         "quotes": ["error", "double"],
         "comma-dangle": ["error", "never"],
         "indent": ["error", 4],
-        "space-infix-ops": ["error"]
+        "space-infix-ops": ["error"],
+        "camelcase": "off",
+        "max-len": ["error", {"code": 256}]
```
kautukkundan, line 24: the current test cases have a max length of 256, so explicitly defined it such that the new code is consistent. Open to suggestions here if we want to reduce it. the default is 80.

.eslintrc.js
```diff
@@ -19,6 +19,8 @@ module.exports = {
         "quotes": ["error", "double"],
         "comma-dangle": ["error", "never"],
         "indent": ["error", 4],
-        "space-infix-ops": ["error"]
+        "space-infix-ops": ["error"],
+        "camelcase": "off",
```
kautukkundan, line 23: typechain uses this format `ContractName__Factory` which is not consistent with the default camelcase format of eslint.

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-    "solidity.compileUsingRemoteVersion": "v0.8.4+commit.c7e474f2"
-}
\ No newline at end of file
+    "solidity.compileUsingRemoteVersion": "v0.8.4+commit.c7e474f2",
+    "solidity.defaultCompiler": "remote"
```
kautukkundan, line 3: VScode kept messing with the solidity version in case of multiple installed compilers. Explicitly defining which version to use fixed it.

kautukkundan:
Added some basic tests for bonding manager bond function.

contracts/bonding/BondingManager.sol
```diff
@@ -137,7 +137,21 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        bondWithHint(_amount, _to, msg.sender, address(0), address(0), address(0), address(0));
+    }
+
+    /**
+     * @notice Delegate stake towards a specific address on behalf of another address
+     * @param _amount The amount of tokens to stake
+     * @param _to The address of the transcoder to stake towards
+     * @param _onBehalfOf The address which will own the stake
+     */
+    function bondOnBehalf(
```
kyriediculous: I think `bondFor` is more concise and conveys the same thing. 

As discussed during the pod meeting and in [this comment ](https://github.com/livepeer/protocol/pull/468#issuecomment-902835911) we can drop the external functions that do not take hints because we want these to be the default .

The reason why they exist without hints is backwards compatibility which is not a requirement here.

contracts/bonding/Delegations.sol
```diff
@@ -160,6 +160,7 @@ library Delegations {
      * @return stake total stake of the delegator
      */
     function stakeOf(Pool storage _pool, address _delegator) internal view returns (uint256 stake) {
+        if (_pool.totalStake == 0) return 0;
         stake = MathUtils.percOf(_pool.totalStake, _pool.delegations[_delegator].shares, _pool.totalShares);
```
kyriediculous, line 164: This will lead to division by zero if `_pool.totalShares` is 0 as well so let's also add a conditional there 

```solidity
if (_pool.totalShares == 0) return 0; 
```

kyriediculous (changes requested):

contracts/bonding/BondingManager.sol
```diff
@@ -163,40 +198,46 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-     * @param _onBehalfOf The address which will own the stake
+     * @param _orchestrator The address of the transcoder to stake towards
+     * @param _for The address which will own the stake
      * @param _oldDelegateNewPosPrev The address of the previous transcoder in the pool for the old delegate
      * @param _oldDelegateNewPosNext The address of the next transcoder in the pool for the old delegate
      * @param _currDelegateNewPosPrev The address of the previous transcoder in the pool for the current delegate
      * @param _currDelegateNewPosNext The address of the next transcoder in the pool for the current delegate
      */
-    function bondWithHint(
+    function bond(
```
kyriediculous: It is convention to prefix internal functions with an underscore ; i.e `_bond`

Tip: another convention to avoid _variable_ naming collisions is adding a trailing underscore

contracts/bonding/BondingManager.sol
```diff
@@ -163,40 +198,46 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
         address _oldDelegateNewPosPrev,
         address _oldDelegateNewPosNext,
         address _currDelegateNewPosPrev,
         address _currDelegateNewPosNext
-    ) public whenSystemNotPaused currentRoundInitialized autoClaimFees(_to) {
-        if (_to != _onBehalfOf) {
-            require(!isRegisteredOrchestrator(_onBehalfOf), "ORCHESTRATOR_CAN_NOT_DELEGATE");
+    ) internal whenSystemNotPaused currentRoundInitialized autoClaimFees(_orchestrator) {
+        if (_orchestrator != _for) {
+            require(!isRegisteredOrchestrator(_for), "ORCHESTRATOR_CAN_NOT_DELEGATE");
```
kyriediculous, line 218: reminder to remove this when separating concerns

kyriediculous (changes requested):

kyriediculous (approved):

PR #469, opened 2021-08-20
sync next with streamflow

kyriediculous:
Sync the next branch with streamflow so it can be cleanly merged


kautukkundan (dismissed):
LGTM 👍🏼

yondonfu:
Is merging this as-is into `next` going to result in a bunch of commits that look like duplicates? i.e. typescript commit -> hardhat commit -> switch to GH action commit -> typescript commit -> hardhat commit.

From looking at the commit history on this branch it looks like it was rebased off of `streamflow`. When I used this approach locally and merged the resulting branch into `next` I saw the duplicates mentioned above. But, when I just pulled `streamflow` into this branch and merged it locally into `next` instead I just got an additional merge commit on `next`.

kyriediculous:
Went with the pull way , this seems to indeed have dragged the initial GH actions commit to behind the head of next and then tagged on the subsequent conflict resolution. Neat ! 👍

yondonfu (approved):
LGTM pending green CI!

PR #475, opened 2021-09-15
CI: Fix failing CI for next branch

kautukkundan:

updated contracts and test cases to fix failing CI for the `next` branch

- re-added deleted (deprecated) contracts `BondingManager.sol`, `IBondingManager.sol`, `EarningsPool.sol`, `EarningsPoolLIP36.sol`, `MathUtils.sol`, `MathUtilsV2.sol`, and `PreciseMathUtils.sol`
- changed `RoundsManager.sol` and `Minter.sol` to use old Math libraries (need to change back to precise in later iterations)
- pruned tests for removed contracts `BondingManagerZeroStartFactorsBug.sol`, `BondingManagerPreLIP36.sol`, `Refunder`, `TestMathUtilsV2 `, `TestPreciseMathUtils`, `TestEarningsPool`
- updated some tests to behave correctly with solidity version 0.8.4

ran `yarn test:coverage`

**Checklist:**


test/unit/VariableSupplyToken.js
```diff
@@ -3,12 +3,12 @@ import chai, {expect, assert} from "chai"
 import {solidity} from "ethereum-waffle"
 chai.use(solidity)
 
-describe("VariableSupplyToken", () => {
+describe("LivepeerToken", () => {
```
yondonfu, line 6: Is this change needed?

test/integration/PoolUpdatesWithHints.js
```diff
@@ -156,18 +156,19 @@ describe("PoolUpdatesWithHints", () => {
 
         // Get gas cost of transcoder()
         let tx = await bondingManager.connect(newTranscoder).transcoder(0, 0)
-        const txResNoHint = await tx.wait()
+        expect(tx).to.emit(bondingManager, "TranscoderDeactivated").withArgs(transcoders[size - 1].address, dr)
```
yondonfu, line 159: Why does moving the expect statement before `.wait()` fix the previous test failure caused by an assertion error with the "TranscoderDeactivated" event?

Does calling `.wait()` on `tx` before running the expect statement mean that the event is already emitted by the time the expect statement is run so the expect statement never resolves?

kautukkundan: Yeah, I also think the same. I guess the reason behind it also has something to do with how hardhat handles events, I found some info here - https://stackoverflow.com/a/69025256.

test/unit/VariableSupplyToken.js
kautukkundan: `VariableSupplyToken` was changed to abstract in the previous PR so need to use `LivepeerToken` now

yondonfu: Ah I see. In that case, can we change the name of this test file to `LivepeerToken.js` to reflect that the contract under test is the LivepeerToken contract?

test/integration/PoolUpdatesWithHints.js
yondonfu: My best guess is that placing `await tx.wait()` before the expect statement may give more time for the polling that happens under the hood to successfully find the event given whatever polling interval is used. And if `await tx.wait()` is after the expect statement there may be less time.

With that being said, I think the root of the problem may be the lack of await usage for the expect statement. You can observe that the error message is different for a failed assertion depending on whether await is used or not.

When you do not use await:

```
  PoolUpdatesWithHints
    ✓ initial transcoder pool order should be correct (107ms)
    ✓ transcoder calls reward with hint (1090ms)
AssertionError: expected '0xa0Ee7A142d267C1f36714E4a8F75612F20a79720' to equal '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266'
AssertionError: expected '0xa0Ee7A142d267C1f36714E4a8F75612F20a79720' to equal '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266'
    1) new transcoder joins the pool
    ✓ delegator bonds with hint (582ms)
    ✓ delegator changes delegation with hint (693ms)
    ✓ transcoder partially unbonds and rebonds (778ms)
    ✓ transcoder rebonds from unbonded (501ms)


  6 passing (13s)
  1 failing

  1) PoolUpdatesWithHints
       new transcoder joins the pool:
     Uncaught RuntimeError: abort(AssertionError: expected '0xa0Ee7A142d267C1f36714E4a8F75612F20a79720' to equal '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266'). Build with -s ASSERTIONS=1 for more info.
      at process.abort (node_modules/@ethereum-waffle/compiler/node_modules/solc/soljson.js:1:13938)
      at process.emit (node_modules/hardhat/node_modules/source-map-support/source-map-support.js:495:21)
      at process.emit (node_modules/@cspotcode/source-map-support/source-map-support.js:527:35)
      at processPromiseRejections (internal/process/promises.js:245:33)
      at processTicksAndRejections (internal/process/task_queues.js:96:32)
      at runNextTicks (internal/process/task_queues.js:64:3)
      at listOnTimeout (internal/timers.js:526:9)
      at processTimers (internal/timers.js:500:7)
```

When you use await:

```
1) PoolUpdatesWithHints
       new transcoder joins the pool:

      AssertionError: expected '0xa0Ee7A142d267C1f36714E4a8F75612F20a79720' to equal '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266'
      + expected - actual

      -0xa0Ee7A142d267C1f36714E4a8F75612F20a79720
      +0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266

      at assertArgsArraysEqual (node_modules/@ethereum-waffle/chai/dist/cjs/matchers/emit.js:58:54)
      at tryAssertArgsArraysEqual (node_modules/@ethereum-waffle/chai/dist/cjs/matchers/emit.js:65:20)
      at /Users/yondonfu/Development/livepeer/protocol/node_modules/@ethereum-waffle/chai/dist/cjs/matchers/emit.js:77:13
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at runNextTicks (internal/process/task_queues.js:64:3)
      at listOnTimeout (internal/timers.js:526:9)
      at processTimers (internal/timers.js:500:7)
```

Note that when await is not used we get an uncaught RuntimeError, but when await is used we get an AssertionError which is what we'd expect to see.

@kautukkundan We don't need to address all of this in this PR, but can you investigate whether we should be using await when running expect statements for event emits?

yondonfu (approved):

test/integration/PoolUpdatesWithHints.js
kautukkundan: opened an issue to track progress - livepeer/protocol/issues/476

PR #480, opened 2021-09-28
changed events testing pattern

### LINKED ISSUE #476: Investigate anomaly in test cases
based on https://github.com/livepeer/protocol/pull/475#discussion_r710460706 - testing events shows different behaviour when used with and without `await` keyword in some test cases. Need to verify the cause of this behaviour and its potential effect on other test cases in the suite.

kautukkundan:

The `expect.to.emit` accepts any kind of tx state as argument - tx promise, resolved tx, or even tx hash. [Internally](https://github.com/EthWorks/Waffle/blob/master/waffle-chai/src/matchers/misc/transaction.ts) it actually waits for the tx to resolve if not already resolved and uses the transaction hash to check for events (also eliminating the need to do tx.wait after the event check). 

In the [official waffle docs](https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#emitting-events) and also in it's [own test suite](https://github.com/EthWorks/Waffle/blob/f35373d822aa85bd99e08ec0bea367188b132433/waffle-chai/test/matchers/events.test.ts) the following pattern is used. 

```javascript
const tx = Contract.method() // tx promise i.e no await
await expect(tx).to.emit("Event")
```
`await` is used with `emit()` as it uses [promises inside itself](https://github.com/EthWorks/Waffle/blob/master/waffle-chai/src/matchers/emit.ts#L9-L18). For https://github.com/livepeer/protocol/pull/475#discussion_r710460706 I guess it is now okay to say that the issue was caused due to missing `await` before expect

other test suites in projects like [uniswap v3 tests](https://github.com/Uniswap/v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/test/UniswapV3Pool.spec.ts#L1203-L1205) and [graph protocol tests](https://github.com/graphprotocol/contracts/blob/f09f9dea65852dc3a74671135d3caedea2db024d/test/graphToken.test.ts#L123-L124) also follow the same pattern. 

The tests were updated to follow the same common pattern except in cases where the tx receipt is needed to get additional data like "cumulativeGasUsed" like [PoolUpdatesWithHints test case](https://github.com/livepeer/protocol/blob/kk/test-anomaly/test/integration/PoolUpdatesWithHints.js#L158-L162) and [PollCreator test case](https://github.com/livepeer/protocol/blob/kk/test-anomaly/test/unit/PollCreator.js#L57-L60) the transaction is resolved first, the tx receipt is extracted from the tx and then `tx.hash` is passed to the emit assertion.

closes #476 

**Checklist:**

### DIFF
--- a/test/integration/PoolUpdatesWithHints.js
+++ b/test/integration/PoolUpdatesWithHints.js
@@ -157,7 +157,7 @@ describe("PoolUpdatesWithHints", () => {
         // Get gas cost of transcoder()
         let tx = await bondingManager.connect(newTranscoder).transcoder(0, 0)
         const txResNoHint = await tx.wait()
-        expect(tx).to.emit(bondingManager, "TranscoderDeactivated").withArgs(transcoders[size - 1].address, dr)
+        await expect(tx.hash).to.emit(bondingManager, "TranscoderDeactivated").withArgs(transcoders[size - 1].address, dr)
 
         assert.equal(await transcoderAtPoolPos(size - 1), newTranscoder.address)
 
--- a/test/integration/PoolUpdatesWithHints.js
+++ b/test/integration/PoolUpdatesWithHints.js
@@ -156,16 +156,16 @@ describe("PoolUpdatesWithHints", () => {
 
         // Get gas cost of transcoder()
         let tx = await bondingManager.connect(newTranscoder).transcoder(0, 0)
+        const txResNoHint = await tx.wait()
         expect(tx).to.emit(bondingManager, "TranscoderDeactivated").withArgs(transcoders[size - 1].address, dr)
 
-        const txResNoHint = await tx.wait()
         assert.equal(await transcoderAtPoolPos(size - 1), newTranscoder.address)
 
         await rpc.revert(testSnapshotId)
 
         // Get gas cost of transcoderWithHint()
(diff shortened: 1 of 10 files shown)


test/integration/PoolUpdatesWithHints.js
```diff
@@ -156,16 +156,16 @@ describe("PoolUpdatesWithHints", () => {
 
         // Get gas cost of transcoder()
         let tx = await bondingManager.connect(newTranscoder).transcoder(0, 0)
+        const txResNoHint = await tx.wait()
         expect(tx).to.emit(bondingManager, "TranscoderDeactivated").withArgs(transcoders[size - 1].address, dr)
```
yondonfu: Should we use the `await expect(tx.hash)...` pattern here as well?

yondonfu (changes requested):
Nice finds with the event testing pattern and with the fact that we can also pass a transaction hash to `expect()`!

Just one question about an additional update.

test/integration/PoolUpdatesWithHints.js
kautukkundan: oops, missed this one! updated. Thanks

yondonfu (approved):
LGTM!

yondonfu (approved):
LGTM!

PR #496, opened 2021-12-08
L2 BondingManager with deprecated code removed

### LINKED ISSUE #491: LIP-73: Remove deprecated code from L2 BondingManager
There will be a fresh deployment of the BondingManager on L2 so any code that paths that depend on the activation of certain LIP upgrades and/or bug fixes on L1 can be removed because by default the BondingManager will contain all the updates from those LIPs and all of the previous bugs already fixed.

yondonfu:

Removes code from the BondingManager that is not needed for a fresh deployment on L2 including:

- Bug fix code
- Pre LIP-36 code
- LIP-52 code

Also, fixes some additional event tests in 8bb731d using the same event testing pattern that was introduced in https://github.com/livepeer/protocol/commit/6b402030394a7c9e3c641728812208b1c5b4921f. Prior to this commit, tests with coverage enabled would fail due to the same problem that the aforementioned commit fixed elsewhere.

See commit history.

Updated unit and integration tests.

Fixes #491 

**Checklist:**


kautukkundan (commented):
Looks good 👍🏼 Happy that contract size reduced

In addition to the removal of unused code, wdyt about removing MathUtils and using PreciseMathUtils everywhere too?

yondonfu:
> wdyt about removing MathUtils and using PreciseMathUtils everywhere too?

I was planning to take a pass at cleaning up the usage of MathUtils/PreciseMathUtils libraries separately (can create an issue) because at the moment commission rates are defined as X out of MathUtils.PERC_DIVISOR (1000000) and there is off-chain code (i.e. the explorer and go-livepeer) that assume commission rates are out of that value as well. So, if we're going to change how commission rates are defined then we'll need to make sure to make the necessary updates in all of those off-chain codebases as well. In the interest of doing all of those things at once to make sure no individual tool breaks and of not increasing the scope of this PR for now I suggest handling the usage cleanup separately.

kautukkundan (approved):
Let's track the preciseMath issue and get this merged in 👍🏼

yondonfu:
> Let's track the preciseMath issue and get this merged in

Tracking in https://github.com/livepeer/protocol/issues/506

PR #497, opened 2021-12-12
Added bondForWithHint() function

### LINKED ISSUE #488: LIP-73: bondForWithHint() in L2 BondingManager
See `bondForWithHint()` description for the L2 BondingManager in LIP-73.

kautukkundan:

- renamed `bondWithHint` to `bondForWithHint` function which accepts `_owner`
- added `bondWithHint` function to call `bondForWithHint` with _owner = msg.sender

closes #488 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -458,20 +458,22 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * is the correct position so no search is executed. See SortedDoublyLL.sol for details on list hints
      * @param _amount The amount of tokens to stake.
      * @param _to The address of the transcoder to stake towards
+     * @param _owner The address of the owner of the bond
      * @param _oldDelegateNewPosPrev The address of the previous transcoder in the pool for the old delegate
      * @param _oldDelegateNewPosNext The address of the next transcoder in the pool for the old delegate
      * @param _currDelegateNewPosPrev The address of the previous transcoder in the pool for the current delegate
      * @param _currDelegateNewPosNext The address of the next transcoder in the pool for the current delegate
      */
-    function bondWithHint(
+    function bondForWithHint(
         uint256 _amount,
+        address _owner,
         address _to,
         address _oldDelegateNewPosPrev,
         address _oldDelegateNewPosNext,
         address _currDelegateNewPosPrev,
         address _currDelegateNewPosNext
     ) public whenSystemNotPaused currentRoundInitialized autoClaimEarnings {
-        Delegator storage del = delegators[msg.sender];
+        Delegator storage del = delegators[_owner];
 
         uint256 currentRound = roundsManager().currentRound();
         // Amount to delegate
@@ -479,7 +481,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Current delegate
         address currentDelegate = del.delegateAddress;
 
-        if (delegatorStatus(msg.sender) == DelegatorStatus.Unbonded) {
+        if (delegatorStatus(_owner) == DelegatorStatus.Unbonded) {
             // New delegate
             // Set start round
             // Don't set start round if delegator is in pending state because the start round would not change
@@ -492,10 +494,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             // In the future, if delegation towards another registered transcoder as an already
             // registered transcoder becomes useful (i.e. for transitive delegation), this restriction
             // could be removed
-            require(
-                !isRegisteredTranscoder(msg.sender),
-                "registered transcoders can't delegate towards other addresses"
-            );
+            require(!isRegisteredTranscoder(_owner), "registered transcoders can't delegate towards other addresses");
             // Changing delegate
             // Set start round
             del.startRound = currentRound.add(1);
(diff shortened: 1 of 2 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -527,10 +528,29 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
         if (_amount > 0) {
             // Transfer the LPT to the Minter
-            livepeerToken().transferFrom(msg.sender, address(minter()), _amount);
+            livepeerToken().transferFrom(_owner, address(minter()), _amount);
```
yondonfu: We should be using `msg.sender` instead of `_owner` here since tokens are pulled from `msg.sender` and staked for `_owner`. Otherwise, the tokens would need to be pulled from `_owner` which doesn't work for the use case where a contract holds LPT and is staking it on behalf of another user (i.e. what the L2Migrator is supposed to do).

contracts/bonding/BondingManager.sol
```diff
@@ -505,15 +504,17 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
-        Transcoder storage newDelegate = transcoders[_to];
-        EarningsPool.Data storage currPool = newDelegate.earningsPoolPerRound[currentRound];
-        if (currPool.cumulativeRewardFactor == 0) {
-            currPool.cumulativeRewardFactor = cumulativeFactorsPool(newDelegate, newDelegate.lastRewardRound)
-                .cumulativeRewardFactor;
-        }
-        if (currPool.cumulativeFeeFactor == 0) {
-            currPool.cumulativeFeeFactor = cumulativeFactorsPool(newDelegate, newDelegate.lastFeeRound)
-                .cumulativeFeeFactor;
+        {
```
yondonfu, line 508: Do we need these brackets around L508-517?

test/unit/BondingManager.js
```diff
@@ -513,6 +513,24 @@ describe("BondingManager", () => {
                 )
             })
 
+            it("should set delegate for different address", async () => {
```
yondonfu: I think it would be clearer if the test cases specific to `bondForWithHint()` are grouped under their own describe block. This would make it easier to review testing of functionality that is specific to `bondForWithHint()` (i.e. the new cases added thus far here). WDYT?

yondonfu (changes requested):

yondonfu:
Note: I suspect the CI test failure will be fixed once this branch is rebased off of #496 to pick up [this commit](https://github.com/livepeer/protocol/pull/496/commits/8bb731d7b0381092f2e0c6ddc9d1bb8a344481cf).

contracts/bonding/BondingManager.sol
```diff
@@ -527,10 +528,29 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
             // Transfer the LPT to the Minter
-            livepeerToken().transferFrom(msg.sender, address(minter()), _amount);
+            livepeerToken().transferFrom(_owner, address(minter()), _amount);
         }
 
-        emit Bond(_to, currentDelegate, msg.sender, _amount, del.bondedAmount);
+        emit Bond(_to, currentDelegate, _owner, _amount, del.bondedAmount);
+    }
+
+    function bondWithHint(
```
yondonfu, line 553: Let's keep a Natspec comment for this one too?

contracts/bonding/BondingManager.sol
kautukkundan: Added to avoid "Stack too deep" error by reducing the local variables in the function scope.

test/unit/BondingManager.js
```diff
@@ -998,6 +998,109 @@ describe("BondingManager", () => {
         })
     })
 
+    describe("bondFor", () => {
```
yondonfu:
```suggestion
    describe("bondForWithHint", () => {
```

test/unit/BondingManager.js
```diff
@@ -998,6 +998,109 @@ describe("BondingManager", () => {
...
+
+                assert.equal(
+                    (await bondingManager.getDelegator(delegator2.address))[2],
+                    transcoder0.address,
+                    "wrong delegateAddress"
+                )
+            })
+
+            it("should update delegate and bonded amount for a different address", async () => {
+                const startDelegatedAmount = (await bondingManager.getDelegator(transcoder0.address))[3]
```
yondonfu: Does `.delegatedAmount` work here since the ABI should include names for each of the fields in the returned tuple?

test/unit/BondingManager.js
```diff
@@ -998,6 +998,109 @@ describe("BondingManager", () => {
...
+                    delegator2.address,
+                    transcoder0.address,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero
+                )
+
+                assert.equal(
+                    (await bondingManager.getDelegator(delegator2.address))[2],
```
yondonfu: Does `.delegateAddress` work here?

test/unit/BondingManager.js
```diff
@@ -998,6 +998,109 @@ describe("BondingManager", () => {
...
+                    delegator2.address,
+                    transcoder0.address,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero,
+                    ethers.constants.AddressZero
+                )
+                const endDelegatedAmount = (await bondingManager.getDelegator(transcoder0.address))[3]
+
+                assert.equal(endDelegatedAmount.sub(startDelegatedAmount), 1000, "wrong change in delegatedAmount")
```
yondonfu: Thoughts on using `expect()` consistently going forward in new tests instead of `assert()`?

yondonfu (changes requested):
A few questions about possibility of using named fields of tuples instead of indexing into the tuple.

test/unit/BondingManager.js
kautukkundan: yeah it does. TS + typechain has spoiled me for expecting autocomplete everwhere.
Thanks for the suggestion 👍🏼

yondonfu (approved):
LGTM

I recommend squashing the commits for this particular PR.

PR #498, opened 2021-12-13
added transferBond() function

### LINKED ISSUE #494: LIP-73: L2 BondingManager transferBond()
See `transferBond()` description in https://www.notion.so/livepeer/L2-Migration-w-Orchestrator-Migration-Delegator-L2-Claiming-7adb658c8f0d4d129096046dca2b827f

kautukkundan:

- added transferBond function
- updated rebondWithHint function
- added test cases

closes #494 

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -460,6 +460,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @param _amount The amount of tokens to stake.
      * @param _owner The address of the owner of the bond
      * @param _to The address of the transcoder to stake towards
+     * @param _owner The address of the owner of the bond
      * @param _oldDelegateNewPosPrev The address of the previous transcoder in the pool for the old delegate
      * @param _oldDelegateNewPosNext The address of the next transcoder in the pool for the old delegate
      * @param _currDelegateNewPosPrev The address of the previous transcoder in the pool for the current delegate
@@ -529,7 +530,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
         if (_amount > 0) {
             // Transfer the LPT to the Minter
-            livepeerToken().transferFrom(msg.sender, address(minter()), _amount);
+            livepeerToken().transferFrom(_owner, address(minter()), _amount);
         }
 
         emit Bond(_to, currentDelegate, _owner, _amount, del.bondedAmount);
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -569,6 +569,38 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         );
     }
 
+    function transferBond(address _delegator, uint256 _amount)
+        public
+        whenSystemNotPaused
+        currentRoundInitialized
+        autoClaimEarnings
+    {
+        unbondWithHint(_amount, address(0), address(0));
+
+        Delegator storage oldDel = delegators[msg.sender];
+        Delegator storage newDel = delegators[_delegator];
+
+        uint256 oldDelUnbondingLockId = oldDel.nextUnbondingLockId.sub(1);
+        uint256 withdrawRound = oldDel.unbondingLocks[oldDelUnbondingLockId].withdrawRound;
+
+        // Burn lock for current owner
+        delete oldDel.unbondingLocks[oldDelUnbondingLockId];
+
+        // Create lock for new owner
+        uint256 newDelUnbondingLockId = newDel.nextUnbondingLockId;
+
+        newDel.unbondingLocks[newDelUnbondingLockId] = UnbondingLock({ amount: _amount, withdrawRound: withdrawRound });
+        newDel.nextUnbondingLockId.add(1);
+
+        // Rebond lock for new owner
+        address delegate = newDel.delegateAddress;
+        if (delegate == address(0)) {
+            delegate = oldDel.delegateAddress;
+        }
+
+        _rebondWithHint(_delegator, newDelUnbondingLockId, address(0), address(0));
+    }
+
     /**
      * @notice Unbond an amount of the delegator's bonded stake and updates the transcoder pool using an optional list hint if needed
      * @dev If the caller remains in the transcoder pool, the caller can provide an optional hint for its insertion position in the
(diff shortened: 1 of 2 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -553,6 +553,37 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         );
     }
 
+    function transferBond(address _delegator, uint256 _amount)
```
yondonfu: WDYT about adding hints to this function?

Ex.

```
function transferBond(
    address _delegator,
    uint256 _amount,
    address _oldDelegateNewPosPrev,
    address _oldDelegateNewPosNext,
    address _newDelegateNewPosPrev,
    address _newDelegateNewPosNext
)
```

where the `oldDel_` hints can be passed to `unbondWithHint()` for the msg.sender delegate and the `newDel_` hints can be passed to `_rebondWithHint()` for the `_delegator` delegate.

yondonfu: Let's add a Natspec comment?

contracts/bonding/BondingManager.sol
```diff
@@ -603,6 +634,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         emit Unbond(currentDelegate, msg.sender, unbondingLockId, _amount, withdrawRound);
     }
 
+    function rebondWithHint(
```
yondonfu: Let's move the original Natspec comment here as well.

test/unit/BondingManager.js
```diff
@@ -639,6 +639,93 @@ describe("BondingManager", () => {
             beforeEach(async () => {
                 await bondingManager.connect(delegator).bond(2000, transcoder0.address)
             })
+
+            describe("caller is transferring bond to another address", () => {
```
yondonfu: I think it would be clearer if the test cases specific to `transferBond()` are grouped under their own describe block. This would make it easier to review testing of functionality that is specific to `transferBond()` (i.e. the new cases added thus far here). WDYT?

yondonfu (changes requested):

contracts/bonding/BondingManager.sol
```diff
@@ -613,15 +654,14 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
         uint256 _unbondingLockId,
         address _newPosPrev,
         address _newPosNext
-    ) public whenSystemNotPaused currentRoundInitialized autoClaimEarnings {
-        require(delegatorStatus(msg.sender) != DelegatorStatus.Unbonded, "caller must be bonded");
-
+    ) internal {
         // Process rebond using unbonding lock
-        processRebond(msg.sender, _unbondingLockId, _newPosPrev, _newPosNext);
+        processRebond(_delegator, _unbondingLockId, _newPosPrev, _newPosNext);
```
kautukkundan, line 664: WDYT about removing this internal function `_rebondWithHint` entirely and calling `processRebond` inside `rebondWithHint` and `transferBond` directly as all it does is call `processBond` anyway

kautukkundan:
e0914c7 removes `_rebondWithHint` internal function and calls `_processRebond` directly inside `rebondWithHint` and `transferBond` as all the internal function did was call `_processRebond` anyway

contracts/bonding/BondingManager.sol
```diff
@@ -1386,7 +1440,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @param _newPosPrev Address of previous transcoder in pool if the delegate is already in or joins the pool
      * @param _newPosNext Address of next transcoder in pool if the delegate is already in or joins the pool
      */
-    function processRebond(
+    function _processRebond(
```
yondonfu: Thoughts on omitting the underscore for now to minimize the diff given that we still have a lot of internal functions in this contract that do not follow that convention (the ones that do only do so to avoid a naming collision)?

test/unit/BondingManager.js
```diff
@@ -1695,6 +1696,139 @@ describe("BondingManager", () => {
...
+                await bondingManager.connect(delegator).bond(2000, transcoder0.address)
+                await fixture.roundsManager.setMockUint256(functionSig("currentRound()"), currentRound + 2)
+            })
+
+            describe("receiver is not bonded", () => {
+                it("should transfer the bond to receiver", async () => {
+                    const startDelegatedAmountD1 = (await bondingManager.getDelegator(delegator.address))[0]
+                    const startDelegatedAmountD2 = (await bondingManager.getDelegator(delegator2.address))[0]
+
+                    expect(startDelegatedAmountD1).to.equal(2000)
```
yondonfu: Since we're checking `bondedAmount` suggest using that terminology here to avoid confusion with the `delegatedAmount` field which is different.

test/unit/BondingManager.js
```diff
@@ -1695,6 +1696,139 @@ describe("BondingManager", () => {
...
+                    const endDelegatedAmountD1 = (await bondingManager.getDelegator(delegator.address))[0]
+                    const endDelegatedAmountD2 = (await bondingManager.getDelegator(delegator2.address))[0]
+
+                    const delegateD1 = (await bondingManager.getDelegator(delegator.address))[2]
+                    const delegateD2 = (await bondingManager.getDelegator(delegator2.address))[2]
+
+                    expect(endDelegatedAmountD1).to.equal(200)
+                    expect(endDelegatedAmountD2).to.equal(3800)
+
+                    expect(delegateD1).to.equal(transcoder0.address)
```
yondonfu: Check that `delegatedAmount` of the delegate did not change?

test/unit/BondingManager.js
```diff
@@ -1695,6 +1696,139 @@ describe("BondingManager", () => {
...
```
yondonfu: Check that delegatedAmount of each delegate was updated correctly?

yondonfu (changes requested):
Are we able to use named fields for returned tuples instead of indexing into the tuple?

test/unit/BondingManager.js
kautukkundan: oops 🤦🏻‍♂️ named fields are indeed helpful

yondonfu (approved):
LGTM

I recommend squashing the commits for this particular PR.

Issue #495, opened 2021-12-07
LIP-73: L2 ClaimableDelegatorPool


yondonfu:
See ClaimableDelegatorPool description in https://www.notion.so/livepeer/L2-Migration-w-Orchestrator-Migration-Delegator-L2-Claiming-7adb658c8f0d4d129096046dca2b827f

yondonfu:
In https://github.com/livepeer/arbitrum-lpt-bridge/pull/30 the L2Migrator clones the delegator pool implementation to create a minimal proxy copy. However, it does not initialize the state of the contract. If we need to initialize the state of the contract (i.e. so it is aware of the total stake that it started with in order to calculate the stake/fees owed to delegators) then we'll need to update the L2Migrator to call an initialization function on the contract.

PR #511, opened 2021-12-28
bonding: Support address + amount params in withdrawFees

### LINKED ISSUE #503: Confluence: Add support for withdrawing fees to specified address in BondingManager
At the moment, `withdrawFees()` sends earned fees to msg.sender.

We can add support for passing an address so that earned fees are withdrawn to the specified address instead of msg.sender

This is required for the integration with the ClaimableDelegatorPool contract in #495.

Should go to `confluence` branch.

yondonfu:

This PR adds support for specifying a recipient address and amount in `withdrawFees()` in the BondingManager. This feature is required for https://github.com/livepeer/protocol/issues/495.

This is a breaking change to the `withdrawFees()` function and `WithdrawFees()` event API and will require the following updates:

- go-livepeer calls to `withdrawFees()`
- explorer calls to `withdrawFees()`
- subgraph indexing of the `WithdrawFees()` event

The rationale for this breaking change:

- The BondingManager is already really large
- The update to support this breaking change is small

May want to consider updating slimming down the BondingManager and TicketBroker further by removing `*for*` functions in favor of accepting args, but we can consider that after completing the current set of tasks to get to feature complete milestone.

See commit history.

Updated tests.

Fixes #503 

**Checklist:**


kautukkundan:
How about we add a test case validating this order of events?
1. claims fees partially
2. more fees is rewarded
3. claims again (fully or partially)

and check if the saved/claimed fees is correct.

yondonfu:
> How about we add a test case validating this order of events?

Added the following test cases in dd72a12:

- Full withdrawal (to complement the existing partial withdrawal test case)
- Withdrawal after earning more fees

Will rebase to resolve conflicts after the above commit is reviewed.

kautukkundan (approved):
LGTM 👍🏼

PR #515, opened 2021-12-30
Add BridgeMinter for funds migration and bridge minting

### LINKED ISSUE #493: LIP-73: L1 Minter support for transferring LPT and ETH across bridge to L2 Migrator

yondonfu:

This PR adds a BridgeMinter contract which will be the new Minter on L1. The existing Minter can migrate to the BridgeMinter via a call to Minter.migrateToNewMinter().

The BridgeMinter supports:

- Withdrawing LPT to the L1Migrator so that the LPT can be migrated to L2
- Withdrawing ETH to the L1Migrator so that the ETH can be migrated to L2
- Minting LPT on L1 to recipients based on calls from the L1LPTGateway which is responsible for the LPT bridging workflow between L1 and L2. This will only need to happen if the L1Escrow is depleted meaning that LPT that was minted on L2 for inflation is now being transferred to L1
- An updated migrateToNewMinter() implementation that still transfers ownership of the L1 LPT to a new Minter address as well as any LPT/ETH that the BridgeMinter currently holds. A few checks have been removed from the existing implementation from the existing Minter:
    - No longer require the Controller to be paused as this feels overly restrictive in the first place since only the Controller owner can call this function anyway
    - No longer require the new Minter to expose a getController() function as this feels overly restrictive
    - No longer require the new Minter's Controller to have the existing Minter to be registered as this feels overly restrictive

See commit history.

Updated unit tests. Also added an integration test for upgrading/migrating the existing Minter to the BridgeMinter.

Partially fixes #493 

**Checklist:**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,7 @@ module.exports = {
         "quotes": ["error", "double"],
         "comma-dangle": ["error", "never"],
         "indent": ["error", 4],
-        "space-infix-ops": ["error"]
+        "space-infix-ops": ["error"],
+        "camelcase": "off"
     }
 }
--- a/.solcover.js
+++ b/.solcover.js
@@ -9,5 +9,6 @@ module.exports = {
         "rounds/AdjustableRoundsManager.sol",
         "pm/mixins/interfaces",
         "bonding/deprecated",
+        "token/ArbitrumLivepeerToken.sol" // testnet only
     ],
 };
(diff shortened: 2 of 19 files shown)


yondonfu:
Rebased off the latest `streamflow` to resolve conflicts and added the following commits:

- 2c7d7d3 adds a missing await for an event test
- 0c12190 fixes the Solidity test runner which also fixes the failing tests in CI. I believe the problem was that the Solidity test runner was previously introducing describe blocks with async function calls outside of hooks and it blocks. Since that test runner was pulled from https://github.com/aragon/aragonOS/blob/next/test/helpers/runSolidityTest.js I just updated the code to reflect the latest changes there which I believe fixes this issue

kautukkundan (dismissed):
Looks good

PR #520, opened 2022-01-07
polling: Use min stake in PollCreator

### LINKED ISSUE #508: Confluence: Switch to minimum stake req in PollCreator
Since we're disabling permissionless burning of LPT on L2, instead of burning the poll creation cost can just use a minimum stake req.

yondonfu:

This PR updates the PollCreator to require a minimum stake >= POLL_CREATION_COST from the caller instead of requiring the caller to burn the POLL_CREATION_COST. The caller can meet the minimum stake req with its own stake (i.e. BondingManager.pendingStake()) *or* its delegated stake (i.e. BondingManager.transcoderTotalStake()).

Since the PollCreator.js unit test was pretty small I updated it to use smock for mocking instead of the GenericMock contract since smock is easier to use for mocking return values.

See commit history.

Updated unit tests.

Fixes #508 

**Checklist:**

### DIFF
--- a/contracts/polling/PollCreator.sol
+++ b/contracts/polling/PollCreator.sol
@@ -30,15 +30,15 @@ contract PollCreator {
      * @param _proposal The IPFS multihash for the proposal.
      */
     function createPoll(bytes calldata _proposal) external {
-        uint256 endBlock = block.number + POLL_PERIOD;
-        Poll poll = new Poll(endBlock);
-
         require(
             // pendingStake() ignores the second arg
             bondingManager.pendingStake(msg.sender, 0) >= POLL_CREATION_COST ||
                 bondingManager.transcoderTotalStake(msg.sender) >= POLL_CREATION_COST,
             "PollCreator#createPoll: INSUFFICIENT_STAKE"
         );
+
+        uint256 endBlock = block.number + POLL_PERIOD;
+        Poll poll = new Poll(endBlock);
 
         emit PollCreated(address(poll), _proposal, endBlock, QUORUM, QUOTA);
     }
--- a/contracts/polling/PollCreator.sol
+++ b/contracts/polling/PollCreator.sol
@@ -1,7 +1,12 @@
 pragma solidity ^0.5.11;
 
 import "./Poll.sol";
-import "../token/ILivepeerToken.sol";
+
+interface IBondingManager {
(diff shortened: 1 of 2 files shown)


contracts/polling/PollCreator.sol
```diff
@@ -12,26 +17,28 @@ contract PollCreator {
...
 
-        require(token.transferFrom(msg.sender, address(this), POLL_CREATION_COST), "LivepeerToken transferFrom failed");
-
-        token.burn(POLL_CREATION_COST);
+        require(
+            // pendingStake() ignores the second arg
+            bondingManager.pendingStake(msg.sender, 0) >= POLL_CREATION_COST ||
+                bondingManager.transcoderTotalStake(msg.sender) >= POLL_CREATION_COST,
+            "PollCreator#createPoll: INSUFFICIENT_STAKE"
+        );
```
kautukkundan, line 38: should move this require statement before contract deployment to considerably reduce consumed gas in case of a failure/revert.

test/unit/PollCreator.js
```diff
@@ -46,27 +60,97 @@ describe("PollCreator", () => {
...
+
             const start = await fixture.rpc.getBlockNumberAsync()
             const end = start + POLL_PERIOD + 1 // + 1 because createPoll tx will mine a new block
-            const tx = await pollCreator.createPoll(hash)
-            const receipt = await tx.wait()
-            await expect(tx.hash)
+
+            let tx = await pollCreator.createPoll(hash)
+            let receipt = await tx.wait()
+            await expect(tx)
```
kautukkundan, line 105: #minor
Since the transaction is already mined, using tx.hash in expect will be tad bit faster than passing the whole tx, as [internally](https://github.com/EthWorks/Waffle/blob/master/waffle-chai/src/matchers/misc/transaction.ts#L7-L8) it will add 1 layer of extra promise resolution to get the tx hash. 
Can skip this edit for now.

kautukkundan (changes requested):
Should be good to go after gas optimization in the contract 👍🏼

contracts/polling/PollCreator.sol
yondonfu: Fixed in c3806a6

kautukkundan (approved):
Looks Good 👍🏼

PR #530, opened 2022-01-26
Transfer current tokens to escrow when setting new token in bridgeMinter

kautukkundan:

- transfers tokens held by BridgeMinter to a specified address before setting a new token

see commit history

**Checklist:**

### DIFF
--- a/contracts/token/BridgeMinter.sol
+++ b/contracts/token/BridgeMinter.sol
@@ -15,7 +15,6 @@ interface IBridgeMinterToken {
 }
 
 contract BridgeMinter is Manager {
-    address public tokenAddr;
     address public l1MigratorAddr;
     address public l1LPTGatewayAddr;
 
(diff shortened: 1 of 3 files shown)


yondonfu (changes requested):
Instead of having a `setToken()` function for changing the token at all I suggest:

- Using a [livepeerToken()](https://github.com/livepeer/protocol/blob/20e7ebb86cdb4fe9285bf5fea02eb603e5d48805/contracts/token/Minter.sol#L247) function to fetch the LPT address from the Controller
- `migrateToNewMinter()` will transfer the contract's balance to a new contract if needed so in the unlikely event that the LPT address returned by the Controller ever changes it would just be important to call `migrateToNewMinter()` first to move the contract's balance for the existing LPT address

This avoids the need for an explicit setter and dealing with the additional logic of transferring tokens in that setter right now.

The above does involve extra gas usage due to the external call to the Controller, but since the operations that involve making calls to the LPT contract are infrequent I don't think its that big of a deal.

contracts/token/BridgeMinter.sol
```diff
@@ -135,4 +124,11 @@ contract BridgeMinter is Manager {
     function getController() public view returns (address) {
         return address(controller);
     }
+
+    /**
+     * @dev Returns IBridgeMinterToken interface
+     */
+    function livepeerToken() private view returns (IBridgeMinterToken) {
+        return IBridgeMinterToken(IController(controller).getContract(keccak256("LivepeerToken")));
```
yondonfu:
```suggestion
        return IBridgeMinterToken(controller.getContract(keccak256("LivepeerToken")));
```

`controller` should already be a `IController` type.

yondonfu (changes requested):
Looks good! Just one suggested change.

contracts/token/BridgeMinter.sol
kautukkundan: fixed in 0e00833

kautukkundan:
rebased from target

yondonfu (approved):
LGTM. Let's squash and merge using a single commit that still references the relevant finding.

PR #533, opened 2022-01-29
streamflow: Audit gas optimizations

yondonfu:

Implement some of the gas optimizations suggested from the audit for the L1 BridgeMinter.

The first commit bumps hardhat to v2.8.3 in order to support solc v0.8.8. The second commit includes links to the relevant gas optimization issues from the audit.

Ran the tests.

N/A

**Checklist:**

### DIFF
--- a/contracts/solc-0.8.8/Manager.sol
+++ b/contracts/solc-0.8.8/Manager.sol
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+// solhint-disable-next-line
+pragma solidity 0.8.8;
+
+import "./IManager.sol";
+import "./IController.sol";
+
+contract Manager is IManager {
+    // Controller that contract is registered with
+    IController public controller;
+
+    // Check if sender is controller
+    modifier onlyController() {
+        _onlyController();
+        _;
+    }
+
+    // Check if sender is controller owner
+    modifier onlyControllerOwner() {
+        _onlyControllerOwner();
+        _;
+    }
+
+    constructor(address _controller) {
+        controller = IController(_controller);
+    }
+
+    /**
+     * @notice Set controller. Only callable by current controller
+     * @param _controller Controller contract address
+     */
+    function setController(address _controller) external onlyController {
+        controller = IController(_controller);
+
+        emit SetController(_controller);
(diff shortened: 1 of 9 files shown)


contracts/solc-0.8.8/Manager.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+// solhint-disable-next-line
+pragma solidity 0.8.8;
+
+import "./IManager.sol";
+import "./IController.sol";
+
+contract Manager is IManager {
```
yondonfu, line 8: Note that the following functions were dropped from the original [Manager](https://github.com/livepeer/protocol/blob/streamflow/contracts/Manager.sol):

- whenSystemNotPaused
- whenSystemPaused

because they are not needed in BridgeMinter.

yondonfu:
https://github.com/livepeer/protocol/pull/533/commits/e6f0b16336b32c5c012dc9fe627d399f2b876f8e is a cherry pick of https://github.com/livepeer/protocol/commit/19ab45be4df2fb71df1e59bd5ac9582bc58c0ec0 from `confluence`.

hardhat.config.ts
```diff
@@ -28,8 +28,28 @@ const config: HardhatUserConfig = {
...
+                settings: {
+                    optimizer: {
+                        enabled: true,
+                        runs: 200
+                    }
+                }
             }
-        ]
+        ],
+        overrides: {
```
kautukkundan: I don't think override is required for individual files (if the settings params remain same). Hardhat picks the target solc version from the pragma statement and compiles the contract accordingly. No harm in having this though.

kautukkundan (dismissed):
Looks Good

hardhat.config.ts
yondonfu: Yep you're right. The `overrides` configuration is only required if a) the pragma statement for a file allows a file to be compiled with a range of solc versions b) > 1 solc version specified under `compilers` could be used for the file and c) you want to use a specific solc version that is not highest version listed under `compilers`.

I ended up dropping the `overrides` configuration in the interest of minimizing config clutter.

PR #537, opened 2022-02-02
Solidity version bump to 0.8.8

kautukkundan:

upgrades solidity version to 0.8.8 for all the protocol contracts and test/mock contracts. Also edited a few tests in case the something changed. For SafeMath, I simply included an upgraded version of the library instead of using the native solidity math features. I felt that this was out of the scope of this PR and if needed could be done separately.

I have tried to combine related contracts together in a single commit for ease in review process and I suggest going over commits individually. 

Most code changes are referenced from an earlier exercise at https://github.com/livepeer/protocol/pull/465

- check commit history

related to https://github.com/code-423n4/2022-01-livepeer-findings/issues/10

**Checklist:**


contracts/bonding/BondingManager.sol
```diff
@@ -258,7 +259,7 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         address _transcoder,
         uint256 _fees,
         uint256 _round
-    ) external whenSystemNotPaused onlyTicketBroker {
+    ) external override whenSystemNotPaused onlyTicketBroker {
```
yondonfu: Why does there need to be an `override` here? AFAIK as of solc 0.8.8 contracts that inherit from interfaces do not need to specify the `override` keyword when overriding functions defined in the interface.

contracts/IController.sol
```diff
@@ -1,17 +1,18 @@
...
         address _contractAddress,
         bytes20 _gitCommitHash
-    ) external;
+    ) external virtual;
 
-    function updateController(bytes32 _id, address _controller) external;
+    function updateController(bytes32 _id, address _controller) external virtual;
 
-    function getContract(bytes32 _id) public view returns (address);
+    function getContract(bytes32 _id) external view virtual returns (address);
```
yondonfu: Could the visibility of `getContract()` stay as `public` here since the implementation in `Controller` is public?

contracts/pm/TicketBroker.sol
```diff
@@ -11,8 +10,8 @@ import "./mixins/MixinWrappers.sol";
 contract TicketBroker is
     MixinContractRegistry,
     MixinReserve,
-    MixinTicketBrokerCore,
     MixinTicketProcessor,
+    MixinTicketBrokerCore,
```
yondonfu: Does the ordering matter here?

contracts/pm/mixins/MixinContractRegistry.sol
```diff
@@ -1,9 +1,12 @@
...
+pragma solidity 0.8.8;
 
 import "../../ManagerProxyTarget.sol";
-import "./interfaces/MContractRegistry.sol";
+import "../../bonding/IBondingManager.sol";
+import "../../token/IMinter.sol";
+import "../../rounds/IRoundsManager.sol";
 
-contract MixinContractRegistry is MContractRegistry, ManagerProxyTarget {
+contract MixinContractRegistry is ManagerProxyTarget {
```
yondonfu: If we're keeping around MContractRegistry anyway could we just inherit from that abstract contract which already imports the interfaces above which would reduce the diff here?

contracts/pm/mixins/MixinReserve.sol
```diff
@@ -1,29 +1,22 @@
...
+pragma solidity 0.8.8;
 
+import "./MixinContractRegistry.sol";
 import "./interfaces/MReserve.sol";
-import "./interfaces/MContractRegistry.sol";
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 
-contract MixinReserve is MContractRegistry, MReserve {
+abstract contract MixinReserve is MixinContractRegistry, MReserve {
```
yondonfu, line 8: Thoughts on continuing to inherit from MContractRegistry instead so this abstract contract doesn't need to be aware of the impl of the functions in MixinContractRegistry and those implementations would just be mixed in via the MixinContractRegistry mixin in TicketBroker? Mainly interested in whether we can minimize the diff here if this change is not needed.

contracts/pm/mixins/MixinTicketBrokerCore.sol
```diff
@@ -1,15 +1,15 @@
...
-import "./interfaces/MContractRegistry.sol";
-import "openzeppelin-solidity/contracts/cryptography/ECDSA.sol";
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "./MixinContractRegistry.sol";
 
-contract MixinTicketBrokerCore is MContractRegistry, MReserve, MTicketProcessor, MTicketBrokerCore {
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+
+abstract contract MixinTicketBrokerCore is MixinReserve, MixinTicketProcessor, MTicketBrokerCore {
```
yondonfu: Thoughts on continuing to inherit from MTicketProcessor instead so this abstract contract doesn't need to be aware of the impl of the functions in MixinTicketProcessor and those implementations would just be mixed in via the MixinTicketProcessor mixin in TicketBroker? Mainly interested in whether we can minimize the diff here if this change is not needed.

contracts/pm/mixins/MixinTicketProcessor.sol
```diff
@@ -1,10 +1,10 @@
...
+pragma solidity 0.8.8;
 
-import "./interfaces/MTicketProcessor.sol";
-import "./interfaces/MContractRegistry.sol";
-import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "./MixinContractRegistry.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 
-contract MixinTicketProcessor is MContractRegistry, MTicketProcessor {
+abstract contract MixinTicketProcessor is MixinContractRegistry {
```
yondonfu: Same comment here about inheriting from MContractRegistry & MTicketProcessor as the other comments about inheriting from the M prefixed contracts instead of the Mixin prefixed contracts with concrete implementations.

contracts/pm/mixins/MixinWrappers.sol
```diff
@@ -1,11 +1,9 @@
...
-pragma experimental ABIEncoderV2;
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.8;
 
-import "./interfaces/MTicketBrokerCore.sol";
-import "./interfaces/MContractRegistry.sol";
+import "./MixinTicketBrokerCore.sol";
 
-contract MixinWrappers is MContractRegistry, MTicketBrokerCore {
+abstract contract MixinWrappers is MixinTicketBrokerCore {
```
yondonfu: Same comment here about inheriting from MContractRegistry & MTicketBrokerCore as the other comments about inheriting from the M prefixed contracts instead of the Mixin prefixed contracts with concrete implementations.

yondonfu (changes requested):
The commit structure was *very* helpful for review - thanks!

Overall looks solid and just left a few small comments.

A note about the current mixin contract architecture used for the TicketBroker since there are a few comments around the changes there. Originally the thought process behind the mixin contract architecture was that it would be helpful in breaking up the functionality of a large contract into smaller contracts. The architecture ended up with the some of the smaller mixin contracts calling functions defined in other mixin contracts so there was an attempt to decouple things by having an individual mixin contract only inherit an "interface mixin contract" (i.e. the M prefixed contracts) to avoid every mixin contract needing to be aware of the implementation of functions that are supposed to be defined in other mixin contracts. TBH, I'm not a huge fan of how this turned out at this point and I think the separation of logic could be cleaner. But, for now since we're not planning on making meaningful changes to the architecture, I'm in favor of minimizing the code diff here and sticking with that architecture if possible.

contracts/IController.sol
kautukkundan: my bad, I first tried converting the contract to an interface when I added the external keyword, then did not change it back to public when I made it abstract again. fixed in ebbb1ef

contracts/bonding/BondingManager.sol
kautukkundan: my bad, did not change it after converting IBondingManager from abstract to interface. fixed in ca65491

kautukkundan:
Will squash fixups before merge once 69ac6cf 51c2bff are good to go

yondonfu (approved):
LGTM after squashing the fixup commits

PR #538, opened 2022-02-06
proxy: Add receive() to ManagerProxy

yondonfu:

Add a payable `receive()` function to ManagerProxy to allow proxies to accept ETH via vanilla transfers (i.e. no msg.data) only if the target implementation contract implements `receive()` as well (since the ManagerProxy `receive()` will trigger a delegatecall with empty msg.data on the target implementation). An example of a contract that needs to accept ETH via vanilla transfers is the [L2Migrator](https://github.com/livepeer/arbitrum-lpt-bridge/blob/8da76d175594cbbeda3b632ffaa9032dc57394d2/contracts/L2/gateway/L2Migrator.sol#L241).

Prior to this change, a ManagerProxy would be able to accept regular ETH transfers when there is no msg.data if the target implementation contract implements a fallback (either `receive()` or `fallback()` with the latter being invoked if `receive()` is not implemented). However, this left room for interface confusion - if a ManagerProxy successfully accepted a regular ETH transfer was it because of `receive()` or `fallback()` being called on the target implementation? This change makes the call path clearer - if a vanilla ETH transfer is accepted `receive()` was triggered on the target implementation and if a complex ETH transfer (i.e. msg.data is included) is accepted `fallback()` was triggered on the target implementation.

So, there aren't any functional changes in this PR, but the changes establish more clarity with call paths.

https://github.com/livepeer/arbitrum-lpt-bridge/blob/main/contracts/proxy/ManagerProxy.sol will need to be updated in arbitrum-lpt-bridge after this PR as well.

See commit history.

Added a test case for ManagerProxy receiving ETH.

N/A

**Checklist:**

### DIFF
--- a/contracts/ManagerProxy.sol
+++ b/contracts/ManagerProxy.sol
@@ -28,12 +28,26 @@ contract ManagerProxy is ManagerProxyTarget {
     }
 
     /**
-     * @notice Uses delegatecall to execute function calls on this proxy contract's target contract using its own storage context.
+     * @notice Fallback function that delegates calls to target contract when there is no msg.data
+     */
+    receive() external payable {
+        _fallback();
+    }
+
+    /**
+     * @notice Fallback function that delegates calls to target contract when there is msg.data
+     */
+    fallback() external payable {
+        _fallback();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


kautukkundan (approved):
LGTM

PR #554, opened 2022-03-16
BondingManager-fuzzing: added fuzzing campaign against BondingManager transferBond

RiccardoBiosas:
Adds a contract under the foundry/dapptools directory to perform a fuzzing campaign against the BondingManager's ```transferBond``` function

No relevant updates to the rest of the codebase

Run the new fuzzing tests and the previous test suites to ensure absence of breaking changes

no

**Checklist:**

### DIFF
--- a/src/test/BondingManagerFuzzer.sol
+++ b/src/test/BondingManagerFuzzer.sol
@@ -25,48 +25,47 @@ contract BondingManagerFuzzer is DSTest {
     CheatCodes public constant CHEATS = CheatCodes(HEVM_ADDRESS);
 
     BondingManager public constant BONDING_MANAGER = BondingManager(0x35Bcf3c30594191d53231E4FF333E8A770453e40);
-    address public constant MINTER_ADDRESS = 0xc20DE37170B45774e6CD3d2304017fc962f27252;
     LivepeerToken public constant LIVEPEER_TOKEN = LivepeerToken(0x289ba1701C2F088cf0faf8B3705246331cB8A839);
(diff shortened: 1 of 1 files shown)


src/test/BondingManagerFuzzer.sol
```diff
@@ -0,0 +1,95 @@
...
+            _amount,
+            address(0),
+            address(0),
+            address(0),
+            address(0)
+        );
+
+        // // time-travels to new new round and initializes it
+        CHEATS.roll(block.timestamp + ROUNDS_MANAGER.roundLength() * 2);
+        ROUNDS_MANAGER.initializeRound();
```
kautukkundan: Delegator's bonded amounts will be reflected in the same round, so we can skip this part. Doesn't hurt to keep it as is though but it saves a bit of runtime.

kautukkundan (approved):
LGTM overall

kautukkundan:
WDYT about testing with changing delegates in addition to changing the amount?

RiccardoBiosas:
> WDYT about testing with changing delegates in addition to changing the amount?

good idea

yondonfu:
@RiccardoBiosas Looks like this is approved. Can this be merged?

RiccardoBiosas:
> WDYT about testing with changing delegates in addition to changing the amount?

It's a good idea, but right now adding additional `address` parameters to the fuzzing test will cause it to freeze indefinitely (might be something related to https://github.com/foundry-rs/foundry/issues/1201). I think it might also help to do a more exhaustive fuzzing campaign in a local e2e setup rather than a live network's fork, so it's something that we can do in a separate issue.

Issue #555, opened 2022-03-18
Refactor Foundry tests: create base test contract to validate upgradeability and governance logic


RiccardoBiosas:
The Livepeer Protocol codebase has recently started implementing [foundry/dapptools unit tests and fuzzing tests](https://github.com/livepeer/protocol/blob/confluence/src/test/) as part of its pipeline.
However, as it stands now, the test contracts include [repeating logic](https://github.com/livepeer/protocol/blob/confluence/src/test/MinterGlobalTotalSupplyFix.sol#L56-L107) and [interfaces](https://github.com/livepeer/protocol/blob/confluence/src/test/BondingManagerForceChangeDelegateFix.sol#L9-L38) which hurt their readability and easiness of refactoring should future changes in the business logic also require changes in the pertinent tests.

For the above purpose, the focus of the current issue is:
- create a base test contract for all the upgradeability and governance logic that will be inherited by other tests whenever necessary 
- remove all the repeating interfaces and move them to a single file

PR #556, opened 2022-04-19
Rb/foundry upgradeability refactoring

RiccardoBiosas:
The current Foundry tests include a lot of [repeating logic](https://github.com/livepeer/protocol/blob/confluence/src/test/MinterGlobalTotalSupplyFix.sol#L56-L107) and [code duplication](https://github.com/livepeer/protocol/blob/confluence/src/test/BondingManagerForceChangeDelegateFix.sol#L9-L38), the PR addresses the issue by moving all the shared interfaces under a single directory and by creating a shared base test contract to encapsulate the governor logic used by other contracts/tests

Refactored all the Foundry tests

Run the the entire hardhat test suite (```yarn test```) and the foundry tests

[Refactor Foundry tests](https://github.com/livepeer/protocol/issues/555)

**Checklist:**


src/test/MinterGlobalTotalSupplyFix.sol
```diff
@@ -1,59 +1,57 @@
...
-        datas = [
-            abi.encodeWithSelector(MINTER.migrateToNewMinter.selector, address(newMinter)),
-            abi.encodeWithSelector(TOKEN.grantRole.selector, MINTER_ROLE, address(newMinter)),
-            abi.encodeWithSelector(TOKEN.revokeRole.selector, MINTER_ROLE, address(MINTER)),
-            abi.encodeWithSelector(CONTROLLER.setContractInfo.selector, MINTER_ID, address(newMinter), gitCommitHash)
-        ];
+        data[0] = abi.encodeWithSelector(MINTER.migrateToNewMinter.selector, address(newMinter));
+        data[1] = abi.encodeWithSelector(TOKEN.grantRole.selector, MINTER_ROLE, address(newMinter));
+        data[2] = abi.encodeWithSelector(TOKEN.revokeRole.selector, MINTER_ROLE, address(MINTER));
+        data[3] = abi.encodeWithSelector(
```
kautukkundan, line 39: What was the idea behind assigning values by index instead of directly creating the array inline with the required values?

src/test/BondingManagerForceChangeDelegateFix.sol
```diff
@@ -148,8 +86,6 @@ contract BondingManagerForceChangeDelegateFix is DSTest {
         bytes32[] memory proof;
         CHEATS.prank(delegator);
 
-        CHEATS.expectEmit(true, true, true, true);
-        emit Bond(delegate, address(0), delegator, 500000000000000000000, 500000000000000000000);
```
kautukkundan: This event should not be removed.

This test is to check for the case where L2Migrator is [allowed to change delegate of the delegator](https://github.com/livepeer/protocol/blob/confluence/contracts/bonding/BondingManager.sol#L505).

The addresses are selected such that a delegatorPool doesn't exist for the delegate and the tokens are directly bonded instead of getting claimed from the pool  [by calling bondFor function](https://github.com/livepeer/arbitrum-lpt-bridge/blob/main/contracts/L2/gateway/L2Migrator.sol#L183-L188) (mimics a situation where the orchestrator hasn't migrated yet) 

In case the tokens are claimed from the delegatorPool, the final assertion will still hold true but TransferBond event will be emitted instead of Bond since `bondForWithHint` function is not called.

kautukkundan (changes requested):

src/test/base/GovernorBaseTest.sol
```diff
@@ -0,0 +1,54 @@
+pragma solidity ^0.8.9;
+
+import "ds-test/test.sol";
```
0xcadams, line 3: Will we remove `ds-test` in the future? I saw this issue so I'm imagining that's future work
https://github.com/livepeer/protocol/issues/578

src/test/MinterGlobalTotalSupplyFix.sol
RiccardoBiosas: They were previously dynamic storage arrays and it is not possible to initialize them in the same way now that they are dynamically-sized memory arrays

src/test/BondingManagerForceChangeDelegateFix.sol
RiccardoBiosas: Thanks for the explanation, I added it back

src/test/base/GovernorBaseTest.sol
RiccardoBiosas: correct!

kautukkundan (approved):
LGTM ✅

src/test/MinterGlobalTotalSupplyFix.sol
kautukkundan: self note : 
dynamic storage arrays were converted to dynamic in-memory arrays to prevent redeclarations

PR #557, opened 2022-05-02
Rb/hardhat geth config

RiccardoBiosas:
Apply the changes required to package the confluence version of the Livepeer protocol in a docker container for the purpose of e2e testing with go-livepeer

- Adds a contract mock for the L2LPTDataCache contract which is a dependency of the Minter contract
- Adds a deployment script for arbitrum mocks (currently only the L2LPTDataCacheMock)
- Fixes a bug which occurs when trying to run any hardhat command without having already generated the typechain's contract types since they are required by some hardhat tasks ([here](https://github.com/livepeer/protocol/blob/rb/hardhat-geth-config/tasks/verify-protocol.ts#L6-L13)) which are loaded on the [hardhat-config](https://github.com/livepeer/protocol/blob/rb/hardhat-geth-config/hardhat.config.ts#L21-L26))
- Adds a hardhat task to print the address of a given deployed contract
- Adds local geth-specific protocol parameters

Run the protocol's test suite and did manual testing with go-livepeer

https://github.com/livepeer/go-livepeer/issues/2315

**Checklist:**

### DIFF
--- a/contracts/test/mocks/L2LPTDataCacheMock.sol
+++ b/contracts/test/mocks/L2LPTDataCacheMock.sol
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+contract L2LPTDataCacheMock {
+    function l1CirculatingSupply() external pure returns (uint256) {
+        return 0;
+    }
+}
--- a/deploy/deploy_arbitrum_lpt_mocks.ts
+++ b/deploy/deploy_arbitrum_lpt_mocks.ts
@@ -1,0 +1,31 @@
+import {HardhatRuntimeEnvironment} from "hardhat/types"
+import {DeployFunction} from "hardhat-deploy/types"
+import {contractId} from "../utils/helpers"
+
+const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
+    const {deployments, getNamedAccounts, ethers} = hre
+    const {deploy, get} = deployments
+
+    const {deployer} = await getNamedAccounts()
+
+    const deployResult = await deploy("L2LPTDataCacheMock", {
+        from: deployer,
+        log: true
+    })
+
+    const controllerDeployment = await get("Controller")
+    const controller = await ethers.getContractAt(
+        "Controller",
+        controllerDeployment.address
+    )
+
+    const id = contractId("L2LPTDataCache")
+    await controller.setContractInfo(
+        id,
+        deployResult.address,
+        "0x1111111111111111111111111111111111111111"
+    )
+}
+
+func.tags = ["ARBITRUM_LPT_MOCK"]
+export default func
--- a/deploy/deploy_arbitrum_lpt_mocks.ts
+++ b/deploy/deploy_arbitrum_lpt_mocks.ts
@@ -8,7 +8,7 @@ const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
 
     const {deployer} = await getNamedAccounts()
 
-    const deployResult = await deploy("L2LPTDataCacheMock", {
+    const deployResult = await deploy("DummyL2LPTDataCache", {
         from: deployer,
         log: true
     })
--- a/deploy/migrations.config.ts
+++ b/deploy/migrations.config.ts
@@ -1,6 +1,6 @@
 import {ethers} from "ethers"
 
-const geth = {
+const gethDev = {
     bondingManager: {
         numTranscoders: 100,
         numActiveTranscoders: 50,
@@ -160,7 +160,7 @@ const networkConfigs: any = {
     arbitrumRinkeby,
     arbitrumRinkebyDevnet,
     arbitrumMainnet,
-    geth
+    gethDev
 }
 
 export default function getNetworkConfig(network: string) {
--- a/deploy/migrations.config.ts
+++ b/deploy/migrations.config.ts
@@ -1,5 +1,33 @@
 import {ethers} from "ethers"
 
+const localGeth = {
+    bondingManager: {
+        numTranscoders: 100,
+        numActiveTranscoders: 50,
+        unbondingPeriod: 7,
+        maxEarningsClaimsRounds: 20
+    },
+    broker: {
+        unlockPeriod: 50,
+        ticketValidityPeriod: ethers.BigNumber.from(2)
+    },
+    roundsManager: {
+        roundLength: 50,
+        roundLockAmount: 100000
+    },
+    faucet: {
+        requestAmount: ethers.utils.parseEther("10"),
+        requestWait: 1,
+        whitelist: []
+    },
+    minter: {
+        inflation: 137,
+        inflationChange: 3,
+        targetBondingRate: 500000
+    }
+}
+
+
 const defaultConfig = {
     bondingManager: {
         numTranscoders: 20,
@@ -132,7 +160,8 @@ const networkConfigs: any = {
     rinkebyDevnet,
     arbitrumRinkeby,
     arbitrumRinkebyDevnet,
-    arbitrumMainnet
+    arbitrumMainnet,
+    localGeth
 }
 
 export default function getNetworkConfig(network: string) {
--- a/deploy/migrations.config.ts
+++ b/deploy/migrations.config.ts
@@ -1,6 +1,6 @@
 import {ethers} from "ethers"
 
-const localGeth = {
+const geth = {
     bondingManager: {
         numTranscoders: 100,
         numActiveTranscoders: 50,
@@ -26,7 +26,6 @@ const localGeth = {
         targetBondingRate: 500000
     }
 }
-
 
 const defaultConfig = {
     bondingManager: {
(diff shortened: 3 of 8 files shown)


contracts/test/mocks/L2LPTDataCacheMock.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+contract L2LPTDataCacheMock {
+    function l1CirculatingSupply() external pure returns (uint256) {
+        return 0;
```
kautukkundan: Since the minter contract is dependent upon this value for calculating inflation rate and subsequently rewards, I believe fixing this value to be 0 may cause some troubles if reward calls etc need to be tested.
(Though, It looks like these tests are not planned yet?) 

WDYT about creating a public variable which can be set accordingly via a task during tests and this function will simply return that value.

RiccardoBiosas: I was assuming that go-livepeer is not directly concerned with the inflation rate as it is more of a protocol-side business logic, so I wanted to keep the mock interface as simple as possible. However sure, if it is actually needed I think it is a good idea. It could be something like the following (+ related hardhat task):
```
// SPDX-License-Identifier: MIT
pragma solidity 0.8.9;

contract L2LPTDataCacheMock {
    uint256 public inflationDelta;

    function l1CirculatingSupply() external pure returns (uint256) {
        return 0;
    }

    function setInflationDelta(uint256 _inflationDelta) external {
        inflationDelta = _inflationDelta;
    }
}
```

Maybe @leszko can chime in

leszko: IIUC We use this mock contract **only** for the [geth-with-protocol](https://github.com/livepeer/docker-livepeer/tree/master/geth-with-protocol) docker image. Then, two questions:

1. Why do we need to hardcode `l1CirculatingSupply` to `0` in the `geth-with-protocol` docker image? What would happen if we don't have it? Or if we deploy the `real` Minter contract, not the mock one? Maybe there is a way to avoid having this `L2LPTDataCacheMock.sol`?

2. I might be wrong, but I see all mock contracts in `contracts/test/mocks` actually serve a different purpose. They are used in the unit tests in `livepeer/protocol`. `L2LPTDataCacheMock.sol` is different, because it's actually not used in `livepeer/protocol`, but in `livepeer/docker-livepeer`. Then, maybe to avoid confusion, we should move it somewhere else? Maybe it's not a mock, but more something like `DummyGateway` and we should call it `DummyMinter`?

deploy/deploy_arbitrum_lpt_mocks.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const id = contractId("L2LPTDataCache")
+    await controller.setContractInfo(
+        id,
+        deployResult.address,
+        "0x1111111111111111111111111111111111111111"
+    )
+}
+
+func.tags = ["ARBITRUM_LPT_MOCK"]
```
leszko: I wonder it `ARBITRUM_LPT_MOCKS` is explicit enough. The same with the file name `deploy_arbitrum_lpt_mocks.ts`. I think that, as per my previous comment, the best would be to not have this `L2LPTDataCacheMock` contract at all. But if we really do need it, then I'd consider renaming it something like `DummyMinter` or maybe `DummyL1Supply` and then follow the naming in the tag and this file name.

deploy/migrations.config.ts
```diff
@@ -1,5 +1,32 @@
 import {ethers} from "ethers"
 
+const geth = {
```
leszko: Maybe we should call it `dev` instead of `geth`? Or does `geth` always mean the local dev network?

hardhat.config.ts
```diff
@@ -26,7 +25,10 @@ function loadTasks() {
     })
 }
 
-if (fs.existsSync(path.join(__dirname, "artifacts"))) {
+if (
+    fs.existsSync(path.join(__dirname, "artifacts")) &&
+  fs.existsSync("./typechain")
```
leszko: Nit: Please reformat. It looks like the indentation is wrong.

leszko (commented):
Added a few comments. But in general I really like that you moved setting contract parameters from `livepeer/docker-livepeer` into `livepeer/protocol`.

deploy/deploy_arbitrum_lpt_mocks.ts
RiccardoBiosas: > Why do we need to hardcode l1CirculatingSupply to 0 in the geth-with-protocol docker image? What would happen if we don't have it? Or if we deploy the real Minter contract, not the mock one? Maybe there is a way to avoid having this L2LPTDataCacheMock.sol?

We are deploying the actual Minter contract. The difference is that the 'post-confluence' Minter contract [relies on an external dependency to calculate the total supply of livepeer tokens](https://github.com/livepeer/protocol/blob/confluence/contracts/token/Minter.sol#L228). The actual L2LPTDataCache is defined elsewhere - [here in the arbitrum-lpt repo](https://github.com/livepeer/arbitrum-lpt-bridge/blob/main/contracts/L2/gateway/L2LPTDataCache.sol#L98-L111)- and, as I understand, its responsibility is to ensure that the global supply takes into account the l1/mainnet token supply difference if necessary (i.e.: in the case of inflation on L1).
Since it seems that the agreed-upon direction was to[ keep the initial docker workflow fairly simple and not supporting cross-chain testing](https://github.com/livepeer/go-livepeer/issues/2297), I opted for the usage of a mock.

> I wonder it ARBITRUM_LPT_MOCKS is explicit enough. The same with the file name deploy_arbitrum_lpt_mocks.ts. I think that, as per my previous comment, the best would be to not have this L2LPTDataCacheMock contract at all. But if we really do need it, then I'd consider renaming it something like DummyMinter or maybe DummyL1Supply and then follow the naming in the tag and this file name.

IMHO I think it falls under the definition of a mock, but if you insist on it I can change the name.

deploy/migrations.config.ts
RiccardoBiosas: Initially I was undecided on whether just use the [localhost network](https://github.com/livepeer/protocol/blob/confluence/hardhat.config.ts#L85). But I guess a combination of the two works fine - `gethDev`. (and no, geth != local dev network)

hardhat.config.ts
RiccardoBiosas: odd, prettier/eslint did not complain about it

deploy/migrations.config.ts
leszko: `gethDev` is ok. Also `dev` or `devnet` would work imo.

deploy/deploy_arbitrum_lpt_mocks.ts
leszko: Thanks for the explanation. It makes sense.

About the naming, since we already use `Dummy` somewhere in the code, It'd use Dummy for it. What you have is closer to Dummy or Stub than to a Mock. Mocking is more about [recording behavior and setting the expected values.](https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing).

leszko: What about renaming `ARBITRUM_LPT_MOCKS` and `deploy_arbitrum_lpt_mocks.ts`? I think the better names would be something like: `DUMMY_L2_CACHE` and `deploy_dummy_l2_cache.ts` or something similar.

leszko (changes requested):
The PR looks good (just added one comment about renaming), but [devtool](https://github.com/livepeer/go-livepeer/tree/master/cmd/devtool) still didn't work with the image built from https://github.com/livepeer/docker-livepeer/pull/92? 

Could you check if devtool works for you?

contracts/test/mocks/L2LPTDataCacheMock.sol
kautukkundan: > We use this mock contract only for the [geth-with-protocol](https://github.com/livepeer/docker-livepeer/tree/master/geth-with-protocol) docker image

gotcha 👍🏼 
In this case we can skip modifying the contract. PR good to go once Rafal's comments are addressed.

deploy/deploy_arbitrum_lpt_mocks.ts
RiccardoBiosas: I'd prefer to keep it more generic - in case in the future the file will also include more L2 dummy deployments. But I replaced the MOCK suffix with DUMMY

RiccardoBiosas:
> The PR looks good (just added one comment about renaming), but [devtool](https://github.com/livepeer/go-livepeer/tree/master/cmd/devtool) still didn't work with the image built from [livepeer/docker-livepeer#92](https://github.com/livepeer/docker-livepeer/pull/92)?
> 
> Could you check if devtool works for you?

It works for me. What issues are you experiencing?

leszko:
> > The PR looks good (just added one comment about renaming), but [devtool](https://github.com/livepeer/go-livepeer/tree/master/cmd/devtool) still didn't work with the image built from [livepeer/docker-livepeer#92](https://github.com/livepeer/docker-livepeer/pull/92)?
> > Could you check if devtool works for you?
> 
> It works for me. What issues are you experiencing?

Ok, it worked now 💪

leszko (approved):
LGTM 👍 

Thanks for all the work Riccardo!

PR #570, opened 2022-05-11
ci: added initial slither to ci

### LINKED ISSUE #567: Add slither to github actions
Slither provided some useful plugin to perform static analysis - in particular the upgradeability plugin might be a useful testing tool.
However, since static analysis is known to be quite noisy/give a fair amount of false positives, we should decide whether we should just run slither in the testing pipeline for 'awareness' or flag the github action as failed under some circumstances (i.e.: if slither detects X possibly critical issues).

0xcadams:
Adds Slither to CI with Github Code Scanning integration.

CI updates + security analysis comments integration w/ GH
Slither has certain noisy configs ignored (for now)

See below

Fixes https://github.com/livepeer/protocol/issues/567

**Checklist:**


slither.config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "detectors_to_exclude": "external-function",
```
0xcadams, line 2: Here is the documentation behind this detector:
https://github.com/crytic/slither/wiki/Detector-Documentation#public-function-that-could-be-declared-external

There is opinion whether this is useful anymore with the latest compiler.

slither.config.json
```diff
@@ -0,0 +1,5 @@
...
+  "filter_paths": "@openzeppelin|contracts/test",
+  "exclude_informational": true
```
0xcadams, line 4: Excluding informational for now, but there are some useful errors in this category which we should include later

kautukkundan (commented):
Changes look good but what could be done with the alerts which fail the CI. 
Would dismissing an alert once cause it to re-emerge when a new PR is made or would it save it's response?

RiccardoBiosas:
> Changes look good but what could be done with the alerts which fail the CI. Would dismissing an alert once cause it to re-emerge when a new PR is made or would it save it's response?

IMHO I don't think it's desirable to force a CI action to fail on the basis of the result of Slither since static analysis is known to give a fair amount of false positives. 
According to [github docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#dismissing--alerts), if you dismiss an alert it should be for good, however github will document the dismissed alerts and the reasons why they have been dismissed

kautukkundan:
> According to [github docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository#dismissing--alerts), if you dismiss an alert it should be for good

Thanks for sharing this, I had been looking in the wrong docs

> IMHO I don't think it's desirable to force a CI action to fail on the basis of the result of Slither since static analysis is known to give a fair amount of false positives.

I agree with this, adding `continue-on-error: true` will [allow-failures](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error). The CI would pass and we would still have result for static code analysis. This should should be suitable for our case.

RiccardoBiosas (changes requested):
Looks good overall and I really like that you paired it with codeQL!
Just a few things:
- can you also run slither with the `slither-check-upgradeability` plugin?
- I think it'd be best to have a separate github action for Slither rather than lumping everything together with the other tests
- [question] seems like codeQL action v1 will be EOL in december, so is worth to switch to v2 right away?

0xcadams:
> adding `continue-on-error: true` will [allow-failures](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error).

This actually already exists on the Slither step - the failure is a GitHub code scanning failure, so we'd have to manually dismiss the security results we'd like to ignore/create issues on the ones we'd like to address.

> IMHO I don't think it's desirable to force a CI action to fail on the basis of the result of Slither since static analysis is known to give a fair amount of false positives.

Yeah, the false positives can be a bit annoying. I think if we remove the failures altogether, we will start to ignore the results of slither. I'd advocate for a process of adding [comments](https://github.com/crytic/slither/wiki/Usage#triage-mode) to manually opt out of Slither false positives, or manually dismissing the security results as "false positive". Otherwise this CI step will get ignored.

Screen Shot 2022-05-16 at 9 53 53 AM

kautukkundan:
> This actually already exists on the Slither step - the failure is a GitHub code scanning failure, so we'd have to manually dismiss the security results we'd like to ignore/create issues on the ones we'd like to address.

Ah yes sorry, [just found out](https://github.com/actions/runner/issues/2347) that it doesn't work as I expected. 

> manually dismissing the security results as "false positive"

I also think this would be an easy way forward. Once we address these alerts they will go away and for future contract updates I think it's nice to have explicit alerts even if they are FPs. The reviewer could address (or dismiss) these before merging. WDYT @RiccardoBiosas.

RiccardoBiosas:
> I also think this would be an easy way forward. Once we address these alerts they will go away and for future contract updates I think it's nice to have explicit alerts even if they are FPs. The reviewer could address (or dismiss) these before merging. WDYT @RiccardoBiosas.

Isn't it what the PR is already doing? The only issues that are dismissed by default are informational

kautukkundan:
> Isn't it what the PR is already doing? 

Yeah, I meant that it's fine only, the way it is, addressing your comment ⬇️

> IMHO I don't think it's desirable to force a CI action to fail on the basis of the result of Slither

0xcadams:
There are actual errors here we need to address - for instance, the unchecked transfer issues are something we should not dismiss.

```
Unchecked transfer
[BondingManager.bondForWithHint(uint256,address,address,address,address,address,address)](https://github.com/livepeer/protocol/pull/570/contracts/bonding/BondingManager.sol#L475-L549) ignores return value by [livepeerToken().transferFrom(msg.sender,address(minter()),_amount)](https://github.com/livepeer/protocol/pull/570/contracts/bonding/BondingManager.sol#L545)
```

I can lump those changes into this PR. As far as the others are concerned, we may want to jump on a 10 min call to make sure we're okay with dismissing them permanently. WDYT @RiccardoBiosas @kautukkundan @yondonfu @red-0ne

0xcadams:
> * can you also run slither with the `slither-check-upgradeability` plugin?

I tried using the upgradeable plugin, but it depends on following the OZ standards for upgradeable contracts:

Screen Shot 2022-05-16 at 11 19 08 AM

https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable

> * I think it'd be best to have a separate github action for Slither rather than lumping everything together with the other tests
> * [question] seems like codeQL action v1 will be EOL in december, so is worth to switch to v2 right away?

I will address these!

RiccardoBiosas:
> I can lump those changes into this PR. As far as the others are concerned, we may want to jump on a 10 min call to make sure we're okay with dismissing them permanently. WDYT @RiccardoBiosas @kautukkundan @yondonfu @red-0ne

yes good idea, i think some of them should be added to the backlog anyway

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,33 @@
+name: Analyze
```
hjpotter92: nit: yaml.org suggests using `.yaml` extension over `.yml` :)

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Needed for commands that depend on git tags
+          fetch-depth: 0
```
hjpotter92: there were some issues w/ checkout action that were faced in go-livepeer. i'd recommend:
```yaml

      - name: Check out code
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
          # Check https://github.com/livepeer/go-livepeer/pull/1891
          # for ref value discussion
          ref: ${{ github.event.pull_request.head.sha }}
```
like in https://github.com/livepeer/go-livepeer/blob/master/.github/workflows/build.yaml

i doubt that this project would face similar issues, but better to prevent the scenario altogether

hjpotter92, line 20: there were some issues w/ checkout action that were faced in go-livepeer. i'd recommend:
```yaml

      - name: Check out code
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
          # Check https://github.com/livepeer/go-livepeer/pull/1891
          # for ref value discussion
          ref: ${{ github.event.pull_request.head.sha }}
```
like in https://github.com/livepeer/go-livepeer/blob/master/.github/workflows/build.yaml

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
```
hjpotter92: you should perhaps use setup-node action here, to keep the version consistent. I think the default github runner comes with node-12 or node-14.

https://github.com/livepeer/livepeer-com/blob/master/.github/workflows/build.yaml#L24-L27

```yaml
      - uses: actions/setup-node@v3
        with:
          node-version: "16"
          cache: "yarn"
```

hjpotter92, line 21: you should perhaps use setup-node action here, to keep the version consistent. I think the default github runner comes with node-12 or node-14.

https://github.com/livepeer/livepeer-com/blob/master/.github/workflows/build.yaml#L24-L27

.github/workflows/analyze.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          yarn
```
hjpotter92: you can also avoid the installation step on each run of this action by using a cache restoration:

https://github.com/actions/cache/blob/main/examples.md#node---yarn

hjpotter92, line 23: you can also avoid the installation step on each run of this action by using a cache restoration:

https://github.com/actions/cache/blob/main/examples.md#node---yarn

hjpotter92 (commented):
left few notes to probably reduce build times and use cache to avoid package installations on each run :)

0xcadams:
@hjpotter92 resolved those issues
@RiccardoBiosas @kautukkundan @yondonfu @red-0ne SARIF upload has been removed for now, and Analyze CI will fail currently until all Slither code quality issues are addressed. Let me know if you'd like to see any changes to this.

.github/workflows/analyze.yaml
```diff
@@ -0,0 +1,49 @@
...
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v3
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: Install dependencies
+        run: |
```
hjpotter92: you need to add `if: steps.yarn-cache.outputs.cache-hit != 'true'` as suggested in the commend from `yarn-cache` step above to avoid running this step entirely :)

.github/workflows/build.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Linter
+        run: |
+          yarn lint
+      - name: Compile contracts
+        run: |
+          yarn compile
+      - name: 💡 Run contract tests with coverage reporting
+        run: |
+          yarn test:coverage
+      - name: Coveralls
```
hjpotter92: we can replace it with `codecov`, as we have better integration (along with more repositories) there.

0xcadams: Created an issue here https://github.com/livepeer/protocol/issues/587

.github/workflows/analyze.yaml
0xcadams: I actually think this caches the yarn cache directory, not the node_modules folder. So we'll still need to run yarn install and it will pull from the cache

0xcadams: https://stackoverflow.com/questions/61010294/how-to-cache-yarn-packages-in-github-actions

hjpotter92: thanks! i did not know that! and also about the caching done by the `setup-node` action!

.github/workflows/analyze.yaml
```diff
@@ -0,0 +1,36 @@
+name: Analyze
+on:
+  pull_request:
+  push:
+    branches:
+      - confluence
+      - next
```
RiccardoBiosas: shouldn't it be 'cf/next'?

0xcadams: Fixed

RiccardoBiosas (approved):

PR #584, opened 2022-05-24
bonding: remove duplicate IBondingManager interface

### LINKED ISSUE #560: Remove duplicate IBondingManager interface
The IBondingManager is declared twice ([here](https://github.com/livepeer/protocol/blob/confluence/contracts/bonding/IBondingManager.sol) and [here](https://github.com/livepeer/protocol/blob/confluence/contracts/polling/PollCreator.sol#L6-L10)).
Consider removing it to improve code's readability and avoid cyclic dependency issues down the road.

red-0ne:
Remove duplicate IBondingManager interface

- Merge IBondingManager interfaces
- Move IBondingManager.sol to `contracts/interfaces`
- Adapt `test/unit/PollCreator.js` test

`yarn test`
`yarn deploy`

Fixes #560

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -2,7 +2,7 @@
 pragma solidity 0.8.9;
 
 import "../ManagerProxyTarget.sol";
-import "./IBondingManager.sol";
+import "../interfaces/IBondingManager.sol";
 import "../libraries/SortedDoublyLL.sol";
 import "../libraries/MathUtils.sol";
 import "../libraries/PreciseMathUtils.sol";
--- a/contracts/interfaces/IBondingManager.sol
+++ b/contracts/interfaces/IBondingManager.sol
@@ -78,4 +78,6 @@ interface IBondingManager {
     function isActiveTranscoder(address _transcoder) external view returns (bool);
 
     function getTotalBonded() external view returns (uint256);
+
+    function pendingStake(address _addr, uint256 _endRound) external view returns (uint256);
 }
--- a/contracts/pm/mixins/interfaces/MContractRegistry.sol
+++ b/contracts/pm/mixins/interfaces/MContractRegistry.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.9;
 
-import "../../../bonding/IBondingManager.sol";
+import "../../../interfaces/IBondingManager.sol";
 import "../../../token/IMinter.sol";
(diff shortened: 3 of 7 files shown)


kautukkundan (approved):
Since the beginning, Interfaces have always been in the same directory as the implementation contract. It is so that the interface of a contract is within the context of the folder name and easier to search. But it also kinda makes sense to group all the interfaces in the same interface directory. 

I am fine with either but it will be great if it remains consistent. This can be addressed in the same or different PR. Apart from this, changes look good 👍🏼

RiccardoBiosas (approved):

red-0ne:
> Since the beginning, Interfaces have always been in the same directory as the implementation contract. It is so that the interface of a contract is within the context of the folder name and easier to search. But it also kinda makes sense to group all the interfaces in the same interface directory.

Both make sense IMO. As long as we keep that consistent. We can view it as @kautukkundan  addressed it and keep context closer to the implementation or separate abstraction from implementation giving maybe more freedom for both of the realms to evolve.

@RiccardoBiosas We just need to make sure to allocate resources to migrate everything to the directory structure. If that's all good. I'll be good to merge.

RiccardoBiosas:
> I am fine with either but it will be great if it remains consistent. This can be addressed in the same or different PR. Apart from this, changes look good 👍🏼

Yes, good point! The scope of the current issue was narrowed down to just removing the occurrences of duplicate interfaces for the BondingManager contract - which were also declared on the same file as the consumer contracts rather than on their own file. There are similar [occurrences in other parts of the codebase](https://github.com/livepeer/protocol/blob/confluence/contracts/token/Minter.sol#L13-L15), which can potentially lead to cyclic dependency issues.
As for where to place the interfaces, IMHO I think they are better off under their own directory `interfaces` since they are meant for project-wide consumption rather than being tightly related to their implementation. Most of the codebases I'm familiar with follow a similar directory structure/pattern ([here](https://github.com/aave/protocol-v2/tree/master/contracts/interfaces), [here](https://github.com/Uniswap/v3-core/tree/main/contracts), [here](https://lbhackney-it.github.io/API-Playbook/clean_architecture/) ). Anyway enforcing consistency in that regard should go in a separate issue.

RiccardoBiosas:
> @RiccardoBiosas We just need to make sure to allocate resources to migrate everything to the directory structure. If that's all good. I'll be good to merge.

Sounds good, but please hold off on merging since we are working out how to structure updates to the contract codebase moving forward - where the main branch should just represent the state of the deployed codebase.
We can discuss it further offline

PR #593, opened 2022-06-23
Move AdjustableRoundsManager from protocol directory to contract tests directory

RiccardoBiosas:
It moves the `AdjustableRoundsManager` contract from the protocol `/rounds` directory to the tests `/mocks` directory as it is a contract that is meant to be used only for testing.

`contracts/rounds/AdjustableRoundsManager.sol -> contracts/test/mocks/AdjustableRoundsManager.sol`

Successfully run the protocol's test suite.

No

**Checklist:**

### DIFF
--- a/contracts/test/mocks/AdjustableRoundsManager.sol
+++ b/contracts/test/mocks/AdjustableRoundsManager.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.9;
 
-import "./RoundsManager.sol";
+import "../../rounds/RoundsManager.sol";
 
 contract AdjustableRoundsManager is RoundsManager {
     uint256 public num;


contracts/test/mocks/AdjustableRoundsManager.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.9;
 
-import "./RoundsManager.sol";
+import "../../rounds/RoundsManager.sol";
 
 contract AdjustableRoundsManager is RoundsManager {
```
leszko, line 6: Don't we need to update some paths also in some `js/ts` files? I see it's used in a number of files, e.g., [here](https://github.com/livepeer/protocol/blob/c3b6273651801121c0fa533d4f68dd1642d16a2e/test/gas-report/redeemTicket.js#L53).

leszko (approved):
Added one comment.

contracts/test/mocks/AdjustableRoundsManager.sol
RiccardoBiosas: It's not necessary because those hardhat/ethers methods are dependent on the name of the compiled contract rather than its path

Issue #576, opened 2022-05-16
Run all the Solidity tests with Foundry


RiccardoBiosas:
**Context**
At the moment the protocol code-base includes two types of Solidity tests: [truffle-based Solidity tests](https://github.com/livepeer/protocol/blob/confluence/contracts/test/TestEarningsPool.sol) and [Foundry Solidity tests.](https://github.com/livepeer/protocol/blob/confluence/src/test/BondingManagerTransferBondFix.sol). The truffle-based Solidity tests are run as part of the unit tests via the [`runSolidityTest` helper](https://github.com/livepeer/protocol/blob/confluence/test/unit/helpers/runSolidityTest.js).
I propose to convert them to Foundry tests to enforce more consistency in the code-base and because Foundry is the clear winner in terms of performance and testing capabilities.

**TODO**
- Move the [Solidity tests](https://github.com/livepeer/protocol/tree/confluence/contracts/test) to the Foundry directory
- Make the necessary refactoring to run them as tests with Foundry/forge - i.e.: replacing the `before`, `beforeEach` hooks with `setUp`. I'm estimating the refactoring process should be fairly straightforward
- Remove the [`runSolidityTest` helper](https://github.com/livepeer/protocol/blob/confluence/test/unit/helpers/runSolidityTest.js)

PR #595, opened 2022-07-13
Rb/truffle to foundry

RiccardoBiosas:
Since the protocol repo has been gradually adopting Foundry to run Solidity-based tests - which provides a better alternative to Truffle-, the PR refactors the current truffle-based Solidity tests so that they can be run as part of the Foundry codebase.

- Moved the [Solidity tests](https://github.com/livepeer/protocol/tree/confluence/contracts/test) to the Foundry directory and refactored them accordingly
- Removed the [`runSolidityTest` helper](https://github.com/livepeer/protocol/blob/confluence/test/unit/helpers/runSolidityTest.js)
- Removed all the [truffle-based Solidity helpers](https://github.com/livepeer/protocol/tree/confluence/contracts/test/helpers/truffle) and moved the [RevertProxy helper](https://github.com/livepeer/protocol/blob/confluence/contracts/test/helpers/RevertProxy.sol) to the Foundry directory

Successfully run `yarn test` and `forge test -v --match-contract Test` to test all the new foundry-based unit tests

https://github.com/livepeer/protocol/issues/576

**Checklist:**


src/test/unit/TestSortedDoublyLLUpdateKey.sol
```diff
@@ -1,33 +1,33 @@
...
-contract TestSortedDoublyLLUpdateKey {
     address[] ids = [address(1), address(2), address(3), address(4), address(5), address(6)];
     uint256[] keys = [uint256(13), uint256(11), uint256(9), uint256(7), uint256(5), uint256(3)];
 
     SortedDoublyLLFixture fixture;
     RevertProxy proxy;
 
-    function beforeAll() public {
+    function setUp() public {
         proxy = new RevertProxy();
```
red-0ne, line 19: Foundry doesn't seem to have a `beforeAll` equivalent. I'm not sure if it's safe to execute `RevertProxy` before each test. We could wrap this into an `if` block if it is not.

(Same of the other occurrences of `beforeAll`)

red-0ne (approved):
Left one non-blocking comment. Other than that, looks all good to me.

kautukkundan (approved):
LGTM

src/test/unit/TestSortedDoublyLLUpdateKey.sol
kautukkundan: should be fine IMO

Also additionally (prob irrelevant to this) - 
I observed that `constructor` behaves the same way as `setUp`. It is also called each time before a test.

kautukkundan (changes requested):
Ah actually it would be great to add additional tests for the reduced coverage - https://coveralls.io/builds/50828602
Should be small only

src/test/unit/TestSortedDoublyLLUpdateKey.sol
RiccardoBiosas: yes, it shouldn't be an issue since the internal state of `RevertProxy` is not meant to be shared among different test cases - but I agree that the lack of some lifecycle methods in the current version of foundry is good to keep in mind

RiccardoBiosas:
> Ah actually it would be great to add additional tests for the reduced coverage - https://coveralls.io/builds/50828602 Should be small only

The only reason why the coverage dropped is because the package `solidity-coverage` is not currently taking into account the foundry tests - unlike the previous truffle-based tests run via [`runSolidityTest.js`](https://github.com/livepeer/protocol/blob/confluence/test/unit/helpers/runSolidityTest.js). Foundry has recently added a `coverage` feature in one of its latest 2.x releases but we cannot upgrade to it right away due to [this](https://github.com/foundry-rs/foundry/issues) issue in the way foundry 2.x handles L2 networks: once the issue will be addressed I think we can fix the 'fake' coverage drop.

kautukkundan (approved):
Gotcha! Thanks for the note.

PR #599, opened 2022-09-12
added tests for lip-83: roundLength change

kautukkundan:

- created tests as mentioned in https://github.com/livepeer/internal-project-tracking/issues/400
 
ran `yarn test:coverage`

closes https://github.com/livepeer/internal-project-tracking/issues/400

**Checklist:**

### DIFF
--- a/test/integration/Delegation.js
+++ b/test/integration/Delegation.js
@@ -219,213 +219,232 @@ describe("Delegation", () => {
         ).to.be.reverted
     })
 
-    it("delegator 1 partially unbonds twice (at different times), rebonds with one unbonding lock and withdraws with the other AFTER roundLength update", async () => {
-        await roundsManager.mineBlocks(roundLength)
+    it("delegator 1 partially unbonds twice (at different times), rebonds with one unbonding lock and withdraws with the other", async () => {
+        const runTest = async _roundLength => {
+            await roundsManager.mineBlocks(_roundLength)
+            await roundsManager.initializeRound()
+
+            const unbondingPeriod = await bondingManager.unbondingPeriod()
+            const startDelegatorBondedAmount = (
+                await bondingManager.getDelegator(delegator1.address)
+            )[0].toNumber()
+            const startDelegatorTokenBalance = await token.balanceOf(
+                delegator1.address
+            )
+            const startMinterTokenBalance = await token.balanceOf(
+                minterAddr.address
+            )
+            const startTranscoderDelegatedAmount = (
+                await bondingManager.getDelegator(transcoder2.address)
+            )[3].toNumber()
+            const startNextTotalStake = (
+                await bondingManager.nextRoundTotalActiveStake()
+            ).toNumber()
+
+            const unbondingLockID0 = 0
+            const unbondingLockID1 = 1
+
+            // Delegator 1 partially unbonds from transcoder 2
+            // unbondingLockID = 0
+            await bondingManager.connect(delegator1).unbond(500)
+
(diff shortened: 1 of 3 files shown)


test/integration/Delegation.js
```diff
@@ -219,8 +219,10 @@ describe("Delegation", () => {
         ).to.be.reverted
     })
 
-    it("delegator 1 partially unbonds twice (at different times), rebonds with one unbonding lock and withdraws with the other", async () => {
+    it("delegator 1 partially unbonds twice (at different times), rebonds with one unbonding lock and withdraws with the other AFTER roundLength update", async () => {
```
yondonfu: It seems that there are two separate test scenarios here:

1. The scenario for an unbond, rebond, withdraw *without* a roundLength update
2. The scenario for an unbond, rebond, withdraw *with* a roundLength update

These two scenarios are very similar and are only distinguished by the existence of a roundLength update. But, for test structure clarity I think it would be beneficial to have a clear distinction between the two scenarios because we want clear confirmation that the unbond, rebond, withdraw flow works properly regardless of whether or not a roundLength update occurred.

I think an easy to way to achieve this clarity would be to move the logic in this block into a helper function that is executed once for scenario 1 and once for scenario 2. And for scenario 2 we would just execute the roundLength update before executing the helper.

test/integration/Delegation.js
```diff
@@ -421,6 +425,8 @@ describe("Delegation", () => {
             startMinterTokenBalance.sub(BigNumber.from(700)).toString(),
             "wrong minter token balance after withdrawStake"
         )
+        await roundsManager.mineBlocks(newRoundsLength)
```
yondonfu: It seems that the next test case starts with block progression + round initialization. Can we omit these lines and let the next test case handle setting up the test properly? If we update the global `roundLength` var then subsequent test cases would use the new roundLength value for handling block progression.

yondonfu (changes requested):

test/integration/Delegation.js
kautukkundan: Since this is an integration test it shares the same contract state across test cases, so either way it needs to be reset. I think resetting it to the original value right after that test makes more sense instead of resetting in next test.

yondonfu: If the next test also calls mineBlocks() and initializeRound() immediately at the start of the test *and* we do the same here and the end of this test wouldn't we be initializing 2 rounds when we only need to initialize 1 round to setup the next test?

I see what you're saying, but primarily highlighting the above because want to clarify if these lines are actually necessary given the presence of the same lines at the start of the next test.

kautukkundan: I agree with you. I had added mineBlocks() based on my previous thought about finishing the round first before updating the roundLength. But based on comments here https://github.com/livepeer/internal-project-tracking/issues/400#issuecomment-1251065972 
I am now convinced to remove this

kautukkundan:
reviewing suggestion for latest commit - @yondonfu 
I moved the entire test into a helper function, hence the diff may seem large. But essentially it just wraps it in a function and changes global variable roundLength with a local variable _roundLength inside the function. 

using evm_snapshot and evm_revert because subsequent tests depends on the global state of the contract. Hence repeating same test twice was causing issues with double state update. 

Also didn't use scoped "it" to separate the two because it was causing the test suite to break _my guess_ is that it was causing sync issues where one test ran before. So, It was important that this ran sequentially.

The actual changes are between lines 431 and 488

yondonfu (approved):
LGTM

PR #611, opened 2023-05-31
spike: Snapshot-based TreasuryGovernor

victorges:
This is to demonstrate the implementation described in [this document](https://bit.ly/3IPacYg)
regarding a stake snapshot-based voting power system for an on-chain
treasury fueled by inflationary rewards.

- `BondingManager` changes to mint a % of inflation to a `treasury()` contract
- `BondingManager` changes to notify treasury governor about staking updates
- `TreasuryGovernor` implementation to handle all of the above. Including:
  - Using transcoder recorded `totalStake` in `BondingManager` for historical transcoder stake
  - A checkpointing ("snapshotting") logic of delegator's bonded amount and delegated address
- Also implemented the voting "override" logic for a delegator changing their transcoder vote. `_countVote` function
- Refactor updates: https://github.com/livepeer/protocol/pull/611#issuecomment-1574483763

There are some obvious missing pieces there like integrating this with our `Controller` abstractions
(and whether we should use our proxy or OZ's).

This is all currently untested, which is what I'll be working on now. 

Accept tips on how to test from an existing state like mainnet right now snapshotted to a test env 🤔 

Δ

**Checklist:**


contracts/bonding/TreasuryGovernor.sol
```diff
@@ -0,0 +1,379 @@
...
+        uint256 timepoint = proposalSnapshot(proposalId);
+        (, , address delegatee) = getDelegatorSnapshot(account, timepoint);
+
+        bool isTranscoder = account == delegatee;
+        if (isTranscoder) {
+            // deduce weight from any previous delegators for this transcoder to
+            // make a vote
+            weight = weight - proposalVote.voteDeductions[account];
+        } else {
+            proposalVote.voteDeductions[delegatee] += weight;
```
victorges: Explain logic behind voteDeductions mapping

victorges:
FTR I've moved the code around a bit to resemble what their final organization might look like. 3 new contracts now:
 - `BondingCheckpoints` contains only the checkpointing logic tied to the bonding manager
 - `GovernorVotesBondingCheckpoints` contains governor voting logic based on the above checkpoints
 - `TreasuryGovernor` is the actual governor impl, which is almost empty now

Also added some other comments to the code as I went through it with more info or TODO reminders (feel free to make comments on them in the PR as well)

contracts/treasury/GovernorVotesBondingCheckpoints.sol
```diff
@@ -0,0 +1,214 @@
...
+        controller = _controller;
+        bondingCheckpointsAddr = _bondingCheckpoints;
+    }
+
+    // Voting power module (GovernorVotes)
+
+    /**
+     * @dev Clock is set to match the current round, which is the snapshotting method supported by BondingCheckpoints.
+     */
+    function clock() public view virtual override returns (uint48) {
```
victorges: Consider moving this to bonding checkpoints to be consistent with ERC20Votes

contracts/treasury/GovernorVotesBondingCheckpoints.sol
```diff
@@ -0,0 +1,214 @@
...
+        return totalVotes >= quorum(proposalSnapshot(proposalId));
+    }
+
+    /**
+     * @dev See {Governor-_voteSucceeded}. In this module, the forVotes must be at least QUOTA of the total votes.
+     */
+    function _voteSucceeded(uint256 proposalId) internal view virtual override returns (bool) {
+        (uint256 againstVotes, uint256 forVotes, uint256 abstainVotes) = proposalVotes(proposalId);
+
+        uint256 totalVotes = againstVotes.add(forVotes).add(abstainVotes);
```
victorges: Maybe disconsider abstain votes for quota calculation. Perhaps they should only affect the quorum

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    struct DelegatorCheckpoints {
+        Checkpoints.Trace224 startRounds;
+        mapping(uint256 => DelegatorInfo) snapshots;
+    }
+
+    mapping(address => DelegatorCheckpoints) private delegatorCheckpoints;
+
+    Checkpoints.Trace224 private totalActiveStakeCheckpoints;
```
victorges: We actually call `checkpointTotalActiveStake()` indirectly from `intializeRound()`, which AFAIK is guaranteed to be called on every round as long as the protocol is active. So we could not even use the checkpoints lib here, but just have a mapping from round number to the total active stake and read that.

Should we just go with that simpler approach, or are there any cases where `initializeRound()` may not be called?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,194 @@
...
+    constructor(BondingManager _bondingManager) {
+        bondingManagerAddr = _bondingManager;
+    }
+
+    // We can't lookup the "checkpoint time" from the Checkpoints lib, only the
+    // current value. So instead of checkpointing the bonded amount and
+    // delegatee we snapshot the start round of each delegator change and lookup
+    // the specific values on the separate delegatorSnapshots mapping.
+    // TODO: Consider writing our own checkpoints lib version instead that
+    // stores directly the data we want inline.
```
victorges: I've decided to do this as the curr implementation is a bit weird. Should be simple enough to implement using a separate "snapshotted rounds" array and the same mapping of round to `DelegatorInfo`. Instead of looking up the round on the hacked `Checkpoints.Trace224`, we lookup the round with [`Arrays.findUpperBound()`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Arrays.sol#L24) and then access the map.

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,194 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
```
0xcadams: Should we be using a more recent Solidity version? I haven't kept a close eye on the changes, but I think they made some improvements to the language (like removing `selfdestruct`)

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,194 @@
...
+import "../IController.sol";
+import "../rounds/IRoundsManager.sol";
+import "./BondingManager.sol";
+
+contract BondingCheckpoints {
+    using Checkpoints for Checkpoints.Trace224;
+
+    uint256 public constant MAX_ROUNDS_WITHOUT_CHECKPOINT = 100;
+
+    BondingManager public immutable bondingManagerAddr;
```
0xcadams: I think the pattern in other contracts is always deferring to a lookup to the contract registry first, instead of storing the source of truth in the contract itself. @yondonfu should be able to comment further on that.

contracts/treasury/GovernorVotesBondingCheckpoints.sol
```diff
@@ -0,0 +1,214 @@
...
+            proposalVote.abstainVotes += weight;
+        } else {
+            revert("Votes: invalid value for enum VoteType");
+        }
+    }
+
+    // Helpers for relations with other protocol contracts
+
+    function bondingCheckpoints() public view returns (BondingCheckpoints) {
+        return bondingCheckpointsAddr;
```
0xcadams: Again, here, just calling out that we *may* want this to be upgradeable

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,194 @@
...
+        address _delegateAddress
+    ) public virtual onlyBondingManager {
+        DelegatorCheckpoints storage del = delegatorCheckpoints[_account];
+
+        uint32 startRound = SafeCast.toUint32(_startRound);
+        del.snapshots[startRound] = DelegatorInfo(_bondedAmount, _delegateAddress);
+
+        // now store the startRound itself in the startRounds checkpoints to
+        // allow us to lookup in the above mapping
+        del.startRounds.push(startRound, startRound);
```
0xcadams: This on first glance feels like it could save gas with a different data structure

contracts/bonding/BondingManager.sol
```diff
@@ -538,6 +597,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Update bonded amount
         del.bondedAmount = currentBondedAmount.add(_amount);
 
+        checkpointDelegator(_owner, del);
```
yondonfu: If a delegator increases its bondedAmount in the current round, the additional amount will be reflected in the transcoder's total active stake for the *next round* (see `increaseTotalStake()` and how it updates the totalStake field for the earnings pool in the next round which tracks the total active stake for the transcoder in the next round). Since we are setting voting power = active stake, we should make sure that the checkpointed bondedAmount value is mapped to the current round + 1 as opposed to the current round.

Generally stake updates in the current round are reflected in the active stake of the next round so attention should around this fact should be paid in other functions that handle stake updates i.e. `unbond()`, `rebond()`, etc.

contracts/bonding/BondingCheckpoints.sol
victorges: Yeah I haven't set up the new contracts on the whole controller/manager yet. I've added some TODOs to make that clear it's a spike-only thing on BondingManager but missed adding them here

contracts/bonding/BondingCheckpoints.sol
victorges: Done here: https://github.com/livepeer/protocol/pull/611/commits/1cbe31d866e5bc8c6ba29cb97f05446093b33ffa

contracts/bonding/BondingCheckpoints.sol
victorges: Done here: https://github.com/livepeer/protocol/pull/611/commits/1cbe31d866e5bc8c6ba29cb97f05446093b33ffa

contracts/bonding/BondingCheckpoints.sol
victorges: Yeah, the checkpoints library was a little weird here cause it doesn't really fit our needs: 
 - It only allows key+values of 32 + 224 bits, or 96 + 160 bits, meaning we have to downcast all of the values from round numbers to stake amounts
 - It only returns the "value", not the "key", so we have to use the `startRound` also as the key, to be able to properly calculate the historical rewards between delegation `startRound` and the proposal round (`_timepoint` in some of these funcs)

So I've just migrated this to use a dedicated sorted `uint256[]` instead, which I believe renders a less confusing, more efficient and reliable implementation: https://github.com/livepeer/protocol/pull/611/commits/1cbe31d866e5bc8c6ba29cb97f05446093b33ffa

contracts/bonding/BondingManager.sol
victorges: Thanks for explaining. I was aware of this in the protocol level and had noticed this in some places in the code, but thought it was missing in others and was confused about what should really be the logic here (I don't remember where  it was tho, so it probably was me still confused/understanding the code).

Notice that the `checkpointDelegator` doesn't use `currentRound()` as the checkpoint round though, but actually the greatest of `del.startRound` (which is always set to `currentRound() + 1`) or `del.lastClaimRound` , so I believe this should already be sorted in most cases. In cases where `startRound` is not updated, like `unbond()`, it will use the `lastClaimRound` which must be updated (to curr round) before making any changes to `bondedAmount`. I'll still triple check some of these checkpoints though, I think I might be missing a +1 especially on the `lastClaimRoiund` paths 🤔 

Regarding the `checkpointDelegator` abstraction: An argument could be made that it is a little too "implicit" now though, with the checkpoint round being defined magically inside `checkpointDelegator`. It could be the case to remove that "lastClaimRound or startRound" logic from it and obligate every caller to specify what round the checkpoint is being made on. This might make the whole logic more explicit everywhere, which is probably preferable for Solidity code even if a little repetitive. Would you you say that would be better?

contracts/bonding/BondingCheckpoints.sol
victorges: Done as I was getting into tests :)

contracts/bonding/BondingManager.sol
victorges: Going further on this and as I did the first tests I came to realize: we should probably ignore the `startRound` for the sake of reward calculation and always grab the `lastClaimRound` instead, or rather just `claimRound`. The `startRound` is actually handled a little differently than the rewards calculation, and to get the historical stake calculated the same way we need to calculate rewards from bonding using the `lastClaimRound` value in the delegator, instead of the `startRound` (which is ignored during rewards calculation).

During an initial bond, the `lastClaimRound` value is set to the current round (not the round+1!) and only the `startRound` is set to round+1. The pending stake is then calculated with `lastClaimRound` [here](https://github.com/livepeer/protocol/blob/f1a6953b2fe952d447eddee0ecd78e8289604f42/contracts/bonding/BondingManager.sol#L1246) in which even though it looks like it considers `lastClaimRound+1`, it actually `.sub(1)`s below). I don't know if this causes any issues in the bigger picture or not, but it was good to learn and it actually simplified this implementation a bit.

The behavior in the code was changed here: https://github.com/livepeer/protocol/pull/611/commits/f1a6953b2fe952d447eddee0ecd78e8289604f42#diff-dccab47588659a417db0f5d573aa0c5c29b97bf949ffbba7a9d825e35f0bfcf7R594

yondonfu: > This might make the whole logic more explicit everywhere, which is probably preferable for Solidity code even if a little repetitive. Would you you say that would be better?

I do feel like having the caller explicitly pass in the round would help with readability and I see that you've made that update already 👍.

> Going further on this and as I did the first tests I came to realize: we should probably ignore the startRound for the sake of reward calculation and always grab the lastClaimRound instead, or rather just claimRound.

I think this is basically right, but just to clarify - we want to keep track of the round that a delegator's data (looks like we specifically care about `bondedAmount` and `delegateAddress` right now for the TreasuryGovernor) was updated. So, in the context of any BondingManager function that updates the delegator's data (i.e. all staking functions `bond()`, `unbond()`, `rebond()`, etc.) this round would be the current round which also happens to be the value of the delegator's `lastClaimRound` as of the execution of the staking function since the `autoClaimEarnings` modifier is applied to all staking functions.

So, given the following function signature:

```
function checkpointDelegator(
    address _account,
    uint256 _round,
    address _bondedAmount,
    address _delegateAddress
) external onlyBondingManager;
```

the semantics of calling `checkpointDelegator()` would be that BondingCheckpoints will store a checkpoint containing the delegator's data (specifically its `bondedAmount` and `delegateAddress` fields) as of round number `_round` so that BondingCheckpoints can lookup the most recent delegator's data as of round number `_round`.

yondonfu:
I don't recall if this was explicitly covered elsewhere, but it is also important to define whether the snapshot of voting power is taken at the start or end of a round.

IIUC:

If the snapshot of voting power is taken at the start of a round N (i.e. the first block of the round) then we should fetch the delegator's stake as of the end of round N - 1. This implies fetching the most recent delegator checkpoint *before* round N and setting the end round for the cumulative earnings calculation to round N - 1.

If the snapshot of voting power is taken at the end of a round N then we should fetch the delegator's stake as of the end of round N. This implies fetching the most recent delegator checkpoint *before or equal to* round N and setting the end round for the cumulative earnings calculation to round N.

contracts/bonding/BondingCheckpoints.sol
victorges: Yeah ideally we should, but upgrading this was a real rabbit hole. Since our project is already using 0.8.9 for now, we'd need to upgrade all contracts to accept the new version or at least `^0.8.9` (not fixed) instead to be able to create a new contract that references them. This would be so much changes that I gave up and just kept them as 0.8.9 for now.

contracts/bonding/BondingVotes.sol
```diff
@@ -55,28 +55,33 @@ contract BondingCheckpoints is ManagerProxyTarget, IBondingCheckpoints {
      * for a given round, find the checkpoint with the highest start round that is lower or equal to the queried round
      * ({SortedArrays-findLowerBound}) and then fetch the specific checkpoint on the data mapping.
      */
-    struct BondingCheckpointsByRound {
+    struct BondingVotesByRound {
```
victorges: Should revert this rename

contracts/bonding/BondingVotes.sol
```diff
@@ -117,33 +188,68 @@ contract BondingCheckpoints is ManagerProxyTarget, IBondingCheckpoints {
...
+        bool wasTranscoder = previousDelegate == _account;
+        if (isTranscoder) {
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, current.delegatedAmount);
+        } else if (wasTranscoder) {
+            // if the account stopped being a transcoder, we want to emit an event zeroing its "delegate votes"
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, 0);
+        }
+
+        // Always send a delegator events since transcoders are underlying delegators to themselves. Notice that by the
+        // nature of this event, delegators will only have their claimed stake checkpointed, without pending rewards.
```
victorges: Can be clarified

contracts/treasury/IVotes.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts-upgradeable/interfaces/IERC5805Upgradeable.sol";
+
+interface IVotes is IERC5805Upgradeable {
+    /**
+     * @dev Emitted when bonding change results in changes to a delegate's number of votes. This complements the events
```
victorges: typo delegates

src/test/BondingVotesStateInitialization.sol
```diff
@@ -4,44 +4,46 @@ import "ds-test/test.sol";
...
 import "contracts/rounds/RoundsManager.sol";
 import "contracts/token/LivepeerToken.sol";
 import "contracts/snapshots/MerkleSnapshot.sol";
 import "./interfaces/ICheatCodes.sol";
 import "./interfaces/IL2Migrator.sol";
 
-// forge test --match-contract BondingCheckpointsStateInitialization --fork-url https://arbitrum-mainnet.infura.io/v3/$INFURA_KEY -vvv --fork-block-number 110930219
-contract BondingCheckpointsStateInitialization is GovernorBaseTest {
+// forge test --match-contract bondingVotesStateInitialization --fork-url https://arbitrum-mainnet.infura.io/v3/$INFURA_KEY -vvv --fork-block-number 110930219
+contract bondingVotesStateInitialization is GovernorBaseTest {
```
victorges: lowercase

victorges:
We are done with this phase!

PR #613, opened 2023-06-07
bonding: Remove artificial treasury stake

victorges:
This is another simpler implementation for the treasury rewards calculation,
mixing #611 and #612. The caveat is that we will have a bit more rounding errors
on this approach, since there are more perc calculations. The losses might be 
relatively trivial tho (worst case 1 fractional unit of LPT for each transcoder, so up to
100/1**18 LPT per round).

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -98,8 +98,6 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
     // Rate of global rewards that will be sent to the treasury
     uint256 public treasuryRewardCutRate;
-    // Value to the reward distribution code t
-    uint256 public currentRoundTreasuryArtificialStake;
 
     // TODO: Move these to controller
     BondingCheckpoints public checkpointsAddr;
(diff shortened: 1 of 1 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -866,17 +853,18 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
         // rewardTokens = (current mintable tokens for the round * active transcoder stake) / (total active stake + treasury stake adjustment)
-        uint256 rewardTokens = minter().createReward(
-            earningsPool.totalStake,
-            currentRoundTotalActiveStake + currentRoundTreasuryArtificialStake
-        );
+        uint256 totalRewardTokens = minter().createReward(earningsPool.totalStake, currentRoundTotalActiveStake);
+
+        // The wholesome of treasury rewards were already minted on round initialization, only calculate them to subtract.
+        uint256 treasuryRewards = PreciseMathUtils.percOf(totalRewardTokens, treasuryRewardCutRate);
+        uint256 transcoderRewards = totalRewardTokens.sub(treasuryRewards);
```
victorges, line 860: Might want to invert this calculation (with a similar "cut rate complement") so we dont **over**mint instead of **under**mint

victorges:
We've decided to go with #611 instead given the `reward()` call requirement is not that big of an issue, especially since by avoiding the "lump sum" minting to the treasury we will also make sure that our effective inflation rate does not change with this upgrade. 
e.g.: on an extreme case, if no transcoder calls `reward`, we get no effective inflation regardless of what the protocol inflation is set to. If we had a lump sum mint to the treasury on every round start, we would have an effective inflation of at least that on every round, which changes the expectations on how the inflation is actually going to work.

PR #614, opened 2023-06-17
bonding: Implement active stake checkpointing

victorges:
This is to introduce a checkpointing logic in bonding manager so that we can lookup historical information about any point in time. We need this in order to support an on-chain governance system with support for OpenZeppelin Governor abstractions. That is why it also implements the `IERC5808`  interface to plug into OZ Governor framework.

It is implemented as "external" to `BondingManager` as possible in a separate `BondingVotes` contract, but there were a couple places where `BondingManager` needed changes to start checkpointing its state.

- Implement historical active stake snapshotting logic in the `BondingVotes` contract
- Make `BondingManager` call it to checkpoint the state where it needs to
- Tests for the checkpointing logic (unit / integration / fork)

 - Unit and integration tests with `yarn test`
 - Fork test with forge

Implements PRO-28

**Checklist:**


contracts/bonding/BondingManager.sol
```diff
@@ -1134,29 +1195,49 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+     * calculations with BondingCheckpoints.
+     * @param _startPool The earning pool from the start round for the start cumulative factors. Normally this is the
+     * earning pool from the {Delegator-lastclaimRound}+1 round, as the round where `bondedAmount` was measured.
+     * @param _endPool The earning pool from the end round for the end cumulative factors
+     * @param _stake The delegator initial stake before including earned rewards. Normally the {Delegator-bondedAmount}
+     * @param _fees The delegator's initial fees before including earned fees
+     * @return cStake , cFees where cStake is the delegator's cumulative stake including earned rewards and cFees is the
+     * delegator's cumulative fees including earned fees
+     */
+    function delegatorCumulativeStakeAndFees(
```
victorges: Should we move this memory/lib version of the func to `EarningsPool(LIP36)?` perhaps?

test/unit/BondingCheckpoints.js
```diff
@@ -0,0 +1,1068 @@
...
+            if (Array.isArray(obj[key])) {
+                new Assertion(obj[key]).to.deep.eq(expected[key])
+            } else {
+                new Assertion(obj[key]).to.eq(expected[key])
+            }
+        })
+    })
+})
+
+describe("BondingCheckpoints", () => {
```
victorges: Realized that these are currently a mix of unit and integration tests. Will be moving and refactoring things around for more well-defined tests

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,351 @@
...
+        return SafeCast.toUint48(roundsManager().currentRound());
+    }
+
+    /**
+     * @notice Machine-readable description of the clock as specified in EIP-6372.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function CLOCK_MODE() public pure returns (string memory) {
+        // TODO: Figure out the right value for this
+        return "mode=livepeer_round&from=default";
```
victorges: After reading through [this](https://eips.ethereum.org/EIPS/eip-6372#clock_mode):
> If operating using any other mode, then this function SHOULD return a unique identifier for the encoded mode field.

I'll just change this to `mode=livepeer_round` (same thing without "from"). We can discuss in here if there's any debate we might wanna have about it.

victorges:
@yondonfu @0xcadams Ok I think NOW this one is ready to go! :)

contracts/bonding/BondingManager.sol
victorges: Done as it made a lot of sense for testability as well. I didn't want to mock this function just because it happened to be in the `BondingManager`. It makes a lot of conceptual sense as well, since it implements the logic defined in LIP-36, so all the more sense for it to be in `EarningsPoolLIP36`.

victorges:
Ran the gas report on `confluence` branch (`current`) compared to this (`checkpoints`). Here are the reports, it seems to have increased around 200k gas on the bond/unbond functions due to the checkpoints that have been made. The checkpoint initialization itself also costs around 170k gas, which should still be relatively cheap on Arbitrum, but I haven't found a way to reliably calculate how it would cost considering the calldata cost on L1.

[gas-report-checkpoints.txt](https://github.com/livepeer/protocol/files/12072152/gas-report-checkpoints.txt)
[gas-report-current.txt](https://github.com/livepeer/protocol/files/12072153/gas-report-current.txt)

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts/utils/Arrays.sol";
```
victorges: This is unused in this file, reminder to remove. (used only on `SortedArrays`)

contracts/bonding/BondingManager.sol
```diff
@@ -548,6 +560,40 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        );
+    }
+
+    /**
+     * @notice Checkpoints the bonding state for a given account.
+     * @dev This is to allow checkpointing an account that has an inconsistent checkpoint with its current state.
+     * Implemented as a deploy utility to checkpoint the existing state when deploying the BondingCheckpoints contract.
+     * @param _account The account to initialize the bonding checkpoint for
+     */
+    function checkpointBondingState(address _account) external {
```
0xcadams: Doesn't this allow anyone to trigger checkpointing?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    /**
+     * @dev Checkpoints by account (delegators and transcoders).
+     */
+    mapping(address => BondingCheckpointsByRound) private bondingCheckpoints;
+
+    /**
+     * @dev Rounds in which we have checkpoints for the total active stake. This and {totalActiveStakeCheckpoints} are
+     * handled in the same wat that {BondingCheckpointsByRound}, with rounds stored and queried on this array and
```
0xcadams: Typo here :)
```suggestion
     * handled in the same way that {BondingCheckpointsByRound}, with rounds stored and queried on this array and
```

contracts/bonding/libraries/SortedArrays.sol
```diff
@@ -0,0 +1,85 @@
...
+
+import "../../libraries/MathUtils.sol";
+
+import "@openzeppelin/contracts/utils/Arrays.sol";
+
+/**
+ * @title SortedArrays
+ * @dev Handles maintaining and looking up on sorted uint256 arrays.
+ */
+library SortedArrays {
```
0xcadams, line 12: Is there any reusable, audited version of this? Seems like something that should be an openzeppelin contract to reduce surface area

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+     * transcoders this also includes all the amount delegated towards them by other delegators.
+     * @return delegateAddress The address the account delegated to. Will be equal to _account in case of transcoders.
+     */
+    function getBondingStateAt(address _account, uint256 _round)
+        public
+        view
+        virtual
+        returns (uint256 amount, address delegateAddress)
+    {
+        BondingCheckpoint storage bond = getBondingCheckpointAt(_account, _round);
```
0xcadams: I think this can be `memory` vs `storage` to save gas costs

0xcadams:
```suggestion
        BondingCheckpoint memory bond = getBondingCheckpointAt(_account, _round);
```

0xcadams: Same across the board for a lot of the view functions

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+     * at or before the given round and using the checkpoint of that round. If there hasn't been checkpoints since then
+     * it means that the state hasn't changed.
+     * @param _account The account whose bonding state we want to get.
+     * @param _round The round for which we want to get the bonding state.
+     * @return The {BondingCheckpoint} pointer to the checkpoints storage.
+     */
+    function getBondingCheckpointAt(address _account, uint256 _round)
+        internal
+        view
+        returns (BondingCheckpoint storage)
```
0xcadams: Same here - I *think* this can be memory since this value isn't modified anywhere

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+     * @dev Return BondingManager interface
+     */
+    function bondingManager() internal view returns (BondingManager) {
+        return BondingManager(controller.getContract(keccak256("BondingManager")));
+    }
+
+    /**
+     * @dev Return IRoundsManager interface
+     */
+    function roundsManager() public view returns (IRoundsManager) {
```
0xcadams: Should this be public vs the internal `boundingManager()` above?

0xcadams (commented):
Preliminary review of code quality & gas improvements, no deep checks of logic

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+ * @dev Checkpointing logic for BondingManager state for historical stake calculations.
+ */
+contract BondingCheckpoints is ManagerProxyTarget, IBondingCheckpoints {
+    using SortedArrays for uint256[];
+
+    constructor(address _controller) Manager(_controller) {}
+
+    struct BondingCheckpoint {
+        /**
+         * @dev The amount of bonded tokens to another delegate as per the lastClaimRound.
```
yondonfu:
```suggestion
         * @dev The amount of bonded tokens to another delegate as of the lastClaimRound.
```

contracts/bonding/BondingManager.sol
```diff
@@ -524,6 +531,9 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             if (currPool.cumulativeRewardFactor == 0) {
                 currPool.cumulativeRewardFactor = cumulativeFactorsPool(newDelegate, newDelegate.lastRewardRound)
                     .cumulativeRewardFactor;
+                if (currPool.cumulativeRewardFactor == 0) {
```
yondonfu: Can we document why this line is needed?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+         * have to self-delegate first and then have tokens bonded from other delegators.
+         */
+        uint256 delegatedAmount;
+        /**
+         * @dev The last round during which the delegator claimed its earnings. This pegs the value of bondedAmount for
+         * rewards calculation in {EarningsPoolLIP36-delegatorCumulativeStakeAndFees}.
+         */
+        uint256 lastClaimRound;
+        /**
+         * @dev The last round during which the transcoder called {BondingManager-reward}. This is needed to find a
```
yondonfu: Can we document whether this field is the `lastRewardRound` value for the address that this checkpoint is for vs. the address' delegate?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+    /**
+     * @dev Rounds in which we have checkpoints for the total active stake. This and {totalActiveStakeCheckpoints} are
+     * handled in the same wat that {BondingCheckpointsByRound}, with rounds stored and queried on this array and
+     * checkpointed value stored and retrieved from the mapping.
+     */
+    uint256[] totalStakeCheckpointRounds;
+    /**
+     * @dev See {totalStakeCheckpointRounds} above.
+     */
+    mapping(uint256 => uint256) private totalActiveStakeCheckpoints;
```
yondonfu: A few observations:

- The general abstraction is a wrapper struct like `BondingCheckpointsByRound` that contains a sorted uint256 array of round numbers and a mapping of each round number to the data type that was checkpointed for that round
- The state variables for the above can be kept private

Given the above, rather than maintaining the separate `totalStakeCheckpointRounds` and `totalActiveStakeCheckpoints` state variables and having to maintain the visibility of both (note that `totalStakeCheckpointRounds` is internal by default right now vs. `totalActiveStakeCheckpoints` which is private), what if a single wrapper struct `TotalActiveStakeCheckpointsByRound` that follows the same convention as `BondingCheckpointsByRound` was used? Ex:

```
struct TotalActiveStakeCheckpointsByRound {
    uint256[] rounds;
    mapping (uint256 => uint256) data;
}

TotalActiveStakeCheckpointsByRound private totalActiveStakeCheckpoints;
```

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+        }
+        return array.findLowerBound(value);
+    }
+
+    // Manager/Controller helpers
+
+    /**
+     * @dev Modified to ensure the sender is BondingManager
+     */
+    modifier onlyBondingManager() {
```
yondonfu: Let's move this after the state variable declarations since that has been the convention.

contracts/bonding/libraries/SortedArrays.sol
```diff
@@ -0,0 +1,85 @@
...
+     * @param val Value to push into array. Must be greater than or equal to the highest (last) element.
+     */
+    function pushSorted(uint256[] storage array, uint256 val) internal {
+        if (array.length == 0) {
+            array.push(val);
+        } else {
+            uint256 last = array[array.length - 1];
+
+            // values must be pushed in order
+            require(val >= last, "pushSorted: decreasing values");
```
yondonfu: Custom error type instead of require statement here?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+        if (pool.cumulativeRewardFactor == 0) {
+            revert MissingEarningsPool(_transcoder, _round);
+        }
+    }
+
+    /**
+     * @dev Helper to return more helpful custom errors in case of bad queries.
+     */
+    function checkedFindLowerBound(uint256[] storage array, uint256 value) internal view returns (uint256) {
+        if (array.length == 0) {
```
yondonfu: Seems redundant if the check is already in `findLowerBound()`.

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+    /**
+     * @notice Called by the BondingManager when the total active stake changes.
+     * @dev This is called only from the {BondingManager-setCurrentRoundTotalActiveStake} function to set the total
+     * active stake in the current round.
+     * @param _totalStake From {BondingManager-currentRoundTotalActiveStake}
+     * @param _round The round for which the total active stake is valid. This is normally the current round.
+     */
+    function checkpointTotalActiveStake(uint256 _totalStake, uint256 _round) public virtual onlyBondingManager {
+        if (_round > clock()) {
+            revert FutureCheckpoint(_round, clock());
```
yondonfu: May want to have a separate error type for `checkpointTotalActiveStake()` vs. `checkpointBondingState()` because the requirements for each function are different and it may not be immediately clear just based on the error emitted which requirement was violated. In theory, you could check which function emitted the error, but having separate error types may be cleaner and avoids that extra lookup.

contracts/bonding/BondingManager.sol
```diff
@@ -799,6 +848,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Set last round that transcoder called reward
         t.lastRewardRound = currentRound;
 
+        checkpointBondingState(msg.sender, delegators[msg.sender], t);
```
yondonfu: Since `updateTranscoderWithRewards()` calls `increaseTotalStake()` which calls `checkpointBondingState()` for the msg.sender in this function, we have two internal calls to `checkpointBondingState()` for the msg.sender in this function - once before `t.lastRewardRound` is updated to the current round and once after.

If L849 is moved into `updateTranscoderWithRewards()` right before the `increaseTotalStake()` call (L1424) I think it would be possible to remove this extra call to `checkpointBondingState()`.

contracts/bonding/BondingManager.sol
```diff
@@ -1441,6 +1484,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         // Rewards are bonded by default
         del.bondedAmount = currentBondedAmount;
         del.fees = currentFees;
+
+        checkpointBondingState(_delegator, del, transcoders[_delegator]);
```
yondonfu: This will result in 2 calls to `checkpointBondingState()` in every bonding related function that uses the `autoClaimEarnings()` modifier or the `_autoClaimEarnings()` internal function. Could we add this line to just the end of `claimEarnings()` instead to avoid this?

contracts/bonding/BondingCheckpoints.sol
yondonfu: I think using `memory` here would actually increase gas costs because `getBondingCheckpointAt()` currently returns a storage pointer to a struct which I believe defers gas cost consumption for storage reads until the individual fields of the struct are read. If we use `memory` here I believe the values of the storage fields of the struct would need to all be read and written into memory.

In functions where we need to read the values of the struct fields infrequently I think it makes sense to continue using a storage pointer because the gas overhead from reading/copying all struct fields from storage to memory likely outweighs the gas cost of reading the field values direct from storage. In functions where we need to read the values of the struct fields super frequently then it might make sense to read/copy all struct fields from storage to memory because the gas overhead could be small relative to the savings from repeatedly reading from memory instead of storage.

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+    }
+
+    /**
+     * @dev Helper to return more helpful custom errors in case of bad queries.
+     */
+    function checkedFindLowerBound(uint256[] storage array, uint256 value) internal view returns (uint256) {
+        if (array.length == 0) {
+            revert NoRecordedCheckpoints();
+        } else if (array[0] > value) {
+            revert PastLookup(value, array[0]);
```
yondonfu: Should this be folded into SortedArrays since it seems that a bunch of checks are already implemented there?

contracts/bonding/BondingManager.sol
```diff
@@ -548,6 +560,40 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            _delegator.delegatedAmount,
+            _delegator.lastClaimRound,
+            _transcoder.lastRewardRound
+        );
+    }
+
+    /**
+     * @notice Checkpoints the bonding state for a given account.
+     * @dev This is to allow checkpointing an account that has an inconsistent checkpoint with its current state.
+     * Implemented as a deploy utility to checkpoint the existing state when deploying the BondingCheckpoints contract.
```
yondonfu: Questions:

1. Is there a description of how this is supposed to work?
2. Is the intent to call this function for every address? If so:
    a. Do function calls against the BondingManager need to be prevented until this is complete?
    b. What is the cost estimate?

contracts/bonding/BondingCheckpoints.sol
```diff
@@ -0,0 +1,393 @@
...
+        BondingCheckpointsByRound storage checkpoints = bondingCheckpoints[_account];
+
+        // Most of the time we will be calling this for a transcoder which checkpoints on every round through reward().
+        // On those cases we will have a checkpoint for exactly the round we want, so optimize for that.
+        BondingCheckpoint storage bond = checkpoints.data[_round];
+        if (bond.bondedAmount > 0) {
+            return bond;
+        }
+
+        if (checkpoints.startRounds.length == 0) {
```
yondonfu: If there are no checkpoints (i.e. startRounds.length == 0) and L255 evaluates to false then it looks like we'll execute L262 which would revert because startRounds.length == 0? This would mean that calling this function for an address without a checkpoint yet (which would be the case immediately post-upgrade for any address that doesn't execute a tx for a bonding action [1]) would cause a revert.

[1] Unless there is a mechanism that guarantees the address has a checkpoint (see my other question about the use of BondingManager.checkpointBondingState()) or that can handle addresses without a checkpoint yet.

yondonfu (changes requested):

contracts/bonding/BondingCheckpoints.sol
victorges: Not anymore hah

contracts/bonding/BondingManager.sol
victorges: Yeah it does, and that's the idea! The checkpoint process is safe, it will only store the current state of the account in the checkpoints contract (now `BondingVotes`). It will be used during state initialization once we deploy this for the first time, and can be used in the future to correct any potential discrepancies that might show up in the checkpointed state compared to the `BondingManager` state. By having this callable for anyone, we also allow people to make sure other addresses have a consistent checkpoint (or even initialize them on the first round).

contracts/bonding/libraries/SortedArrays.sol
victorges: The reusable version of this is the `Arrays.findUpperBound`, which we do use internally for the more complex logic of actually implementing a binary search. Here we just adapt the "finde upper" to a "find lower" instead which is what we need for the bonding checkpoints. I might be making the interfaces even more similar right next, like making this function return just the index in the same way that OpenZeppelin's interface does it. Will make error checking less redundant in the checkpoint accessing logic.

contracts/bonding/BondingCheckpoints.sol
victorges: As a sanity check, I ran the gas reports with [the `memory` version](https://github.com/livepeer/protocol/pull/611/commits/e74f3a4e879dc766b27d4515990f119ba821c93d) for these variables and return values, and the cost of `checkpointBondingState` doesn't change much as expected, while the `getBondingStateAt` function increases gas cost in up to 10k (or 7k on average for delegator/transcoder/nonparticipant). So it is in fact more efficient to keep it as a storage so we don't have to copy the whole struct back to memory.

Attaching the report outputs here FTR.
[gas-memory.txt](https://github.com/livepeer/protocol/files/12361511/gas-memory.txt)
[gas-storage.txt](https://github.com/livepeer/protocol/files/12361512/gas-storage.txt)

contracts/bonding/BondingManager.sol
victorges: Fixed on https://github.com/livepeer/protocol/pull/614/commits/0c41f78401a97ba4d92def55ba21f2c1820ab6b2

On the first implementation I didn't think that this would be a strict requirement, but as I progressed on it and especially after learning about the events interfaces it made a lot more sense to make sure we checkpoint only once per transaction.

The thinking of the original implementation was "whenever the storage is mutated, make sure we checkpoint afterwards" which resulted in the duplicate checkpoints mentioned. I changed this rationale to start from the high-level transaction instead, to make sure that they either checkpoint the state in the end themselves or call some other function (but only once) that checkpoints for them instead.
- Creating a modifier for the checkpointing process made it easier as well so it shows up in the signature of functions that do checkpoint and we can reason about it more quickly.
- There was the need of only 1 function to have an alternate version that doesn't make checkpointing, which I added with the `Uncheckpointed` suffix
- Good thing is I had tests for all those calls already and could verify that all of them were fixed, with only 1 checkpoint per address per tx now

contracts/bonding/BondingManager.sol
victorges: Also fixed on https://github.com/livepeer/protocol/pull/614/commits/0c41f78401a97ba4d92def55ba21f2c1820ab6b2 as mentioned in the other comment

contracts/bonding/BondingManager.sol
victorges: This was meant to account for the case where the delegator bonds to a transcoder that has never called reward. In that case, the `cumulativeRewardsFactor` of the round where the bond was made would stay zeroed and that would affect the pending rewards calculation on the checkpoints contract.

After thinking a bit further about this, also related to the other things we've talked about, I've decided to undo this change to `BondingManager`. Will be good to reduce the amount of changes here and instead have the checkpointing code resilient to such imperfect states that might exist in the bonding manager.

This is more aligned with that idea of not failing on read transactions as much as possible, assuming a default uninitialized state instead. I'm compiling a commit to fix this issue together with the whole mindset shift to return 0 for un-checkpointed accounts or when querying before their first checkpoint. This would make the contract better to use, which I've learned the hard way while trying to interact with the current implementation on the testnet. Will send a detailed explanation of the bigger changes here after I update the PR!

victorges: Reverted this change on https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3

contracts/bonding/BondingCheckpoints.sol
victorges: Fixed on https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3

contracts/bonding/BondingCheckpoints.sol
victorges: Fixed on https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3
Now we just return 0 balance for any uncheckpointed state

contracts/bonding/BondingCheckpoints.sol
victorges: Fixed on https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3

contracts/bonding/BondingManager.sol
victorges: I'll answer shortly here for the record, but let's continue the discussion on Discord.

1. I could swear I had included in the spec but apparently not. Need to write this down in some places like LIP and docs.
2. Yeah, to call it for every address that is participating in the protocol, which is around ~4000 from our explorer rn.
   a. They don't! Any action on BondingManager will automatically make a checkpoint as well and has the same end result of calling this function directly.
   b. Around ~200k gas per checkpoint call. The L2 cost seems negligible and something we could pay for, but I don't know much about L1 calldata costs yet.
   
Uncheckpointed accounts used to not be queryable, but after https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3 they just return a 0 voting power until checkpointing happens. If the account hasn't been checkpointed, it will have no voting power. So it is in the best interest of the account to have this checkpoint if they intend to vote, and we can even leave for users themselves to call this instead.

So there are some levels of investment we could make which would be fine, just need proper communication with users:
- We can checkpoint all the participant accounts and leave no actions required from users
- We can checkpoint only the active transcoders (they get their whole voting power from that, dont need each delegator to do it) and leave the task for delegators that want to guarantee their voting power to either call this function or make any bonding change (including `claimRewards`, which is the easiest)
- We checkpoint no account ourselves and have an integration in the explorer to call this function if needed
- We checkpoint no account at all and just let users automatically do it when making changes to BondingManager (`reward()`, `claimEarnings()`, etc)

victorges:
OK @yondonfu @0xcadams I've addressed the feedback on your comments and this is ready for another review round. I have included some other changes as well that seemed necessary after the rounds of testing with the testnet and some discussions.

I recommend reviewing commit by commit, I avoided making a force push here not to lose history, and also isolated the less trivial changes in their own commits. As an overview of each:
- https://github.com/livepeer/protocol/pull/614/commits/0c41f78401a97ba4d92def55ba21f2c1820ab6b2: This addresses the comments about multiple checkpoints being made for the same account in the same transaction. It makes sure we only call it once in the end of all mutations instead on all code paths.
- https://github.com/livepeer/protocol/pull/614/commits/d518cd6058e87c7bc10c8ecfc05696e8c30655e8: This merges the `BondingCheckpointsVotes` contract from https://github.com/livepeer/protocol/pull/615 into the `BondingCheckpoints` contract here (renamed to just `BondingVotes`). This is mainly to allow it to send the `ERC5805` events on changes.
- https://github.com/livepeer/protocol/pull/614/commits/ccc3ddf071a4c7b63fe22ca9531118b858a8b390: This is a change on the way that we assign voting power to accounts, to use their active stake at the **end** of the round instead of **start**. I didn't change any of the core checkpointing logic, just changed the voting power functions to read state from the next round instead, also using the `nextRoundTotalActiveStake` when querying for the next round to the current one.
  - After further testing with the OZ governor, I came to understand that their definition of "voting power at a timepoint" actually refers to the voting power **at the end** of that timepoint. This makes sense if you think about using block numbers as the clock, and is also reflected in the way that the OZ Governor works for some things like vote snapshotting: voting starts only **after** the "snapshot round", not **on** it.
  - A practical effect of this is that, in the previous implementation, if we had a `votingDelay` of 1 round the voting would be delayed in 1-2 rounds, but users would only really have 0-1 rounds to make any staking changes. Any staking changes made on the round before the voting actually starts would not be effective, since we would take the active stake from the start of that round for the voting power. Now we take it from the end (start of next round), and things work a bit more as expected.
- https://github.com/livepeer/protocol/pull/614/commits/d5062fe2db7d5037988c2f1a5989886d7ce087d3: This changes the behavior of most of the "read" logic in the `BondingVotes` contract to avoid reverting queries that could simply return a 0 value. There is only 1 specific case now where we will revert instead of returning zero, which is when there is no earnings pool on a transcoder's last reward round, which is an illegal state from the way that `BondingManager` works. This also fixed a lot of the redundant `reverts` with the `SortedArrays` implementation.
- https://github.com/livepeer/protocol/pull/614/commits/292f9542610d6f270966570135915a12ce6fd531: This one addresses the other review comments which were simpler so I just combined them in this commit.

contracts/bonding/BondingCheckpoints.sol
victorges: Liked this very much. Included this fix on https://github.com/livepeer/protocol/commit/d518cd6058e87c7bc10c8ecfc05696e8c30655e8

contracts/test/mocks/BondingVotesERC5805Harness.sol
```diff
@@ -0,0 +1,34 @@
...
+pragma solidity 0.8.9;
+
+import "../../bonding/BondingVotes.sol";
+import "./GenericMock.sol";
+
+/**
+ * @dev This is a tets utility for unit tests on the ERC5805 functions of the BondingVotes contract. It overrides the
+ * functions that should be used to derive the values returned by the ERC5805 functions and checks against those.
+ */
+contract BondingVotesERC5805Harness is BondingVotes {
```
victorges, line 11: So the idea here is that the `BondingVotes` implementation has 2 "layers". The first layer deals with checkpointing directly, while the second layer is the logic that was in the old `BondingCheckpointsVotes`. So the tests for that second layer are simpler, just checking that it proxies to the checkpointing logic.

This "harness" works by mocking only that first layer of the implementation, then checking if the second layer, the ERC5805 functions, are depending only on it.

Let me know if you think this testing strategy is too confusing and we need more guarantees here. One idea I tried was creating a separate abstract `BondingVotesERC5805` contract, which got inherited by the main implementation. We could have some more isolated/guaranteed tests and clearer implementation, but it felt like a little overengineering and I reverted that. Let me know if it'd be preferrable, or if you have any other ideas. Should be easy to adapt here.

contracts/bonding/BondingManager.sol
```diff
@@ -123,6 +124,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         _;
     }
 
+    modifier autoCheckpoint(address account) {
```
yondonfu:
```suggestion
    modifier autoCheckpoint(address _account) {
```

to follow convention elsewhere.

contracts/bonding/BondingManager.sol
```diff
@@ -1378,11 +1427,14 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
         earningsPool.updateCumulativeRewardFactor(prevEarningsPool, delegatorsRewards);
         // Update transcoder's total stake with rewards
-        increaseTotalStake(_transcoder, _rewards, _newPosPrev, _newPosNext);
+        increaseTotalStakeUncheckpointed(_transcoder, _rewards, _newPosPrev, _newPosNext);
     }
 
     /**
      * @dev Update a delegator with token pools shares from its lastClaimRound through a given round
+     *
+     * Notice that this function udpates the delegator storage but does not checkpoint its state. Since it is internal
```
yondonfu:
```suggestion
     * Notice that this function updates the delegator storage but does not checkpoint its state. Since it is internal
```

contracts/bonding/BondingManager.sol
```diff
@@ -546,6 +559,43 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+            _delegator.delegatedAmount,
+            _delegator.lastClaimRound,
+            _transcoder.lastRewardRound
+        );
+    }
+
+    /**
+     * @notice Checkpoints the bonding state for a given account.
+     * @dev This is to allow checkpointing an account that has an inconsistent checkpoint with its current state.
+     * Implemented as a deploy utility to checkpoint the existing state when deploying the BondingVotes contract.
```
yondonfu: Based on latest discussion on how checkpointing will work this line seems like it can be removed.

contracts/bonding/BondingManager.sol
```diff
@@ -546,6 +559,43 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+
+        // the `autoCheckpoint` modifier has been replaced with its internal function as a `Stack too deep` error work-around
+        checkpointBondingState(_owner, del, transcoders[_to]);
+    }
+
+    /**
+     * @notice Checkpoints a delegator state after changes, to be used for historical voting power calculations in
+     * on-chain governor logic.
+     */
+    function checkpointBondingState(
```
yondonfu: The convention has been for internal functions to be grouped together later in the file after public functions.

contracts/bonding/BondingManager.sol
```diff
@@ -546,6 +559,43 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         }
 
         emit Bond(_to, currentDelegate, _owner, _amount, del.bondedAmount);
+
+        // the `autoCheckpoint` modifier has been replaced with its internal function as a `Stack too deep` error work-around
+        checkpointBondingState(_owner, del, transcoders[_to]);
```
yondonfu: Shouldn't the last arg be diff?

```suggestion
        checkpointBondingState(_owner, del, transcoders[_owner]);
```

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+        ) = bondingManager().getTranscoderEarningsPoolForRound(_transcoder, _round);
+    }
+
+    // Manager/Controller helpers
+
+    /**
+     * @dev Return BondingManager interface
+     */
+    function bondingManager() internal view returns (BondingManager) {
+        return BondingManager(controller.getContract(keccak256("BondingManager")));
```
yondonfu: Can this be an IBondingManager interface?

yondonfu (commented):
Still need to do the follow up for BondingVotes.sol.

contracts/bonding/BondingManager.sol
victorges: Wow yeah. I'll make sure to add some test checks to ensure this. The only reason it didn't affect them is because we only use the `lastRewardRound` from the `Transcoder` struct right now, which doesn't affect the bonding state calculation for delegators (cause we check the `delegateAddress` before that).

contracts/bonding/BondingManager.sol
victorges: Oh cool, I hadn't noticed this was the pattern. Will update it here and might need in some other places like the `_setTreasuryRewardCut` internal function on https://github.com/livepeer/protocol/pull/616

While at it, should internal functions have a `_` prefix as well? What is the current pattern? I couldn't tell from the existing functions. In the case of `_setTreasuryReward` cut it needs it not to name-clash with the external one tho

contracts/bonding/BondingVotes.sol
victorges: It currently cannot as it doesn't include the 2 functions that we call from it: `nextRoundTotalActiveStake` and `getTranscoderEarningsPool`

What's the process with these interfaces? Should I just add the methods that I need to the interface? They generally don't seem to be extensive of the contracts public functions.

contracts/bonding/BondingManager.sol
victorges: Addressed on https://github.com/livepeer/protocol/pull/614/commits/fecf5a2e0e0cf76472ef1c23d543bc6f9512fb9f together with the test updates

contracts/bonding/BondingManager.sol
yondonfu: IIRC the convention has been to have `_` prefix if the internal function signature would clash with the public function signature. I recall internal functions getting names that describe the logic they contain and in the scenario that the internal function contains almost exactly the same logic as a corresponding public function then the internal function gets the public function name with the `_` prefix.

contracts/bonding/BondingVotes.sol
yondonfu: > What's the process with these interfaces?

In the past, we've just added the external/public functions that need to be called by other protocol contracts.

I believe it would be beneficial to use the IBondingManager interface here if possible as it would avoid importing the entirety of BondingManager.sol which I *think* affects the deployment bytecode size.

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+    constructor(address _controller) Manager(_controller) {}
+
+    // IVotes interface implementation.
+    // These should not access any storage directly but proxy to the bonding state functions.
+
+    /**
+     * @notice Returns the name of the virtual token implemented by this.
+     */
+    function name() external pure returns (string memory) {
+        return "Livepeer Stake";
```
yondonfu: Recommend something more descriptive of what the virtual token is used for like "Voting LPT".

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+     */
+    function delegates(address _account) external view returns (address) {
+        return delegatedAt(_account, clock());
+    }
+
+    /**
+     * @notice Returns the delegate that _account had chosen in a specific round in the past. See `delegates()` above
+     * for more details.
+     * @dev This is an addition to the IERC5805 interface to support our custom vote counting logic that allows
+     * delegators to override their transcoders votes. See {GaovernorCountingOverridable-_handleVoteOverrides}.
```
yondonfu:
```suggestion
     * delegators to override their transcoders votes. See {GovernorCountingOverridable-_handleVoteOverrides}.
```

contracts/bonding/BondingManager.sol
victorges: OK, I think I'll rename the internal function to have the `_` prefix then. Even though the signature doesn't clash cause the arguments are different, I think it's worth the differentiation for clarity. Will also make the `public` one `external` and call the internal one directly from the `modifier`

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+        // Always send delegator events since transcoders are delegators themselves. The way our rewards work, the
+        // delegator voting power calculated from events will only reflect their claimed stake without pending rewards.
+        if (previous.bondedAmount != current.bondedAmount) {
+            emit DelegatorVotesChanged(_account, previous.bondedAmount, current.bondedAmount);
+        }
+    }
+
+    /**
+     * @notice Returns whether an account already has any checkpoint.
+     * @dev This is meant to be called by a checkpoint initialization script once we deploy the checkpointing logic for
```
yondonfu: Seems that this line can be removed given latest thinking on checkpoint initialization by voters prior to voting.

contracts/bonding/BondingVotes.sol
victorges: OK, will do. 

As for the interface I created here, `IBondingVotes`, I've actually included all of the public functions that it exports. Let me know if it'd be preferrable to trim that interface down to only the methods used externally (likely only the checkpoint ones used by `BondingManager`, since the treasury uses only the `IVotes` interface directly.

Should events/errors also be only in the implementation or kept in the interface? Or keep in the interface only the ones emitted from functions in the interface?

contracts/bonding/IBondingVotes.sol
```diff
@@ -0,0 +1,48 @@
...
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts-upgradeable/interfaces/IERC6372Upgradeable.sol";
+
+import "../treasury/IVotes.sol";
+
+/**
+ * @title Interface for BondingVotes
+ */
+interface IBondingVotes is IERC6372Upgradeable, IVotes {
```
victorges: Self-reminder: `IERC6372Upgradeable` is redundant here since it's included in `IVotes`. Remove

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+        bool isTranscoder = newDelegate == _account;
+        bool wasTranscoder = previousDelegate == _account;
+        if (isTranscoder) {
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, current.delegatedAmount);
+        } else if (wasTranscoder) {
+            // if the account stopped being a transcoder, we want to emit an event zeroing its "delegate votes"
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, 0);
+        }
+
+        // Always send delegator events since transcoders are delegators themselves. The way our rewards work, the
```
yondonfu: This event name feels misleading as the event doesn't actually get emitted every time a delegator's voting power changes.

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+
+    /**
+     * @notice Called by the BondingManager when the total active stake changes.
+     * @dev This is called only from the {BondingManager-setCurrentRoundTotalActiveStake} function to set the total
+     * active stake in the current round.
+     * @param _totalStake From {BondingManager-currentRoundTotalActiveStake}
+     * @param _round The round for which the total active stake is valid. This is normally the current round.
+     */
+    function checkpointTotalActiveStake(uint256 _totalStake, uint256 _round) public virtual onlyBondingManager {
+        if (_round > clock()) {
```
yondonfu: Should this be `_round != clock() + 1` to mirror the conditional in `checkpointBondingState()`?

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+     */
+    function name() external pure returns (string memory) {
+        return "Livepeer Stake";
+    }
+
+    /**
+     * @notice Returns the symbol of the token underlying the voting power.
+     */
+    function symbol() external view returns (string memory) {
+        return livepeerToken().symbol();
```
yondonfu: Recommend using a different symbol to distinguish from LPT i.e. "vLPT".

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    /**
+     * @notice Returns the decimals places of the token underlying the voting.
+     */
+    function decimals() external view returns (uint8) {
+        return livepeerToken().decimals();
```
yondonfu: Should be able to avoid referencing the LPT contract here by just returning 18 - doesn't seem necessary to call the LPT contract here.

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+        if (upper == 0) {
+            // Return a zero voting power supply for any round before the first checkpoint. This also happens if there
+            // are no checkpoints at all.
+            return 0;
+        } else if (upper < initializedRounds.length) {
+            // Use the checkpoint from the next initialized round, which got the next total active stake checkpointed.
+            uint256 nextInitedRound = initializedRounds[upper];
+            return totalStakeCheckpoints.data[nextInitedRound];
+        } else {
+            // Here the _round is after any initialized round, so grab its stake from nextRoundTotalActiveStake()
```
yondonfu, line 346: Shouldn't ever get here right due to L325?

contracts/bonding/BondingVotes.sol
yondonfu: > As for the interface I created here, IBondingVotes, I've actually included all of the public functions that it exports

I think that is fine and probably the overall direction to go in moving forward for the reasoning I mention below (though don't think its necessary to update existing interfaces that are already being used immediately to follow that convention and can get to that separately i.e. IBondingManager).

> Should events/errors also be only in the implementation or kept in the interface? Or keep in the interface only the ones emitted from functions in the interface?

I think having events/errors in the interface is a good idea as it makes it easier to just look at the interface contract and know what a consumer of the contract has access to/needs to be aware of in the public API.

contracts/test/mocks/BondingVotesERC5805Harness.sol
yondonfu: Not sure I understand how this is being used from a quick glance. If you link to the relevant areas that this contract is used and the test cases written for it can prob comment on that.

contracts/bonding/BondingVotes.sol
victorges: That makes sense. How about `Livepeer Voting Power`?

contracts/bonding/BondingVotes.sol
victorges: Makes sense. I can move it to the `IBondingVotes` interface (as opposed to the "generic" `IVotes`) and rename it to `DelegatorBondedAmountChanged`, since it is an internal thing from our abstractions anyway

contracts/bonding/BondingVotes.sol
yondonfu: SGTM

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    /**
+     * @notice Returns the amount of votes that `_account` had at a specific moment in the past. If the `clock()` is
+     * configured to use block numbers, this will return the value at the end of the corresponding block.
+     * @dev Keep in mind that since this function should return the votes at the end of the _round (or timepoint in OZ
+     * terms), we need to fetch the bonding state at the next round instead. That because the bonding state reflects the
+     * active stake in the current round, which is the snapshotted stake from the end of the previous round.
+     */
+    function getPastVotes(address _account, uint256 _round) public view returns (uint256) {
```
victorges: FTR of an offline discussion: The ERC5805 specifies that the `getPast...` functions should only be callable with a past timepoint value, so that its results can always be immutable (compared to `getVotes` that returns votes in current round [1]).

So we decided to make that change here and add an additional revert in case this is called for the current round. State from the current round 

[1] This behavior feels even more aligned with the change to return the stake in the end of the round instead of the start. The `getVotes` **is** actually mutable now, which it wasn't when we were returning the stake at the beginning of the round.

contracts/test/mocks/BondingVotesERC5805Harness.sol
victorges: This contract is used in [ERC5805 unit tests here](https://github.com/livepeer/protocol/blob/1692ec11e1cf29d3a26f29092b9e9fe12890b51d/test/unit/BondingVotes.js#L1161).

What feels wrong there is that even though it is a unit test of the `BondingVotes` contract, those tests actually replace the unit under test with another one 😅 
It makes sense for the current version of the code, but it could get confusing when someone else looks at this or maybe makes changes to the "ERC5805-layer".

Separating it to an actual separate "module" on the implementation would allow for a more proper unit test, which could create the "harness" only surrounding that module in a separate unit test [1] (like [on treasury counting module](https://github.com/livepeer/protocol/pull/615/files#diff-ad14a7fcaebacc5393295641e68e2eee7817d93aa51ecbeaeffc5e2f14e2d96aR16)), instead of replacing the whole `BondingVotes` under test with a contract inheriting from it that overrides a couple functions. This is the path that also feels like overengineering though, so I left it like this for now.

Let me know if you have a preference for the separate module and I can change it if so.

[1] e.g.:
```
abstract contract BondingVotesERC5805 is IVotes {
   // implement ERC5805 functions 
   // requires implementation of `clock()`, `getBondingStateAt` and `getTotalActiveStakeAt`
}

contract BondingVotes is IBondingVotes, BondingVotesERC5805 {
  // implements the core checkpointing logic incl the abstract functions that BondingVotesERC5805 wants
}
```
like this we could test `BondingVotesERC5805` more isolatedly (tho no actual difference in the tests themselves, only on the setup)

src/test/BondingVotesStateInitialization.sol
```diff
@@ -0,0 +1,278 @@
...
+            assertTrue(!bondingVotes.hasCheckpoint(_testAddresses[i]));
+        }
+    }
+
+    function testDisallowsQueryingParticipantUncheckpointedAccount() public {
+        uint256 currentRound = ROUNDS_MANAGER.currentRound();
+
+        address[2] memory testAddresses = [DELEGATOR, TRANSCODER];
+        for (uint256 i = 0; i < testAddresses.length; i++) {
+            CHEATS.expectRevert(IBondingVotes.NoRecordedCheckpoints.selector);
```
victorges: These tests are outdated, update

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        bool isTranscoder = newDelegate == _account;
+        bool wasTranscoder = previousDelegate == _account;
+        if (isTranscoder) {
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, current.delegatedAmount);
+        } else if (wasTranscoder) {
+            // if the account stopped being a transcoder, we want to emit an event zeroing its "delegate votes"
+            emit DelegateVotesChanged(_account, previous.delegatedAmount, 0);
+        }
```
victorges: Improving this part a bit to only send these events when the `delegatedAmount` actually changed.

contracts/bonding/BondingVotes.sol
victorges: I've changed this to `_round != clock()` and renamed the error to `InvalidTotalStakeCheckpointRound`.

The checkpointing rounds are actually different between individual accounts and total active stake tho. While individual accounts are always checkpointed on the next round, from the way that active stake update works on BondingManager, the total active stake is checkpointed for the current round instead, when the round is initialized and `nextRoundTotalActiveStake` becomes the current.

We can see it as having an optimization for the total active stake: since we have that single point where we update 1 variable for the total stake, we can leave the checkpointing for that time only. That is cheaper than checkpointing the `nextRoundTotalActiveStake` for the next round every time it changes, which would be the analogous flow to the individual accounts checkpoints.

contracts/bonding/BondingVotes.sol
victorges: We actually might in the following cases:
 - The current round has not been initialized yet, so instead of reverting or returning an inaccurate value until the round-initializer gets to it, we can just return `nextRoundTotalActiveStake`
 - This is being called for the next round (e.g. when querying for the current `totalSupply()`), in which case the checkpoint will never be present since we only checkpoint it on the round initialization
 - A couple rounds have passed without having been initialized. All those rounds should use the `nextRoundTotalActiveStake` value as well to be internally consistent with the values we return for `getBondingCheckpointAt`. 
 
 This is related to that different behavior between individual & total checkpoints, explained in the other comment https://github.com/livepeer/protocol/pull/614#discussion_r1303341851
If we checkpointed the total active stake the "expensive way", every time `nextRound...` is changed, we could have the same logic for querying here: grab the lower bound checkpoint.
Since we checkpoint only on the next round instead, it means that the "uncheckpointed total stake value" stays in the `BondingManager` contract until another round is initialized. Which is why here we:
- Use "upper bound" instead of "lower bound" to find the stake of a previous round
- Use `nextRoundTotalActiveStake` if there are no rounds initialized after the searched round

contracts/bonding/BondingVotes.sol
```diff
@@ -0,0 +1,558 @@
...
+     * active stake in the current round, which is the snapshotted stake from the end of the previous round.
+     */
+    function getPastVotes(address _account, uint256 _round) external view onlyPastRounds(_round) returns (uint256) {
+        (uint256 amount, ) = getBondingStateAt(_account, _round + 1);
+        return amount;
+    }
+
+    /**
+     * @notice Returns the current total supply of votes available.
+     * @dev This value is the sum of all *active* stake, which is not necessarily the sum of all voting power.
```
yondonfu, line 174: After thinking on this a bit, this feels weird.

I would expect that totalSupply() == sum(voting power). While not having this invariant doesn't seem like it would break anything, not having this invariant in place feels like smell...if I query totalSupply() and then index all votes for a proposal I would expect to be able to calculate the % participation in a vote as sum(voting power indexed) / totalSupply(), but that would not be wrong because the denominator is incorrect.

I know the current offchain poll tallying logic has the same problem, but if the lift to correct this problem is small (keeping track of an additional state variable in checkpoints to determine if an address was either an active transcoder or delegated to an active transcoder as of the timepoint?) to I would lean towards fixing it to start off on solid footing with proper invariants in place for onchain voting.

contracts/test/mocks/BondingVotesERC5805Harness.sol
yondonfu: Ah I see. 

Hmm it seems that the bulk of the core logic in BondingVotes right now *is* the checkpointing logic and then all the ERC-5805 functions are either fairly simple or rely on calling functions that implement the checkpointing logic. While the module approach might make testing a bit easier, it does feel like it would end up breaking up into a module (i.e. BondingVotesERC5805) that doesn't do much that just relies on abstract functions that need to be implemented in the child contract. I feel ok about the current testing approach..., but I would just document the rationale in the tests themselves that use the harness contract.

yondonfu (approved):
LGTM

contracts/bonding/BondingVotes.sol
yondonfu: Discussed offline addressing this separately from this PR

victorges: FTR we discussed offline and concluded the current behavior is reasonable given the current audit contest timelines. Might followup with another PR changing this behavior if there's time, but focusing instead on documenting the current behavior and contest for now.

edit: always surprised how github doesnt autorefresh the page

PR #571, opened 2022-05-13
removed safeMath

kautukkundan:
**What does this pull request do? Explain your changes. (required)**

Since we have updated the contracts to 0.8.9 which includes safe arithmetics, OZ's SafeMath library is no longer required.

**Specific updates (required)**

- removed imports for SafeMath
- updated contracts to use inbuilt arithmetic functions

**How did you test each of these updates (required)**

**Does this pull request close any open issues?**

closes #559  

**Checklist:**

- [ ] README and other documentation updated
- [x] All tests using `yarn test` pass

PR #615, opened 2023-07-17
treasury: Delta Governor v1

victorges:
This is the implementation for the LivepeerGovernor specified in [LIP-89](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-89.md).
It depends on the `BondingVotes` contract implementation from #614.

The higher level output of this PR is a `LivepeerGovernor` contract compatible with
the OZ Governor framework, whose participants bonding power is derived from their
staked LPT in the same way as [LIP-16](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-16.md)).

- Create governance contracts described in [LIP-89](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-89.md) apart from `BondingVotes`.
- Unit tests for `GovernorCountingOverridable`
- Integration tests to `LivepeerGovernor` incl deployment scripts

- `yarn test` all the old and new tests

Implements PRO-31

**Checklist:**


contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,222 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/governance/extensions/GovernorVotesQuorumFractionUpgradeable.sol";
```
victorges: This can be just `GovernorUpgradeable.sol` instead. Reminder to fix

contracts/treasury/LivepeerGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+     * - "Treasury" (TimelockControllerUpgradeable)
+     * - "BondingCheckpointsVotes"
+     * - "PollCreator"
+     */
+    function initialize() public initializer {
+        __Governor_init("LivepeerGovernor");
+        __GovernorSettings_init(
+            1, /* 1 round/day voting delay */
+            10, /* 10 rounds/days voting period */
+            100e18 /* 100 LPT min proposal threshold */
```
victorges: Might want to make these `initialize` args instead for configurability per network on deploys?

contracts/treasury/BondingCheckpointsVotes.sol
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * @notice Machine-readable description of the clock as specified in EIP-6372.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function CLOCK_MODE() public view returns (string memory) {
+        return bondingCheckpoints().CLOCK_MODE();
+    }
+
+    /**
+     * @notice Returns the current amount of votes that `account` has.
```
0xcadams:
```suggestion
     * @notice Returns the current amount of votes that `_account` has.
```

contracts/treasury/BondingCheckpointsVotes.sol
```diff
@@ -0,0 +1,106 @@
...
+     */
+    function getVotes(address _account) external view returns (uint256) {
+        return getPastVotes(_account, clock());
+    }
+
+    /**
+     * @notice Returns the amount of votes that `account` had at a specific moment in the past. If the `clock()` is
```
0xcadams:
```suggestion
     * @notice Returns the amount of votes that `_account` had at a specific moment in the past. If the `clock()` is
```

contracts/treasury/BondingCheckpointsVotes.sol
```diff
@@ -0,0 +1,106 @@
...
+     * Bonded stake that is not part of the top 100 active transcoder set is still given voting power, but is not
+     * considered here.
+     */
+    function getPastTotalSupply(uint256 _round) external view returns (uint256) {
+        return bondingCheckpoints().getTotalActiveStakeAt(_round);
+    }
+
+    /**
+     * @notice Returns the delegate that _account has chosen. This means the delegated transcoder address in case of
+     * delegators, and the account own address for transcoders (self-delegated).
```
0xcadams:
```suggestion
     * delegators, and the account's own address for transcoders (self-delegated).
```

contracts/treasury/BondingCheckpointsVotes.sol
```diff
@@ -0,0 +1,106 @@
...
+     */
+    function delegates(address _account) external view returns (address) {
```
0xcadams: Should this be singular?

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,222 @@
...
+    function delegatedAt(address account, uint256 timepoint) external returns (address);
+}
+
+/**
+ * @title GovernorCountingOverridable
+ * @notice Implements the Counting module from OpenZeppelin Governor with support for delegators overriding their
+ * delegated transcoder's vote. This module is used through inheritance by the Governor contract.
+ */
+abstract contract GovernorCountingOverridable is Initializable, GovernorUpgradeable {
+    using SafeMath for uint256;
```
0xcadams: Is this still needed with the latest solidity version?

contracts/treasury/LivepeerGovernor.sol
0xcadams: Yeah, I think this is a good move

contracts/treasury/LivepeerGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+    GovernorCountingOverridable
+{
+    /**
+     * @notice TreasuryGovernor constructor. Only invokes constructor of base Manager contract with provided Controller.
+     * @dev This constructor will not initialize any state variables besides `controller`. The `initialize` function
+     * must be called through the proxy after construction to initialize the contract's state in the proxy contract.
+     * @param _controller Address of Controller that this contract will be registered with
+     */
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor(address _controller) Manager(_controller) {
```
0xcadams, line 43: I am a little confused by this, and calling it out as a code smell to have an upgradeable contract with a constructor.

contracts/treasury/LivepeerGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * @title LivepeerGovernor
+ * @notice Core contract for Livepeer governance, starting as the treasury governor.
+ * @dev If we ever add fields to this class or more extensions, make sure to add a storage gap to our custom
+ * GovernorCountingOverridable extension.
+ */
+contract LivepeerGovernor is
+    Initializable,
+    ManagerProxyTarget,
```
0xcadams, line 27: I think a problem we will run into with this contract is that the upgradeable have all of the contract storage magic [1] to ensure that storage variables are easily upgradeable and don't conflict. We will need something like this in the `ManagerProxyTarget` as well.

[1] https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/f58645db634073c2c9be827c6959ace2f3733804/contracts/governance/GovernorUpgradeable.sol#L785

0xcadams (commented):
Preliminary review of code quality & gas improvements, no deep checks of logic

contracts/treasury/BondingCheckpointsVotes.sol
victorges: I don't really like it either but this actually comes from the OZ interface that we have to implement: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/f58645db634073c2c9be827c6959ace2f3733804/contracts/governance/utils/IVotesUpgradeable.sol#L53 (this is inherited by `IERC5805Upgradeable`). `delegate` is already used for the function that actually changes the delegate, as a verb.

If you think it improves readability I can add the `override` modifier (`virtual` as well?) to all the functions here that implement those interface methods. Solidity apparently doesn't require it anymore tho it might make things clearer here.

The only new method we've added here apart from the interface is `delegatedAt`, as defined in the `IVotes` interface on `GovernorCountingOverridable.sol`.

contracts/treasury/GovernorCountingOverridable.sol
victorges: I don't think so, even for the version we're using since it was fixed in 0.8.0. I blindly followed the existing code patterns for consistency, but can remove its usage in the new contracts code at least if we prefer that.

contracts/treasury/LivepeerGovernor.sol
victorges: So this is an existing code pattern in our protocol, since I assume OZ's proxy didn't even exist back when we wrote these. For general context I think this is the best documentation:
https://github.com/livepeer/protocol/blob/77a37e58192a6017a200920f58551d2cfc2873a3/contracts/ManagerProxy.sol#L7
(can go up on `ManagerProxyTarget` as well)

The core differences from OZ proxy, from my understanding, are:
1. It doesn't have a fixed target, but it fetches it from the `Controller` instead which holds an association of "contract id -> address/other info"
2. Thus it also contains some storage on its own, to hold: the `controllerAddress` and the `targetContractId`

The TLDR is that targets then need to also inherit from the `ManagerProxyTarget` storage layout otherwise they would mismatch. The constructor comes as an obligation of inheriting from `Manager`, but in practice the storage variables initialized there are never really used on the implementation contract.

We could use the pure OZ transparent proxy for this, but it would be the only contract in our protocol to use a different proxying mechanism, which would also need custom tooling only for it. So I think there are some edges to clear (as detailed in the other comment), but I think it's worth trying to stick to our existing proxy/upgradeable pattern for this.

contracts/treasury/LivepeerGovernor.sol
victorges: Thanks for pointing these out, I think it is definitely something that needs some close attention. I've tried to explain how our existing proxy system works [on the other comment](https://github.com/livepeer/protocol/pull/615#discussion_r1275360487), but I did notice one potential problem that may be happening here:

The proxy and the target contracts need a compatible storage layout. The proxy only defines 2 variables, the `controller` and the `targetContractId`. This means that the target must have exactly those 2 variables as their first fields as well, otherwise we'd be mixing them up. 

The problem that I think might exist here is that I put the `Initializable` parent before the `ManagerProxyTarget`, meaning that the variables that should match each other are not aligned. That's an easy fix, making sure the `ManagerProxyTarget` is the first inherited contract, I only don't know how this hasn't exploded in the most basic tests (like corrupting the controller address on `initialize()`). I'll try to poke this around to make sure we have some tests covering this.

victorges: Regarding storage gaps: I don't think we need it in the `ManaferProxyTarget` as those contracts are pretty much already tied to never change: all our existing contracts like `BondingManager`, `RoundsManager` etc already inherit from them. If we ever need to make changes to it we'll need to make a `V2` and only make new contracts use it instead.

I did wonder if we wanted to add a storage gap to `GovernorCountingOverridable` though, since if we add other contracts after it or any custom fields to `LivepeerGovernor`, we would be block from upgrading. We don't need it right now, but could be a safety measure for future upgrades so we don't need to remember it on a future code review. WDYT?

victorges: OK I fixed the 2 storage layout issues on https://github.com/livepeer/protocol/pull/615/commits/0ad3a8fca7b121e82d67034cd92585967dc782a9

Specifically:
- Fix the inheritance order so that `ManagerProxyTarget` comes first, to guarantee storage layout compatibility with the proxy
- Add a storage `__gap` to `GovernorCountingOverridable` for safety in future upgrades

I also understood why nothing was exploding: due to C3-linearization of the parent contracts, the order in which the fields were laid out in storage was:
```
┌────────────────────────────────────────┬──────────────────────────────────┬──────────────┬────────┬
│                contract                │          state_variable          │ storage_slot │ offset │                         >
├────────────────────────────────────────┼──────────────────────────────────┼──────────────┼────────┼
│            LivepeerGovernor            │            controller            │      0       │   0    │                         >
│            LivepeerGovernor            │           _initialized           │      0       │   20   │                         >
│            LivepeerGovernor            │          _initializing           │      0       │   21   │                         >
│            LivepeerGovernor            │         targetContractId         │      1       │   0    │
```

Meaning a linearization as `Manager`, `Initializable`, `ManagerProxyTarget`, ...
The lucky strike on that order is that the fields from the `Initializable` parent (`_initialized` and `_initializing`) ended up using  a storage space that is comlpetely empty on the `ManagerProxy`, since they used the empty space after the `controller` field. So there was no conflict there, but out of luck and also not guaranteed by our code, so rather flimsy.

After re-ordering, the `Initializable` fields come only after the `targetContractId` from the `ManagerProxyTarget`, meaning 1 additional storage slot on the whole contract, which I think is worth it due to clearer storage layout guarantees.

Finally, I've also included a couple tests on that commit to exercise upgradability a bit, which was missing in the test package. Some lower level tests could probably be done using foundry, but I believe those are enough. Let me know if y'all think we need any more guarantees there apart from the tests we're gonna do in testnet+forked mainnet as well.

contracts/treasury/LivepeerGovernor.sol
victorges: Resolving as per https://github.com/livepeer/protocol/pull/615#discussion_r1276844201

contracts/treasury/BondingCheckpointsVotes.sol
0xcadams: I don't think it actually matters - just nitpicking! I think it makes sense to keep as-is if the parent uses that

contracts/treasury/LivepeerGovernor.sol
0xcadams: Yeah, I agree that sticking to our existing proxy makes sense. It looks like you resolved the contract storage issues, which is really what I am concerned about. And, since this is only calling `_disableInitializers()`, there's no issue here.

contracts/treasury/GovernorCountingOverridable.sol
0xcadams: Yeah I'd defer to @yondonfu on this - I know we also were planning on moving to a post-SafeMath world previously, but to my knowledge these changes were never deployed to Arbitrum mainnet.

https://github.com/livepeer/protocol/pull/571

contracts/treasury/LivepeerGovernor.sol
0xcadams: Ha, interesting, yeah I see. So we stumbled into storage compatibility. Yeah, I think the extra storage is worth it for the upgradeability guarantees. Thanks for getting those tests in place, that's good to increase confidence.

Definitely something we should also have more experienced auditor(s) look at to make sure we're future-proof.

contracts/treasury/GovernorCountingOverridable.sol
yondonfu: I think we should drop SafeMath usage for new contracts since they (along with the rest of the contracts in the repo) are compiled with solc v0.8.9 and the compiler has built in overflow checking in >= v0.8.0.

I think we should leave SafeMath usage as-is in existing contracts that are being updated in Delta PRs for now to minimize the scope of changes in existing contracts even if SafeMath is not needed in those contracts.

contracts/treasury/LivepeerGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+        __Governor_init("LivepeerGovernor");
+        __GovernorSettings_init(initialVotingDelay, initialVotingPeriod, initialProposalThreshold);
+        __GovernorTimelockControl_init(treasury());
+
+        // The GovernorVotes module will hold a fixed reference to the votes contract. If we ever change its address we
+        // need to call the {bumpVotesAddress} function to update it in here as well.
+        __GovernorVotes_init(votes());
+
+        // Initialize with the same value from the existing polling system.
+        uint256 initialQuorum = pollCreator().QUORUM();
```
yondonfu: IMO simpler to not sync with the PollCreator here and to just have independent quorum and quota state variables tracked by LivepeerGovernor giving flexibility for the parameters of the two systems to be independent one another even if they initially have the same values.

contracts/treasury/LivepeerGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+     */
+    function votes() public view override returns (IVotes) {
+        return bondingCheckpointVotes();
+    }
+
+    /**
+     * @dev See {GovernorCountingOverridable-quota}. We use the same QUOTA value from the protocol governance system for
+     * now, but can consider changing this in the future (e.g. to make it updateable through proposals without deploys).
+     */
+    function quota() public view override returns (uint256) {
```
yondonfu: See previous comment re: independent state variables for LivepeerGovernor for quorum and quota.

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,229 @@
...
+     * @return quota value as a MathUtils percentage value (e.g. 6 decimal places).
+     */
+    function quota() public view virtual returns (uint256);
+
+    /**
+     * @dev This empty reserved space is put in place to allow future versions to add new
+     * variables without shifting down storage in the inheritance chain.
+     * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
+     */
+    uint256[50] private __gap;
```
yondonfu: Any reason for 50 vs 49 as is used here? https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/3d4c0d5741b131c231e558d7a6213392ab3672a5/contracts/governance/extensions/GovernorCountingSimpleUpgradeable.sol#L112

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    mapping(uint256 => ProposalTally) private _proposalTallies;
+
+    /**
+     * @dev See {IGovernor-COUNTING_MODE}.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function COUNTING_MODE() public pure virtual override returns (string memory) {
+        return "support=bravo&quorum=for,abstain,against";
```
yondonfu, line 77: Why does this include against, but [GovernorCountingSimpleUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/3d4c0d5741b131c231e558d7a6213392ab3672a5/contracts/governance/extensions/GovernorCountingSimpleUpgradeable.sol#L43) does not?

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,229 @@
...
+        voter.support = support;
+
+        _weight = _handleVoteOverrides(_proposalId, tally, voter, _account, _weight);
+
+        if (support == VoteType.Against) {
+            tally.againstVotes += _weight;
+        } else if (support == VoteType.For) {
+            tally.forVotes += _weight;
+        } else {
+            assert(support == VoteType.Abstain);
```
yondonfu: Don't think this is needed if there is a check on L135.

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,229 @@
...
+        return totalVotes >= quorum(proposalSnapshot(_proposalId));
+    }
+
+    /**
+     * @dev See {Governor-_voteSucceeded}. In this module, the forVotes must be at least QUOTA of the total votes.
+     */
+    function _voteSucceeded(uint256 _proposalId) internal view virtual override returns (bool) {
+        (uint256 againstVotes, uint256 forVotes, ) = proposalVotes(_proposalId);
+
+        // we ignore abstain votes for vote succeeded calculation
```
yondonfu, line 121: Why? If Abstain is a valid vote option shouldn't the total valid votes include abstain votes?

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -0,0 +1,229 @@
...
+
+import "../bonding/libraries/EarningsPool.sol";
+import "../bonding/libraries/EarningsPoolLIP36.sol";
+
+import "../Manager.sol";
+import "../IController.sol";
+import "../rounds/IRoundsManager.sol";
+import "../bonding/IBondingCheckpoints.sol";
+
+interface IVotes is IERC5805Upgradeable {
```
yondonfu: Move this to own file so that BondingCheckpointsVotes doesn't need to ref this file? Was a bit confusing to see a ref to this file in BondingCheckpointsVotes despite BondingCheckpointsVotes not directly depending on GovernorCountingOverridable.

yondonfu (changes requested):

contracts/treasury/GovernorCountingOverridable.sol
victorges: Moved on https://github.com/livepeer/protocol/pull/614/commits/d518cd6058e87c7bc10c8ecfc05696e8c30655e8 in the previous PR, when merging `BondingCheckpoints` with `...Votes`

contracts/treasury/LivepeerGovernor.sol
victorges: Cool, I think it makes a lot of sense as well and I had to do additional stuff to make this work (e.g. registering the poll creator on controller) which shouldn't be necessary. 

I'll change these to be first-party parameters from the governor, with the caveat that I think the `quota()` should not be updateable for now. That because we either have to implement a full checkpointing logic [1] for the quota value or we would be subject to a similar vulnerability like this [2]. So in order to simplify this first version, I'll just make it a constant or a parameter on `initialize` for now.

[1] The simplest implementation being something [like this](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/0de6a5343637ee02455119b5c137fa33ec952705/contracts/governance/extensions/GovernorVotesQuorumFractionUpgradeable.sol) which is for the quorum. It's practically >100 additional LoC + tests and spec etc, even using the OZ checkpoints lib underneath.
[2] https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories/GHSA-xrc4-737v-9q75

contracts/treasury/GovernorCountingOverridable.sol
victorges: In the lack of further documentation about standards for this apart from ["mind the gap"](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#storage-gaps) [1], I looked around on OZ contracts and my first understanding was that they always started these gaps as `50` and then reduced it as new state variables were introduced. After your comment I took another look and realized that the idea (at least around OZ) is to always keep the contracts with a storage layout of 50 slots, reducing the gap depending on the amount of vars it has even on the first version of the contract.

So I'm changing this to `49` to have the whole contract have 50 slots size. Actually 48, since I'm adding a new variable for the `quota` now.

[1] I had to make that joke, but just before replying here I found [these docs](https://docs.openzeppelin.com/contracts/3.x/upgradeable#storage_gaps) which does mention the "50 slots" standard. It's still just an OZ convention anyway, but no harm in following it and better to keep the consistency.

contracts/treasury/GovernorCountingOverridable.sol
victorges: 2 main reasons:
- Our [current poll tallying logic](https://github.com/livepeer/explorer/blob/c6c7cb1b8cc5fab66c575dd63f03a53aab87c2b1/lib/api/polls.ts#L76) considers both "yes" and "no" votes as participation. So I guess we'd also need to rethink our quorum/quota values if we were to change this.
- My personal opinion, as from every source I searched for the term "quorum" in the abstract sense, it means "the amount of people in the room". IMHO anyone voting is in the room and so should be considered for quroum. I found the `GovernorCountingSimple` logic a bit counterintuitive (and even more the bravo impl which only considers `for` votes!)

But if it wasn't for our current voting system I'd pick the consistency with OZ as well, even if it seems weird on a first glance. After all our quorum logic will be different from most Governors out there if we implement that differently here.

Let me know if you'd prefer to keep compatibility with OZ instead of our current polls, or anything else. Don't have a strong opinion about this anyway.

contracts/treasury/GovernorCountingOverridable.sol
victorges: I think this is mostly a naming issue on the "valid votes" variable. The logic here is that we are fetching the amount of votes with an opinion about the proposal outcome (which `abstain` doesn't [1]), and then check if we have more `for` votes than the quota of that total. I have renamed the variable to `opinionatedVotes` which I think makes more sense. This is basically the same implementation as `GovernorCountingSimple`, the only difference being that we allow the quota to be configurable instead of forcing 50% (they just do `for > against`)

Lmk if you have any other suggestion!

[1] IMO "abstain" means that you're casting a vote to communicate that you are ok with any outcome of the proposal, so you don't really have an opinion about it. That is my reasoning to think it makes sense to consider it for `_quorumReached`, but not for `_voteSucceeded`. If we did include them here there wouldn't be a behavioral difference between `abstain` and `against` (unless we removed `against` from quorum count I guess).

victorges:
@yondonfu Also ready for another round! 😁 

Notice that I made a "merge" with the parent branch just to avoid rebasing and losing all history. Let me know if you'd prefer a clean rebase. When merging, my plan is to either do the full linearization of the branches or squash&merge them instead not to keep the non-linear git history.

contracts/treasury/LivepeerGovernor.sol
dob: > I think the quota() should not be updateable for now

Sounds ok, but just so I understand...is LivepeerGovernor upgradeable? Meaning if we wanted to update the quota we have the ability to do so via updating the LivepeerGovernor contract itself? And if so, is the mechanism that it is owned by itself (IE upgradable via a voting action), or owned by the security committee multisig?

victorges: Yeah, it is upgradeable! Just note that if we do upgrade it for the quota we'll need to upgrade it to have the full checkpoint history instead of just changing the current value, or we would be subject to the same vulnerability linked above for the updatable quorum.

As for how the upgrades themselves work, the governor is upgradeable through the same mechanism as our other contacts in the protocol today, through a Target contract registered in the controller which can only be changed by the controller owner. When we do make the change for the governor to own the protocol then the governor would be able to upgrade itself (as well as the rest of the protocol), but for now it's subject to our multisig.

@dob Does that make sense/sound good?

dob: Yes, that makes sense to me and sounds good.

contracts/treasury/GovernorCountingOverridable.sol
yondonfu: Including all these options for quorum seems right. I'm guessing that specifying that here allows an offchain tool to determine which vote options are part of quorum?

victorges: Yeah exactly! So the `support=bravo` indicates our association of vote type as a `uint8` to what they represent for the vote (0 against, 1 for, 2 abstain), and the second `quorum` param indicates what we use internally for the quorum. Both of which UIs like Tally _should_ (😆) use to display vote counting info for a proposal.

yondonfu (approved):
LGTM

PR #616, opened 2023-07-20
 bonding: Treasury rewards contribution

victorges:
This implements the behavior described in [LIP-92](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-92.md), for a logic in `BondingManager`
to send a percentage of rewards to a treasury contract. This treasury contract
was implemented in #615.

It also supports a balance ceiling which, if reached, will automatically halt treasury contributions until
either the ceiling or the cut rate parameters (or both) are updated again.

Changing the treasuryCutRate also takes place only on the next round, to safeguard againsts
possible games about when to execute a transaction in a round.

- Add updateable parameter on BondingManager
- Add logic on `rewardWithHint` and `updateTranscoderWithFees` to consider treasury contribution
- Increment unit tests for the new logic

- `yarn test`

Implements PRO-27

**Checklist:**


contracts/bonding/BondingManager.sol
```diff
@@ -307,6 +336,11 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
                 totalStake,
                 currentRoundTotalActiveStake
             );
+
+            // Deduct what would have been the treasury rewards
+            uint256 treasuryRewards = MathUtils.percOf(rewards, treasuryRewardCutRate);
+            rewards = rewards.sub(treasuryRewards);
```
victorges, line 342: Might need some special attention here, as I had trouble understanding what is really the purpose of the whole logic in this `if` branch.

I understand _when_ it is called, but it seems like it is only capable of making an approximation of what the previous round rewards would have been, but it's not an exact calculation. 

On line 335 it also uses `mintableTokens+mintedTokens` as the estimate for the "total previous round rewards", when from my understanding it should be only `mintableTokens` assuming the previous round had similar mintable tokens as the current round. Is there any similar LIP or post explaining the mathematics done here?

So there's also a theoretical problem here as well, which is that we always use the current cut rate to calculate what the previous round treasury cut would have been. This can be theoretically a problem if the cut rate has changed since the last round, but it seems like this is too much of a corner case + already an approximation for us to need any code complexity to make this more exact.

WDYT?

victorges:
Btw, not sure what is wrong with the tests. If I run the unit tests by themselves they all work, but they are failing under coverage for some reason. I didn't make changes to the tests that are failing (`RoundsManager`). Any idea of what could be causing this?

Edit: Found out a flaky test that wasn't `awaiting` and fixed it later on

contracts/bonding/BondingManager.sol
```diff
@@ -839,18 +873,37 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             earningsPool.setStake(t.earningsPoolPerRound[lastUpdateRound].totalStake);
         }
 
+        if (treasuryBalanceCeiling > 0) {
+            uint256 treasuryBalance = livepeerToken().balanceOf(treasury());
+            if (treasuryBalance >= treasuryBalanceCeiling) {
+                // halt treasury contributions until the cut rate param is updated again
+                treasuryRewardCutRate = 0;
```
yondonfu: Instead of explicitly setting treasuryRewardCutRate to 0 here what if we just skip the treasury transfer if if we've hit the ceiling? Then, if the balance of the treasury falls below the ceiling the contributions automatically continue instead of requiring an additional param update to change treasuryRewardCutRate from 0.

Could update the conditional on L889 to something like:

```
if (treasuryRewards > 0 && treasuryBalance < treasuryBalanceCeiling) { .. }
```

contracts/bonding/BondingManager.sol
```diff
@@ -1553,6 +1606,10 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         return IRoundsManager(controller.getContract(keccak256("RoundsManager")));
     }
 
+    function treasury() internal view returns (address payable) {
+        return payable(controller.getContract(keccak256("Treasury")));
```
yondonfu: Does the return value need to be payable? Doesn't seem that the treasury address ever receives ETH.

contracts/bonding/BondingManager.sol
yondonfu: A couple resources:

- [LIP-36 section on fees - search for the phrase "retroactively calculate"](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-36.md#fees)
- [Original comment discussing the retroactive calculation](https://github.com/livepeer/LIPs/issues/35#issuecomment-673659199)
- [Cumulative Reward and Fee Factors for Delegators section of LIP-36 discussion thread](https://forum.livepeer.org/t/a-more-gas-efficient-earnings-calculation-approach/1097)

IIRC the logic here is:

- `cumulativeRewardFactor(currentRound) = cumulativeRewardFactor(currentRound - 1) * (1 + delegatorsRewards(currentRound) / transcoderActiveStake(currentRound)`
- Suppose we know `cumulativeRewardFactor(currentRound)`, but we do not know `cumulativeRewardFactor(currentRound - 1)`
- We can rearrange the first equation to get: `cumulativeRewardFactor(currentRound - 1) = cumulativeRewardFactor(currentRound) / (1 + delegatorsRewards(currentRound) / transcoderActiveStake(currentRound)`

> On line 335 it also uses mintableTokens+mintedTokens as the estimate for the "total previous round rewards"

This is supposed to be the total mintable tokens for the current round which is used to derive the rewards that the transcoder minted in the current reward in a reward call via the equation: `rewards = total mintable tokens * (transcoderActiveStake(currentRound) / totalActiveStake(currentRound))`. Then, `delegatorsRewards(currentRound)` can be derived from `rewards` by subtracting the transcoder's reward cut.

> So there's also a theoretical problem here as well, which is that we always use the current cut rate to calculate what the previous round treasury cut would have been.

Since we're trying to derive `delegatorsRewards(currentRound)` as long as `treasuryRewardCutRate` does not change in between the transcoder's reward call in the current round and the time at which this logic is executed we avoid a problem...

A few options to avoid that scenario:

1. Protocol governance should only execute an update that changes `treasuryRewardCutRate` towards the end of a round so that by the time reward calls in the next round occur the `treasuryRewardCutRate` is locked in and won't change after that
2. Store additional state to track `treasuryRewardCutRate` for each round so that even if there is an update can always use the stored value for the round
3. Store an extra `nextTreasuryRewardCutRate`. `setTreasuryRewardCutRate()` updates `nextTreasuryRewardCutRate` and then `setCurrentRoundTotalActiveStake()` also sets `treasuryRewardCutRate = nextTreasuryRewardCutRate` which means that when a round is initialized (and `setCurrentRoundTotalActiveStake()` is called) the `treasuryRewardCutRate` is locked in for the round

Slight pref for 3 as it seems to be the cleanest approach. WDYT?

yondonfu:
> Btw, not sure what is wrong with the tests.

Not immediately sure, but looks like coverage tests actually started failing in https://github.com/livepeer/protocol/pull/615.

`yarn test:coverage` does result in failures locally as well so its not just a CI thing. I tried upgrading to latest hardhat & solidity-coverage locally, but still seeing failures with a couple different ones as well:

```
1) Rewards
       correctly calculates reward shares for delegators and transcoders:
     Error: Timeout of 100000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/yondonfu/Development/livepeer/protocol/test/integration/Rewards.js)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)

  2) Minter
       migrateToNewMinter
         should transfer current token balance and current ETH balance to new minter:
     TypeError: vm.on is not a function
      at Function.fromEvent (node_modules/@defi-wonderland/smock/src/observable-vm.ts:42:8)
      at new ObservableVM (node_modules/@defi-wonderland/smock/src/observable-vm.ts:19:35)
      at new Sandbox (node_modules/@defi-wonderland/smock/src/sandbox.ts:36:15)
      at Function.create (node_modules/@defi-wonderland/smock/src/sandbox.ts:82:12)
      at async init (node_modules/@defi-wonderland/smock/src/index.ts:27:13)
      at async Object.fake (node_modules/@defi-wonderland/smock/src/index.ts:13:17)

  3) Minter
       trustedBurnTokens
         should burn tokens:
     TypeError: vm.on is not a function
      at Function.fromEvent (node_modules/@defi-wonderland/smock/src/observable-vm.ts:42:8)
      at new ObservableVM (node_modules/@defi-wonderland/smock/src/observable-vm.ts:19:35)
      at new Sandbox (node_modules/@defi-wonderland/smock/src/sandbox.ts:36:15)
      at Function.create (node_modules/@defi-wonderland/smock/src/sandbox.ts:82:12)
      at async init (node_modules/@defi-wonderland/smock/src/index.ts:27:13)
      at async Object.fake (node_modules/@defi-wonderland/smock/src/index.ts:13:17)

  4) PollCreator
       "before all" hook in "PollCreator":
     TypeError: vm.on is not a function
      at Function.fromEvent (node_modules/@defi-wonderland/smock/src/observable-vm.ts:42:8)
      at new ObservableVM (node_modules/@defi-wonderland/smock/src/observable-vm.ts:19:35)
      at new Sandbox (node_modules/@defi-wonderland/smock/src/sandbox.ts:36:15)
      at Function.create (node_modules/@defi-wonderland/smock/src/sandbox.ts:82:12)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async init (node_modules/@defi-wonderland/smock/src/index.ts:27:13)
      at async Object.fake (node_modules/@defi-wonderland/smock/src/index.ts:13:17)

  5) RoundsManager
       setRoundLength
         should set roundLength before lastRoundLengthUpdateRound and lastRoundLengthUpdateStartBlock when roundLength = 0:
     AssertionError: wrong lastRoundLengthUpdateRound: expected { Object (_hex, _isBigNumber) } to equal 8
      at /Users/yondonfu/Development/livepeer/protocol/test/unit/RoundsManager.js:94:13
      at step (test/unit/RoundsManager.js:52:23)
      at Object.next (test/unit/RoundsManager.js:33:53)
      at fulfilled (test/unit/RoundsManager.js:24:58)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at runNextTicks (node:internal/process/task_queues:64:3)
      at listOnTimeout (node:internal/timers:533:9)
      at processTimers (node:internal/timers:507:7)
```

contracts/bonding/BondingManager.sol
dob: @yondonfu this may be a philosophical discussion better suited for the forum, but I think the main reason we set it to zero was in the spirit of the "ice age" - where it forced a community to come together to take governance action at an important moment needing review. Rather than the ceiling being some important balance limit that needs to be maintained. IE if the ceiling gets hit in the first place, it means the treasury isn't successfully allocating funds, and should be reviewed, or at least funding put on pause until it can justify that it needs more.

contracts/bonding/BondingManager.sol
dob: Might be beyond my technical depth on what the payable function means here, but in spirit the treasury can receive any asset, including ETH. Even if the Livepeer protocol only will send it LPT.

contracts/bonding/BondingManager.sol
yondonfu: Got it. Thanks for the clarification. With that in mind, I think the current logic is fine.

contracts/bonding/BondingManager.sol
yondonfu: The payable keyword here doesn't impact the Treasury contract's ability to receive ETH and is just used to inform the compiler and the programmer that the address will need be able to receive ETH from code implemented in this contract (BondingManager). So, if the BondingManager never needs to send ETH to the Treasury technically speaking the payable keyword isn't necessary.

victorges: I think I ended up adding the `payable` when I was converting this to the explicit type `Treasury/TimelockController`, which has payable functions and thus the compiler complains when you try to convert a non-payable address to it. Then I stopped referencing the specific contract implementation here and just returned an address and didn't realize that I didn't need the payable conversion anymore. Will fix!

victorges: This was actually a leftover from a previous version where I returned the concrete instance of the contract. Since the `Treasury` (`TimelockController`) has a fallback payable function, solidity requires the address to be converted to `payable` first.

As we're only returning the address now this is not necessary anymore, so I've removed the `payable` conversion. Thanks for noticing this.

contracts/bonding/BondingManager.sol
victorges: Thanks for the references! I believe I understand the overall need for this logic. In practice, it does not make an estimate but calculates exactly the value of `cumulativeRewardFactor` for the previous reward round. This _should_ result in exactly the value on the earning pool on the previous reward call. 

In a sense, this could've been solved with the checkpointing system that we are building for this. IOW if we had a field like `previousRewardRound` inside every earnings pool, we could fetch the value directly from that last round where reward() was run before the current one and this logic wouldn't be needed.

There is still something that looks weird to me though. Specifically:
 - [This line](https://github.com/livepeer/protocol/blob/cf12cb4734dc1e7e734faed18513f38502d61095/contracts/bonding/BondingManager.sol#L340) calculates the `rewards` value, which are the reward tokens distributed to this transcoder in the current round. It uses `currentMintableTokens + currentMintedTokens` to calculate the total round rewards and take the perc of
 - OTOH on the actual reward call, we calculate this via the [`createReward()`](https://github.com/livepeer/protocol/blob/cf12cb4734dc1e7e734faed18513f38502d61095/contracts/bonding/BondingManager.sol#L889) function on the minter which actually only uses [`currentMintableTokens`](https://github.com/livepeer/protocol/blob/d1e2222173618cb5cd905fd58bb78c72e521b20f/contracts/token/Minter.sol#L153) for the "total round rewards" and take the same perc of
 
So it looks like the first one might be calculating them wrong, inflating the value of what the transcoder rewards were in the current round. This doesn't affect the treasury reward logic tho so doesn't block this PR, but if there is indeed a problem there I could include the fix on this changeset (basically remove the `+ currentMintedTokens`).

Will send a separate msg about the `treasuryReardCutRate` itself

contracts/bonding/BondingManager.sol
```diff
@@ -841,16 +888,35 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             earningsPool.setStake(t.earningsPoolPerRound[lastUpdateRound].totalStake);
         }
 
+        if (treasuryBalanceCeiling > 0) {
+            uint256 treasuryBalance = livepeerToken().balanceOf(treasury());
+            if (treasuryBalance >= treasuryBalanceCeiling && nextRoundTreasuryRewardCutRate > 0) {
+                // halt treasury contributions until the cut rate param is updated again
+                _setTreasuryRewardCutRate(0);
```
victorges, line 875: "reposting" here cause github hid the comment in the commit.
https://github.com/livepeer/protocol/commit/ead1edb572319df53dbfa06c3466edb4ca8c0575#r124819767

contracts/bonding/BondingManager.sol
victorges: lol github was buggy for me yesterday, I could swear I had already replied here but it wasn't showing me

contracts/bonding/BondingManager.sol
```diff
@@ -150,6 +158,34 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+    /**
+     * @notice Set treasury reward cut rate. Only callable by Controller owner
+     * @param _cutRate Percentage of newly minted rewards to route to the treasury. Must be a valid PreciseMathUtils
+     * percentage (<100% specified with 27-digits precision).
+     */
+    function setTreasuryRewardCutRate(uint256 _cutRate) external onlyControllerOwner {
+        _setTreasuryRewardCutRate(_cutRate);
+    }
+
+    function _setTreasuryRewardCutRate(uint256 _cutRate) internal {
```
yondonfu: Group with other internal functions as mentioned in another PR.

contracts/bonding/BondingManager.sol
yondonfu: Will follow up offline!

contracts/bonding/BondingManager.sol
yondonfu: Left my comment in that link.

yondonfu (approved):
LGTM

PR #620, opened 2023-08-25
deploy: Delta devnet 

victorges:
This is a pull request with the changes for deploying a devnet for the Delta project. It also includes a
few improvements to our deployment/verification tooling and should be worth merging even if the 
devnet goes away soon.

- Fix .env files which weren't really being used without the `dotenv` package import
- Add network&migrations configs for the devnet on Arbitrum Goerli
- Create a task that verifies the source code of all deployed contracts
- Add some simple docs on how to deploy a test/devnet yourself
- Actually deployed and committed the deployment artifacts (only reason why this is huge in line numbers)

Ran deploys a bunch of times.

Implements PRO-36

**Checklist:**


victorges:
we are now on mainnet 😎

PR #621, opened 2023-08-25
[delta-audit] bonding: Only provide voting power for active stake

victorges:
This is to fix the `totalSupply = sum(getVotes(*))` invariant from `BondingVotes`,
which wasn't held before when we provided voting power to both active and inactive
transcoders and their delegators.

With this change, we make sure the transcoder active stake is also taken into account
and provide zero voting power when the transcoder is not active.

- Start checkpointing the `isActiveTranscoder` bool, derived from activation and deactivation rounds
- Fix tests that broke and added some new tests to the new logic (like checkpoints on active set changes)

`yarn test`

Fixes PRO-39

Also ended up being a potential fix for https://github.com/code-423n4/2023-08-livepeer-findings/issues/194

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -505,13 +505,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         t.feeShare = _feeShare;
 
         if (!transcoderPool.contains(msg.sender)) {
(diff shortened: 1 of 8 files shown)


victorges:
we're going with #625 for now. this can be reopen in the future as a separate improvement

PR #622, opened 2023-09-07
[delta-audit] bonding: Various fixes from audit findings

victorges:
This mitigates a couple minor findings from the audit that either haven't been flagged by wardens but we found through self-review (e.g. duplicate checkpoints) or too minor issue to have its own PR (solidity version). Even though I'm combining all the fixes in this PR, each mitigation is done on a separate commit to facilitate review.

- Avoid double-checkpointing of transcoders when making self-bonding actions
- Add the `previous/newLastClaimRound` to the `DelegatorBondedAmountChanged` event (otherwise some changes would be silent on the event side, like unbonding exactly the amount of pending rewards)

The minorest:
- Fix solidity version in `IBondingVotes.sol` file
- Add a `remappings.txt` file to the project, which improves VS Code suggestions and navigation
- Fix a few misleading or typod documentations

Added tests to all the functional changes and `yarn test`

Will deploy on devnet after merged.

N/A

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingManager.sol
+++ b/contracts/bonding/BondingManager.sol
@@ -441,8 +441,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
 
     /**
      * @notice Claim token pools shares for a delegator from its lastClaimRound through the end round
-     * @param _endRound Unused, represented the last round for which to claim token pools shares for a delegator.
-     * Currently, the earnings are always claimed until the current round instead.
+     * @param _endRound Unused, but used to represented the last round for which to claim token pools shares for a
+     * delegator. Currently, the earnings are always claimed until the current round instead.
      */
     function claimEarnings(uint256 _endRound)
         external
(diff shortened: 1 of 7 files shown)


contracts/bonding/BondingManager.sol
```diff
@@ -587,6 +587,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
             delegationAmount = delegationAmount.add(currentBondedAmount);
 
             decreaseTotalStake(currentDelegate, currentBondedAmount, _oldDelegateNewPosPrev, _oldDelegateNewPosNext);
+            // currentDelegate is always different than msg.sender so no need to avoid double checkpointing
```
yondonfu: Suggest changing msg.sender to _owner in this comment.

And just to clarify - this statement is only true because L582 prevents execution from reaching here if the address is a transcoder (i.e. self-delegated) correct?

contracts/bonding/BondingManager.sol
```diff
@@ -902,7 +912,8 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
     /**
      * @notice Returns pending bonded stake for a delegator from its lastClaimRound through an end round
      * @param _delegator Address of delegator
-     * @param _endRound The last round to compute pending stake from
+     * @param _endRound Unused, but represented the last round to compute pending stake from. Currently, the pending
```
yondonfu: The same param doc needs to be updated for pendingFees().

yondonfu (changes requested):
Just a few minor comments.

contracts/bonding/BondingManager.sol
victorges: > And just to clarify - this statement is only true because L582 prevents execution from reaching here if the address is a transcoder (i.e. self-delegated) correct?

Correct! I've rephrased the doc, including the `_owner` fix you suggested

yondonfu (approved):
LGTM

PR #625, opened 2023-09-19
[delta-audit] bonding: Provide votes only for registered Ts and their Ds

victorges:
This fixes another issue from the audit regarding voting power provided to delegators. The specific issue is when a delegator has a bond with an account that is not a "registered transcoder" (`delegateAddress == self && bondedAmount > 0`). 

This can happen when bonding to a random account that is not a transcoder, when the transcoder unbonds or gets slashed 100%. The main problem created from this is that the transcoder would then have `0` voting power, while the delegators would still hold on to their delegated voting power, messing up vote overriding on `GovenorCountingOverridable`. Even with the https://github.com/livepeer/protocol/pull/626 fix, there would still be a problem in case `bondedAmount == 0` but still `delegateAddress == self` (can happen in the slash case).

- Update BondingVotes to not give voting power to delegators to non-transcoders (transcoder defined as self-delegate a non 0 amount)
- Add tests

`yarn test` with pending devnet deploy

Fixes:
- https://github.com/code-423n4/2023-08-livepeer-findings/issues/194
- https://github.com/code-423n4/2023-08-livepeer-findings/issues/96

**Checklist:**

### DIFF
--- a/contracts/bonding/BondingVotes.sol
+++ b/contracts/bonding/BondingVotes.sol
@@ -107,7 +107,7 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
     constructor(address _controller) Manager(_controller) {}
 
     // IVotes interface implementation.
-    // These should not access any storage directly but proxy to the bonding state functions.
+    // These should not access any storage directly but proxy to the historical stake functions below.
 
     /**
      * @notice Returns the name of the virtual token implemented by this.
@@ -149,28 +149,21 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
     /**
      * @notice Returns the current amount of votes that `_account` has.
      *
-     * The value returned by this function can be equivalently calculated using {BondingManager} functions as follows:
-     * `isRegisteredTranscoder(_account) ? transcoderTotalStake(_account) : pendingStake(_account, 0)`
-     *
-     * @dev Keep in mind that since this function should return the votes at the end of the current round, we need to
-     * fetch the bonding state at the next round instead. That because the bonding state reflects the active stake in
-     * the current round, which is the snapshotted stake from the end of the previous round.
+     * The voting power for a delegator is the amount they are delegating to a transcoder, while for transcoders it is
+     * all the stake delegated to them. If an account is not a registered transcoder
+     * ({BondingManager-isRegisteredTranscoder}), the voting power of itself and of all its delegators will be zero.
      */
     function getVotes(address _account) external view returns (uint256) {
-        (uint256 amount, ) = getBondingStateAt(_account, clock() + 1);
-        return amount;
+        (uint256 votes, ) = getVotesAndDelegateAtRoundStart(_account, clock() + 1);
+        return votes;
     }
 
     /**
-     * @notice Returns the amount of votes that `_account` had at a specific moment in the past. If the `clock()` is
-     * configured to use block numbers, this will return the value at the end of the corresponding block.
-     * @dev Keep in mind that since this function should return the votes at the end of the _round (or timepoint in OZ
-     * terms), we need to fetch the bonding state at the next round instead. That because the bonding state reflects the
-     * active stake in the current round, which is the snapshotted stake from the end of the previous round.
+     * @notice Returns the amount of votes that `_account` had at the end of the provided past `_round`.
      */
     function getPastVotes(address _account, uint256 _round) external view onlyPastRounds(_round) returns (uint256) {
-        (uint256 amount, ) = getBondingStateAt(_account, _round + 1);
-        return amount;
+        (uint256 votes, ) = getVotesAndDelegateAtRoundStart(_account, _round + 1);
+        return votes;
     }
 
     /**
@@ -178,22 +171,16 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
      * @dev This value is the sum of all *active* stake, which is not necessarily the sum of all voting power.
      * Bonded stake that is not part of the top 100 active transcoder set is still given voting power, but is not
      * considered here.
-     * @dev Keep in mind that since this function should return the votes at the end of the current round, we need to
-     * fetch the total active stake at the next round instead. That because the active stake in the current round is the
-     * snapshotted stake from the end of the previous round.
      */
     function totalSupply() external view returns (uint256) {
         return getTotalActiveStakeAt(clock() + 1);
     }
 
     /**
-     * @notice Returns the total supply of votes available at a specific round in the past.
+     * @notice Returns the total supply of votes available at the end of the provided past `_round`.
      * @dev This value is the sum of all *active* stake, which is not necessarily the sum of all voting power.
      * Bonded stake that is not part of the top 100 active transcoder set is still given voting power, but is not
      * considered here.
-     * @dev Keep in mind that since this function should return the votes at the end of the _round (or timepoint in OZ
-     * terms), we need to fetch the total active stake at the next round instead. That because the active stake in the
-     * current round is the snapshotted stake from the end of the previous round.
      */
     function getPastTotalSupply(uint256 _round) external view onlyPastRounds(_round) returns (uint256) {
         return getTotalActiveStakeAt(_round + 1);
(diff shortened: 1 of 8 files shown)


contracts/bonding/BondingVotes.sol
```diff
@@ -461,17 +465,20 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
...
-            bond.delegateAddress,
-            _round
-        );
+        (
+            BondingCheckpoint storage transcoderBond,
+            EarningsPool.Data memory endPool
+        ) = getLastTranscoderRewardsEarningsPool(transcoder, _round);
+
+        if (!isRegisteredTranscoder(transcoder, transcoderBond)) {
+            // Delegating to an account that is not actually a transcoder should render no voting power.
```
yondonfu, line 478: IIUC the `delegatorCumulativeStakeAt()` function is now responsible for the following:

- If the delegator is NOT delegated to a registered transcoder as of the snapshot round, return 0 voting power
- If the delegator is delegated to a registered transcoder as of the snapshot round, return the delegator's pending stake as of the snapshot round

Given this, would a more descriptive name for this function be `delegatorVotesAt()` since the function returns voting power which depends on the delegator's pending stake, but is not necessarily equal to the delegator's pending stake?

If you agree with this, I also suggest updating the comments to reflect this.

contracts/bonding/BondingVotes.sol
```diff
@@ -148,6 +148,10 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
 
     /**
      * @notice Returns the current amount of votes that `_account` has.
+     *
+     * The value returned by this function can be equivalently calculated using {BondingManager} functions as follows:
```
yondonfu: I don't think this is true because if `_account` is a delegator delegated to a non-registered transcoder then this function will return 0 instead of `BondingManager.pendingStake(_account, 0)` as of the snapshot round.

yondonfu:
Not sure if this is already addressed elsewhere, but making a note that there is also a [conditional in _handleVoteOverrides() of GovernorCountingOverridable](https://github.com/livepeer/protocol/blob/delta/contracts/treasury/GovernorCountingOverridable.sol#L184) that does not implement the complete registered transcoder check. This doesn't seem to cause any issues in that function at the moment given the changes in this PR, but could be a good idea to add some clarifications there. A few options come to mind:

a. Update `IVotes.delegatedAt()` to return `(address, bool)` where the bool value indicates whether the delegate is a registered transcoder
b. Update the comments around the referenced conditional explaining why checking if the address == delegate is sufficient + safe despite there not being a complete registered transcoder check since `_weight = 0` and `_voter.deductions = 0` if the address is not a registered transcoder

EDIT: I think https://github.com/livepeer/protocol/pull/626 kind of addresses this, but a bit more info with b) might still be helpful for a future reader.

contracts/bonding/BondingVotes.sol
victorges: Yeah I agree that the implementation here is diverging from simply returning "checkpointed stake". I'm ok with the renaming, tho a weird bit to me would be the `getBondingState` function (which doesn't "know" votes, it comes from that previous implementation of `BondingCheckpoints`) calling a function `delegatorVotesAt`. We could rename `getBondingState` to something else as well (any ideas?), but keep in mind there would be a lot of changes since its a public function used in tests etc (OTOH if the "bonding state" name seems inaccurate as well, we probably want to rename it and reduce the confusing name from a public func)

One alternative could be moving the `transcoderBond` variable creation and `isRegisteredTranscoder` check to the `getBondingState` function. That might need some moving around tho but could be neat, with `isRegisteredTranscoder` only used there. Then it passes both `bond` and `transcoderBond` down to `delegatorCumulativeStakeAt`, 

Let me know which path you prefer!

contracts/bonding/BondingVotes.sol
victorges: Ah, you're right. I thought it would be a simple adaptation from that simpler fix I mentioned (`bondedAmount == 0`). I'll see if I can represent the actual logic simply enough (maybe using english lol), otherwise will just revert this doc change.

victorges:
@yondonfu Regarding not having a full check on `_handleVoteOverrides()`, honestly I decided to skip it to avoid another call to `votes(delegate)` when casting a vote, which could significantly increase the gas costs for a delegator to vote. I think the `a` idea would solve this gas cost issue, tho I think it diverges from `IERC5805.delegates` so would feel a bit weird (could be a better-named `getBondingState()` instead then we make a single call?)

My rationale in the new version of the code is:
- `GovernorCountingOverridable` expects a single invariant: self-delegating accounts should have a voting power **>=** than the sum of all its delegators (only requirement for overriding logic to work properly)
  - This is why I also removed any "isTranscoder" terms from there. It only cares about self-delegation now
- `BondingVotes` implements that invariant, since when the delegate is self-delegate but has no `bondedAmount`, everyone gets `0` voting power

So it'd be an "implementation detail" for the `IVotes` implementation provided. I agree it's very worth to document it on `GovernorCountingOverridable` tho as it could be confusing. I'll do just that but not change any of the implementation, if you think that's ok. If you have a strong preference for `a` or some other solution lmk tho!

contracts/bonding/BondingVotes.sol
yondonfu: What about `getBondingState()` -> `getVotingState()`?

> One alternative could be moving the transcoderBond variable creation and isRegisteredTranscoder check to the getBondingState function. That might need some moving around tho but could be neat, with isRegisteredTranscoder only used there.

This makes sense, but IIUC the result would be:

- The `delegatorCumulativeStakeAt()` name makes sense
- The `getBondingStateAt()` name still doesn't make sense because the function is not necessarily returning bonding state at the provided timepoint since the `uint256 amount` return value could be 0 for a delegator address even if the stake of the delegator is > 0 at the provided timepoint

> but keep in mind there would be a lot of changes since its a public function used in tests

IMO if the changes are primarily in tests then that seems like a reasonable cost (assuming the cost is just a find+replace & bigger diff in this PR) to pay for clearer naming.

yondonfu:
> I agree it's very worth to document it on GovernorCountingOverridable tho as it could be confusing. I'll do just that but not change any of the implementation, if you think that's ok

Sounds reasonable.

contracts/bonding/BondingVotes.sol
victorges: OK I went with:
- `getBondingStateAt` -> `getVotesAndDelegateAtRoundStart`
- `delegatorCumulativeStakeAt` -> `delegatorVotesAtRoundStart`

Even tho the names are a bit long, I wanted to be extra specific there because there's another key difference: these functions return the state at the **start** of the round while the ERC-5805 ones return at the _end_. So this doesn't return the same "votes" as the other functions. I also cleaned up some documentation about it, moving the disclaimers about this difference to these functions, instead of repeating them on all the ERC-5805 functions.

> IMO if the changes are primarily in tests then that seems like a reasonable cost (assuming the cost is just a find+replace & bigger diff in this PR) to pay for clearer naming.

Restricted the changes to function&variable names, documentation and tests. Self-reviewed a couple of times as well and it looks right to me.

contracts/bonding/BondingVotes.sol
victorges: Fixed this on the same commit of `getBondingStateAt` rename as a lot of docs changed there (kept a higher level description here and moved the low-level representation to `getVotesAndDelegateAtRoundStart`)

victorges:
Just realized tests are failing after the merge. Brb.

victorges:
Fixed!

contracts/bonding/BondingVotes.sol
```diff
@@ -349,35 +338,41 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
...
-     */
-    function getBondingStateAt(address _account, uint256 _round)
+     * @notice Gets the voting power and delegate of an account at the start of a given round.
+     *
+     * Notice that this function is different from the {IERC5805Upgradeable} functions above that return the state at
+     * the *end* of the round. The state at the end of a round is equal to the state at the start of the next round, so
+     * to get the same result here, call this function with `round+1` instead.
+     * @dev The value returned by this can also be calculated with the following logic using BondingManager functions at
+     * the start of the corresponding round:
+     * - If `isRegisteredTranscoder(_account)`, the result is `(_account, transcoderTotalStake(_account))`
```
yondonfu: The values in the return tuple here seem to be in reversed order i.e. (address, uint256) instead of (uint256, address) relative to the return tuple definition for the function.

contracts/bonding/BondingVotes.sol
```diff
@@ -349,35 +338,41 @@ contract BondingVotes is ManagerProxyTarget, IBondingVotes {
...
+     * - Otherwise, the `delegate` is obtained from `getDelegator(_account).delegateAddress`
+     *  - If `isRegisteredTranscoder(delegate)`, the result is `(delegate, pendingStake(_account, 0))`
```
yondonfu: The values in the return tuple here seem to be in reversed order i.e. (address, uint256) instead of (uint256, address) relative to the return tuple definition for the function.

yondonfu (changes requested):

contracts/bonding/BondingVotes.sol
victorges: 🤦 I remember I had the right ordering at first and went to check if it was right, but probably saw the func args instead of retval and flipped them here.

victorges:
@yondonfu Fixed

yondonfu (approved):

PR #626, opened 2023-09-20
[delta-audit] treasury: Avoid overriding vote of non self-bonding accounts

victorges:
This fixes a finding from the audit regarding on overriding votes. It addresses the problem of delegating to a non-transcoder account and then overriding their votes imprperly. 

The known corner cases that can trigger this inconsistency are:
- delegating to an account that isn't a transcoder (either another delegator or a non participant)
- transcoder unbonding their full amount (delegators stay delegated to it)

- Update GovernorCountingOverridable to not override votes of accounts that don't self-delegate 

`yarn test` to be deployed on devnet

Fixes:
- https://github.com/code-423n4/2023-08-livepeer-findings/issues/96
- https://github.com/code-423n4/2023-08-livepeer-findings/issues/206

**Checklist:**

### DIFF
--- a/contracts/treasury/GovernorCountingOverridable.sol
+++ b/contracts/treasury/GovernorCountingOverridable.sol
@@ -181,11 +181,18 @@ abstract contract GovernorCountingOverridable is Initializable, GovernorUpgradea
         uint256 timepoint = proposalSnapshot(_proposalId);
         address delegate = votes().delegatedAt(_account, timepoint);
 
-        bool isTranscoder = _account == delegate;
-        if (isTranscoder) {
-            // deduce weight from any previous delegators for this transcoder to
-            // make a vote
+        bool selfDelegates = _account == delegate;
+        if (selfDelegates) {
+            // deduce weight from any previous delegators for this self-delegating account to cast a vote
             return _weight - _voter.deductions;
+        }
+
+        address delegatesDelegate = votes().delegatedAt(delegate, timepoint);
+        bool delegateSelfDelegates = delegate == delegatesDelegate;
+        if (!delegateSelfDelegates) {
+            // do not override votes of non-self-delegating accounts since those don't get their voting power from the
+            // sum of delegations to them, so the override logic doesn't apply.
+            return _weight;
         }
 
         // this is a delegator, so add a deduction to the delegated transcoder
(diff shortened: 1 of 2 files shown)


contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -181,13 +181,20 @@ abstract contract GovernorCountingOverridable is Initializable, GovernorUpgradea
         uint256 timepoint = proposalSnapshot(_proposalId);
         address delegate = votes().delegatedAt(_account, timepoint);
 
-        bool isTranscoder = _account == delegate;
-        if (isTranscoder) {
-            // deduce weight from any previous delegators for this transcoder to
-            // make a vote
+        bool selfDelegates = _account == delegate;
```
yondonfu: IMO `isSelfDelegated` would be easier to read.

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -181,13 +181,20 @@ abstract contract GovernorCountingOverridable is Initializable, GovernorUpgradea
...
+        if (selfDelegates) {
+            // deduce weight from any previous delegators for this self-delegating account to cast a vote
             return _weight - _voter.deductions;
         }
 
+        address delegatesDelegate = votes().delegatedAt(delegate, timepoint);
+        bool delegateSelfDelegates = delegate == delegatesDelegate;
```
yondonfu: IMO `isDelegateSelfDelegated` would be easier to read.

contracts/treasury/GovernorCountingOverridable.sol
```diff
@@ -181,13 +181,20 @@ abstract contract GovernorCountingOverridable is Initializable, GovernorUpgradea
...
-        if (isTranscoder) {
-            // deduce weight from any previous delegators for this transcoder to
-            // make a vote
+        bool selfDelegates = _account == delegate;
+        if (selfDelegates) {
+            // deduce weight from any previous delegators for this self-delegating account to cast a vote
             return _weight - _voter.deductions;
         }
 
+        address delegatesDelegate = votes().delegatedAt(delegate, timepoint);
```
yondonfu: If this var is not used outside of the next line, moving the call into the next line might help with readability since there isn't a need to come up with a good name for this var:

```
bool isDelegateSelfDelegated = delegate == votes().delegatedAt(delegate, timepoint)
```

yondonfu (commented):
Generally looks good and just a few comments.

contracts/treasury/GovernorCountingOverridable.sol
victorges: Good call I struggled with this one 😅

yondonfu (approved):
LGTM

PR #631, opened 2023-10-03
deploy: Delta production deployment scripts

victorges:
This is to update and create deployment scripts in order to update Delta (LIP-91 and LIP-92)
to production.

- Fix dotenv and contract verification, as well as create a task to verify all deployed contracts
- Update `ContractDeployer` to support deploy-only (no register) mode
- Write `deploy_audit.ts` script to update the contracts in question and print required governance commands

- Run the deploy over a local and testnets
- Run the deploy over a fork

Implements PRO-32

**Checklist:**


utils/deployer.ts
victorges, line 1: FTR if not reviewing by commit, I made changes to the `ContractDeployer` in order to:
- Allow deploying proxy without registering them on controller
- Fix the names of the deployments that are saved/overridden
  - Deploying only the target will not set the "main" deployment file
  - Deploying with a proxy will not set the `ManagerProxy` deployment
- Also found out that this fixes the behavior in our deploys which re-deploys contracts even when their bytecode hasn't changed.
  - This happened because we deployed them with mixed up names between proxy and target. So when we deployed a target, hardhat thought it was different than the currently deployed contract which actually pointed to the proxy, and then also redeployed the proxy cause it was almost always different from the last `ManagerProxy` to have been deployed (constructor args).
  - Even though I don't think we should be relying on this on production deploys, it is more efficient on dev/testing so I think it's helpful.

deploy/deploy_delta_upgrade.ts
```diff
@@ -0,0 +1,208 @@
...
+    const contractDeployer = new ContractDeployer(
+        deploy,
+        deployer,
+        deployments,
+        skipRegister
+    )
+    const controller = await contractDeployer.fetchDeployedController()
+
+    const gitCommitHash = await contractDeployer.getGitHeadCommitHash()
+    const governanceActions = []
```
victorges, line 59: I'm also not attached to this logic to print the pending goverance actions, can easily remove it if you prefer less entropy here. It has already helped me write the initial governor script anyway.

victorges:
FTR I've created a fork on Tenderly and ran the deployment script over there to ensure the expected behavior:
- Replaced the [`arbitrumMainnet` RPC URL on the hardhat config](https://github.com/livepeer/protocol/blob/2e9c7735d88976ac713989f35b11c85ac406fb80/hardhat.config.ts#L79) with the the Tenderly fork URL
- Ran `npx hardhat deploy --tags DELTA_UPGRADE --network arbitrumMainnet`
- All transactions ran can be seen here: https://dashboard.tenderly.co/shared/fork/1fe3a622-01c6-4f3c-8bd7-c26903949c0f/transactions
- I only verified the `Treasury` contract for now since it's the only one where functions have been called

Will follow-up later with simulating the governance script to register contracts on controller.

deploy/migrations.config.ts
```diff
@@ -139,8 +139,9 @@ const arbitrumMainnet = {
     },
     bondingManager: {
         numActiveTranscoders: 100,
-        // Rounds
-        unbondingPeriod: 7
+        unbondingPeriod: 7, // 7 rounds
+        treasuryRewardCutRate: 100000, // 10%
```
yondonfu: According to [this line](https://github.com/livepeer/protocol/blob/e1f8b361f5beddac35bde35fa978f79a3d86dadf/contracts/bonding/BondingManager.sol#L1200), this parameter should be a value out of 10^27.

Shouldn't this value be 10^26 in order to represent 10%?

deploy/deploy_contracts.ts
```diff
@@ -159,6 +161,8 @@ const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
             proxy: true,
             args: [Controller.address]
         })
+        // tests expect it to be saved as AdjustableRoundsManager as well
+        await deployments.save("AdjustableRoundsManager", roundsManager)
```
yondonfu, line 165: Did this fix a specific problem with the tests? If so, can you share a brief note on what the issue was?

hardhat.config.ts
```diff
@@ -1,3 +1,6 @@
+import * as dotenv from "dotenv"
```
yondonfu, line 1: FWIW I'm not sure if the use of a .env file for loading environment variables for this config file ever worked properly. In the past, I usually manually set the environment variables in my terminal (i.e. `export ENV_VAR=...`). 

If using `dotenv` here ensures that a .env file can be used then sgtm.

doc/upgrade.md
```diff
@@ -52,6 +52,16 @@ You can also use Tenderly to manually submit contract code for [private verifica
 
 If you see an error related to multiple SPDX license identifiers, remove all SPDX license identifiers from `flattened.sol` except for a single one.
 
+## Task
+
+Alternatively, you can also run the `etherscan-verify-deployments` task from this repository available as a `yarn` script:
```
yondonfu: Any reason to list this as an alternative vs. replacing the docs about running `npx hardhat verify` with the docs here? If the former is the easiest way to verify the source code of all contracts I think it makes sense to just highlight that as the primary method.

utils/deployer.ts
```diff
@@ -86,54 +95,54 @@ export default class ContractDeployer {
         libraries?: Libraries | undefined
     }): Promise<DeployResult> {
         const {contract, name, proxy, args, libraries} = config
-        const targetName = `${name}Target`
 
+        const shouldRegister = this.controller && !this.skipRegister
```
yondonfu: Feels like this function should be renamed to avoid confusion i.e. calling deployAndRegister() but registration does not occur due to the ContractDeployer being configured with skipRegister = true.

Just `deploy()`?

deploy/deploy_delta_upgrade.ts
yondonfu: I'm fine with it for now. For the future, would be good to think about ways to just directly output the list of updates for the Protocol Governor in an upgrade script instead of having to print a JSON object here and copy that over into a different repo.

deploy/deploy_delta_upgrade.ts
```diff
@@ -0,0 +1,208 @@
...
+        executor: await Treasury.EXECUTOR_ROLE(),
+        admin: await Treasury.TIMELOCK_ADMIN_ROLE()
+    }
+    for (const role of [roles.proposer, roles.canceller, roles.executor]) {
+        await Treasury.grantRole(role, livepeerGovernor.address).then(tx =>
+            tx.wait()
+        )
+    }
+
+    await Treasury.renounceRole(roles.admin, deployer).then(tx => tx.wait())
```
yondonfu: Recommend executing this step only after validating that everything is configured properly.

yondonfu (changes requested):

deploy/deploy_contracts.ts
victorges: This fixes a backward incompatibility of the new `ContractDeployer` with the existing tests. The previous deployer code was also registering the deployed contract with the `contract` (implementation) name. So given some tests look for the `AdjustableRoundsManager` directly, I added the additional `save` here to keep them as is without needing to modify them. I avoided as it would be a larger change, but I can make those changes if you prefer.

doc/upgrade.md
victorges: OK, will reorganize!

utils/deployer.ts
victorges: Yeah agreed. I thought of making that change but it would be a bigger refactor so I tried to keep changes to a minimum. If you're seeing value in that direction as well though I'll make a patch to evaluate!

victorges: Implemented on https://github.com/livepeer/protocol/pull/631/commits/fccc671a90258b168b65aba0a0356741876a430f LMK WDYT!

deploy/migrations.config.ts
victorges: You are correct. Second time I get confused with these precisions 😓 
Paying extra care in this last mile.

yondonfu (approved):

victorges:
@yondonfu FYI added 2 more utilities here:
- A task to renounce the admin role from the treasury post deploy/validations
- A document explaining all the deployment steps 
  - (included a deployment validation that I made for the fork, tho haven't included the raw task script here)
  - This other fork test I made is documented in this branch tho I wasn't planning on opening a PR for it: https://github.com/livepeer/protocol/compare/vg/chore/delta-deployment...vg/fork-test-2

LMK if you'd suggest anything, like including that "verify delta deployment" task in here as well or if its fine like this. Tomorrow I plan on merging this and running the contract deploys. Will ping you about how we should merge this into `confluence` branch for the deploy.

victorges:
FTR this is the second fork I made, where I ran both the contract deploys and the governance actions: https://dashboard.tenderly.co/shared/fork/2f6f7f38-4c86-4b24-a165-2b535747ba26/transactions

Unfortunately, controlling the block number of the fork is harder than I thought, so I couldn't advance the round to test treasury contribution. What I might do next is make a fork right at the beginning of a round, so I can initialize it myself after making the deploy.

The code changes for that deploy is on this branch: https://github.com/livepeer/protocol/compare/vg/chore/delta-deployment...vg/fork-test-2
Where I also wrote a task to verify the delta deployment and it is currently passing.

doc/deploy_delta.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+This will print out the transaction that should be run by the governor owner to stage the update
+on the governance contract.
+
+This should now be provided to the wallet interface to stage and then execute the governance update.
+
+## Step 4: Validate the update
+
+You can run the task at
```
yondonfu: Can this script be checked into the `delta` branch as well instead of referencing a file on a separate fork?

The inclusion of a script for verifying a deployment makes sense to me.

doc/deploy_delta.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Step 3: Run governance update
+
+In the `governance-scripts` repository, run the governance update script:
+
+```
+node index.js create ./updates/l2-lip-delta-91-92.js 0
+```
+
+This will print out the transaction that should be run by the governor owner to stage the update
```
yondonfu: Let's make it clear that the "protocol governor" is being referenced here and below to avoid confusion with the LivepeerGovernor being deployed for LIP-91/92.

doc/deploy_delta.md
```diff
@@ -0,0 +1,58 @@
...
+
+To do so, copy that task to your local task folder and then run:
+
+```
+npx hardhat verify-delta-deployment --network arbitrumMainnet
+```
+
+## Step 5: Renounce deployer admin role
+
+Once the governance update has been executed and the update validated (e.g. wait for next round and some treasury
```
yondonfu: IMO the deployer should renounce its admin role after the deployment is validated and before the update is executed. The order of ops would be:

1. Deploy contracts
2. Verify deployed contracts
3. Deployer renounces admin role
4. Stage and execute update with protocol governor
5. Monitor behavior i.e. treasury contributions

The rationale is that there was no expectation set that there would be an admin configured on the Treasury once the upgrade for LIP-91/92 is complete and IMO the existence of an admin after the protocol governor update is staged/executed conflicts with this even if the plan would be to renounce the role after treasury contributions are observed in the next round. Also IIUC, in the event of some issue with treasury contributions in the next round (which is not expected), the protocol governor would need to be the one to initiate a fix anyway.

yondonfu:
> Unfortunately, controlling the block number of the fork is harder than I thought, so I couldn't advance the round to test treasury contribution.

You should be able to control the block number using the [evm_increaseBlocks](https://docs.tenderly.co/simulations-and-forks/reference/tenderly-fork-customization-via-api/how-to-advance-mine-the-block) Tenderly RPC call when connected to the fork JSON RPC URL.

victorges:
@yondonfu 
> You should be able to control the block number using the [evm_increaseBlocks](https://docs.tenderly.co/simulations-and-forks/reference/tenderly-fork-customization-via-api/how-to-advance-mine-the-block) Tenderly RPC call when connected to the fork JSON RPC URL.

Yeah, I did do that and expected it to work, but it just stayed on the same block. I used this task and it has no effect no matter how many blocks I increase:
```ts
import {task} from "hardhat/config"
import {ethers} from "ethers"

const BLOCK_PERIOD = 12 // seconds

task(
    "evm-increase-blocks",
    "Helper task to increase the block number in the current EVM"
)
    .addOptionalPositionalParam(
        "blocks",
        "How many blocks to increase by (defaults to 1)"
    )
    .setAction(async (taskArgs, hre) => {
        const {network} = hre

        const provider = network.provider
        const blocks = parseInt(taskArgs.blocks ?? "1")

        await provider.send("evm_increaseBlocks", [
            ethers.utils.hexValue(blocks) // hex encoded number of blocks to increase
        ])
        // helpfully increase the time by 12s per block as well
        await provider.send("evm_increaseTime", [
            ethers.utils.hexValue(BLOCK_PERIOD * blocks) // hex encoded number of seconds
        ])
    })
```

Maybe it is increasing the L2 block number instead of the observed L1 block number. Couldn't find any docs on increase L1 block number on an arbitrum fork instead.

victorges:
OK I managed to deploy an `AdjustableRoundsManager` as the `RoundsManagerTarget` on the fork, and am able to manipulate the observed L1 block number now as well :)

Was then able to adjust the block to the next round after deploy, initialize that round and run some reward calls observing the 10% of the rewards going to the treasury [1] :)

All transactions inspectable on the fork here [2] and the corresponding code/deployment registry on that same branch [3]

Pending now only updating the deploy docs here and merging the PR into `delta` branch. Then we can start doing the real thing :)

[1] https://dashboard.tenderly.co/livepeer/livepeer/fork/2f6f7f38-4c86-4b24-a165-2b535747ba26/simulation/6b163c5d-9c29-46a4-9cbd-31a8ce446d87
[2] https://dashboard.tenderly.co/shared/fork/2f6f7f38-4c86-4b24-a165-2b535747ba26/transactions
[3] https://github.com/livepeer/protocol/compare/vg/chore/delta-deployment...vg/fork-test-2

victorges:
Unintended close 😅

doc/deploy_delta.md
victorges: Fixed!

yondonfu:
> Maybe it is increasing the L2 block number instead of the observed L1 block number

Ah I think you're right that's probably what is happening.

PR #633, opened 2023-10-13
bonding: Skip fee calculation for voting power

victorges:
This fixes a bug observed after deploying the new BondingVotes to production.

It happens when trying to calculate voting power for a delegator whose transcoder hadn't
initialized their `cumulativeFeeFactor` on the corresponding round.

This causes the voting power calculation to underflow, because we don't initialize the
`cumulativeFeeFactor` of the `endPool` like `BondingManager` does. Even though we 
don't really care about fees for voting power, the underflow causes a revert on any 
transaction that needs to access the delegator votes at the uninitialized round.

- Break `delegatorCumulativeStakeAndFees` logic in 2, to allow calculating only stake or only fees
- Call the stake-only version from BondingVotes
- PENDING: a couple tests

`yarn test` that it didnt change any existing behavior on BondingManager
New tests on BondingVotes
May do a tenderly fork test before actual deploy

N/A

**Checklist:**

### DIFF
--- a/src/test/BondingVotesFeeLessVotesFix.sol
+++ b/src/test/BondingVotesFeeLessVotesFix.sol
@@ -1,0 +1,58 @@
+pragma solidity ^0.8.9;
+
+import "ds-test/test.sol";
+import "./base/GovernorBaseTest.sol";
+import "contracts/Controller.sol";
+import "contracts/bonding/BondingVotes.sol";
+import "contracts/bonding/BondingManager.sol";
+import "./interfaces/ICheatCodes.sol";
+
+// forge test --match-contract BondingVotesFeeLessVotesFix --fork-url https://arbitrum-mainnet.infura.io/v3/$INFURA_KEY -vvv --fork-block-number 140314540
+contract BondingVotesFeeLessVotesFix is GovernorBaseTest {
+    bytes public constant arithmeticError = abi.encodeWithSignature("Panic(uint256)", 0x11);
+
+    BondingManager public constant BONDING_MANAGER = BondingManager(0x35Bcf3c30594191d53231E4FF333E8A770453e40);
+    IBondingVotes public constant BONDING_VOTES = IBondingVotes(0x0B9C254837E72Ebe9Fe04960C43B69782E68169A);
+
+    bytes32 public constant BONDING_VOTES_TARGET_ID = keccak256("BondingVotesTarget");
+
+    BondingVotes public newBondingVotesTarget;
+
+    address public DELEGATOR = 0xdB18A9353139880d73616e4972a855d66C9B69f0;
+
+    function setUp() public {
+        newBondingVotesTarget = new BondingVotes(address(CONTROLLER));
+    }
+
+    function doUpgrade() internal {
+        (, gitCommitHash) = CONTROLLER.getContractInfo(BONDING_VOTES_TARGET_ID);
+
+        stageAndExecuteOne(
+            address(CONTROLLER),
+            0,
+            abi.encodeWithSelector(
+                CONTROLLER.setContractInfo.selector,
+                BONDING_VOTES_TARGET_ID,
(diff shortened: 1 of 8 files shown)


src/test/BondingVotesFeeLessVotesFix.sol
```diff
@@ -0,0 +1,68 @@
+pragma solidity ^0.8.9;
+
+import "ds-test/test.sol";
+import "./base/GovernorBaseTest.sol";
+import "contracts/Controller.sol";
+import "contracts/bonding/BondingVotes.sol";
+import "contracts/bonding/BondingManager.sol";
+import "contracts/snapshots/MerkleSnapshot.sol";
```
yondonfu: Can remove.

src/test/BondingVotesFeeLessVotesFix.sol
```diff
@@ -0,0 +1,68 @@
...
+    BondingManager public constant BONDING_MANAGER = BondingManager(0x35Bcf3c30594191d53231E4FF333E8A770453e40);
+    IBondingVotes public constant BONDING_VOTES = IBondingVotes(0x0B9C254837E72Ebe9Fe04960C43B69782E68169A);
+
+    bytes32 public constant BONDING_VOTES_TARGET_ID = keccak256("BondingVotesTarget");
+
+    BondingVotes public newBondingVotesTarget;
+
+    address public DELEGATOR = 0xdB18A9353139880d73616e4972a855d66C9B69f0;
+
+    event Bond(
```
yondonfu: Can remove.

src/test/BondingVotesFeeLessVotesFix.sol
```diff
@@ -0,0 +1,68 @@
+pragma solidity ^0.8.9;
+
+import "ds-test/test.sol";
+import "./base/GovernorBaseTest.sol";
+import "contracts/Controller.sol";
+import "contracts/bonding/BondingVotes.sol";
+import "contracts/bonding/BondingManager.sol";
+import "contracts/snapshots/MerkleSnapshot.sol";
+import "./interfaces/ICheatCodes.sol";
+import "./interfaces/IL2Migrator.sol";
```
yondonfu: Can remove.

yondonfu (changes requested):
A few minor nits, but generally looks good and await the updated unit tests (which should also fix converage).

test/integration/BondingVotes.js
```diff
@@ -922,5 +934,128 @@ describe("BondingVotes", () => {
...
+                    .fundReserve({value: reserve})
+
+                // Round R+1
+                await nextRound()
+
+                // Now redeem a ticket for the transcoder
+                const creationRound = await roundsManager.currentRound()
+                const creationRoundBlockHash =
+                    await roundsManager.blockHashForRound(creationRound)
+                console.log(creationRoundBlockHash)
```
yondonfu: Can clean up.

yondonfu (changes requested):
Looks like coverage is failing because EarningsPoolLIP36.delegatorCumulativeFees() lacks coverage for initial conditional statements. Those conditionals aren't triggered in tests right now because this function is only ever called by EarningsPoolLIP36.delegatorCumulativeStakeAndFees() which will always initialize _startPool.cumulativeRewardFactor and _endPool.cumulativeRewardFactor if needed before calling delegatorCumulativeFees().

A few options I see to address:

1. Write tests for EarningsPoolLIP36.delegatorCumulativeFees()
2. Inline the fee calculation from delegatorCumulativeFees() into delegatorCumulativeStakeAndFees() after the call to delegatorCumulativeStake()

victorges:
@yondonfu added tests to the `EarningsPoolLIP36` new funcs! Getting 100% coverage on those now

yondonfu (approved):

PR #642, opened 2024-04-12
feat(ai): introduce AIServiceRegistry contract

rickstaa:

This pull request introduces the `AIServiceRegistry` contract. This contract serves as a registry for AI subnet orchestrators, enabling them to register and make their services discoverable within the AI subnet.

- A deployment script to deploy a new `ServiceRegistry` contract under the `AIServiceRegistry` name was added.
- Information about the deployed contracts was commited to the `deployments directory`.

I ensured that all tests passed. Deployed the contract and interacted with the contract using [foundry cast](https://book.getfoundry.sh/cast/).

**Test Deployments**

- ~**Controller:** https://arbiscan.io/tx/0xf5daeb4fee628130c0d6a12a53439baad67c5bc788d811735a89db66cdfcbb22~
- ~**AIServiceRegistryTarget:** https://arbiscan.io/tx/0x1c7282a6a36b88b669201a74caa2f24243024d853e4499ed5d9e8372e803489d~
- ~**AIServiceRegistry:** https://arbiscan.io/tx/0x4f6d608a8aefa913503ff610a5754451a73acc3b5511f005fc3258b7a66a739a~

**EDIT**: We decide to just deploy a target contract connected to the [mainnet controller ](https://arbiscan.io/address/0xD8E8328501E9645d16Cf49539efC04f734606ee4) right now.

- **AIServiceRegistry**: https://arbiscan.io/address/0x04c0b249740175999e5bf5c9ac1da92431ef34c5#writeContract.

https://linear.app/livepeer-ai-spe/issue/LIV-90/design-and-implement-solution-for-advertising-separate-service-uri-for

**Checklist:**


rickstaa:
@yondonfu there are multiple ways to do this. I now created a seperate `AIServiceRegistry` contract that inherits all properties from the regular `ServiceRegistry` contract but we can also keep one `ServiceRegistry` contract and deploy the `AIServiceRegistry` contract based on that.

I've deployed it using my own developer wallet, along with a brand new controller contract. However, it might be more effective if someone else on the team handles the deployment so that it can be directly linked to the main [controller](https://arbiscan.io/address/0xD8E8328501E9645d16Cf49539efC04f734606ee4) contract. One potential issue with this approach is the absence of a `RemoveContractInfo` method in the [controller](https://github.com/livepeer/protocol/blob/delta/contracts/Controller.sol) contract, as far as I am aware. This means we might only be able to reset it to a default setting should we decide to consolidate the two Service Registries in the future 🤔.

rickstaa:
@yondonfu I also noticed although the proxy contract works automatic verification seems to fail for this contract (see https://arbiscan.io/address/0x73f1755f34c2e769209e0f91a0c385722ed81b9c#code). Are you familiar with this problem?

deploy/deploy_ai_service_registry.ts
```diff
@@ -0,0 +1,25 @@
+import {HardhatRuntimeEnvironment} from "hardhat/types"
+import {DeployFunction} from "hardhat-deploy/types"
+
+import {Controller} from "../typechain"
+
+import ContractDeployer from "../utils/deployer"
+
+const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
```
yondonfu, line 4: I suggest following the structure of [this deploy script](https://github.com/livepeer/protocol/blob/delta/deploy/deploy_ticket_broker.ts) since in practice this deploy script would just deploy a new instance of the ServiceRegistry contract without deploying a new Controller.

Something like:

```
const deployResult = await deploy("ServiceRegistry", {
    from: deployer,
    args: [controllerDeployment.address],
    log: true
})
await deployments.save("AIServiceRegistry", deployResult)
```

- The first line would deploy a *non-proxy* version of the ServiceRegistry contract.
- The second line would save the metadata for the deployment under the name "AIServiceRegistry" in the `deployments` directory

deploy/deploy_contracts.ts
```diff
@@ -180,6 +180,14 @@ const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
         args: [Controller.address]
     })
 
+    // AIservice registry
+    await contractDeployer.deployAndRegister({
```
yondonfu: Since the AIServiceRegistry is a separate contract that wouldn't be a part of the core protocol contracts I don't think this deploy is necessary here and we also wouldn't be registering the AIServiceRegistry address with the Controller for now.

yondonfu (commented):
> there are multiple ways to do this. I now created a seperate AIServiceRegistry contract that inherits all properties from the regular ServiceRegistry contract but we can also keep one ServiceRegistry contract and deploy the AIServiceRegistry contract based on that.

IMO the latter option would be simpler - if we just deploy another instance of the ServiceRegistry code and refer to it as the AIServiceRegistry that would remove the need for additional tests or custom code that specifically references AIServiceRegistry by name.

yondonfu:
> I also noticed although the proxy contract works automatic verification seems to fail for this contract

If a contract is a proxy, you would have to use the arbiscan UI to confirm it is a proxy, make sure it is pointing at the right target implementation address.

Screenshot 2024-04-15 at 2 47 14 PM

Though, as mentioned in my previous comment, if a non-proxy version of the ServiceRegistry is deployed as the AIServiceRegistry then won't have to deal with proxy verification on arbiscan.

deploy/deploy_ai_service_registry.ts
rickstaa: @yondonfu, understood. I've already deployed a single `ServiceRegistry` as `AIServiceRegistry` on [sepolia](https://sepolia.arbiscan.io/) so deploying one on Abritrum should cause no problems.

I now kept the AIServiceRegistry contract closer to the mainnet implementation. However, I agree that doing so might unnecessarily complicate matters. We can always consider migrating later if we decide to merge the subnet into the mainnet 👍🏻.

deploy/deploy_contracts.ts
rickstaa: I agree if we don't want to include it yet in the mainnet contracts we can get rid of the controller and Proxy for now 👍🏻.

rickstaa:
> > I also noticed although the proxy contract works automatic verification seems to fail for this contract
> 
> If a contract is a proxy, you would have to use the arbiscan UI to confirm it is a proxy, make sure it is pointing at the right target implementation address.
> 
> Screenshot 2024-04-15 at 2 47 14 PM
> 
> Though, as mentioned in my previous comment, if a non-proxy version of the ServiceRegistry is deployed as the AIServiceRegistry then won't have to deal with proxy verification on arbiscan.

Yea looks like the target is verified automatically causing the automatic Proxy verification on Arbitrum explorer to fail 🤔. @iameli stated that he also had problems with that some time ago and was able to get it verified manually. However let's for now just simply things by deploying a non-proxy version 👍🏻.

deploy/deploy_ai_service_registry.ts
rickstaa: @yondonfu done -> https://arbiscan.io/address/0x04c0b249740175999e5bf5c9ac1da92431ef34c5#writeContract.

rickstaa:
@yondonfu, I've re-deployed the contract as an immutable standalone contract. The potential drawback is that if we decide to register this contract with the controller for future upgrades, similar to the standard ServiceRegistry contract, we would need to deploy a new contract. This change would require us to either migrate the existing service URIs to the new contract (though it’s unclear if this is feasible) or compel orchestrators to resubmit their URIs. It's manageable, but worth considering. Additionally, I've set up https://github.com/livepeer/protocol/pull/643, which, from my understanding, would allow us to register both the Proxy and target with the controller 🤔. I'm fine with using any of the two versions 👍🏻.

yondonfu:
@rickstaa IMO using an immutable standalone contract at this stage would be preferable because deploying an upgradeable proxy involves additional complexity:

- If the AIServiceRegistry is deployed using the ManagerProxy pattern (which is how all core protocol contracts are deployed), then the Governor would be the only address that is able to upgrade the implementation.
- If the AIServiceRegistry is deployed with another upgradeable proxy pattern, then there is still some admin address that needs to be defined that would be able to upgrade the implementation.

In both cases, the operational setup for the contract becomes more complex and I think the tradeoff of potentially having to migrate service URIs to another contract in the future is reasonable to avoid that extra complexity especially since the path to "merging" the AI subnet and the transcoding mainnet at this point is still unclear.

rickstaa:
> @rickstaa IMO using an immutable standalone contract at this stage would be preferable because deploying an upgradeable proxy involves additional complexity:
> 
> * If the AIServiceRegistry is deployed using the ManagerProxy pattern (which is how all core protocol contracts are deployed), then the Governor would be the only address that is able to upgrade the implementation.
> * If the AIServiceRegistry is deployed with another upgradeable proxy pattern, then there is still some admin address that needs to be defined that would be able to upgrade the implementation.
> 
> In both cases, the operational setup for the contract becomes more complex and I think the tradeoff of potentially having to migrate service URIs to another contract in the future is reasonable to avoid that extra complexity especially since the path to "merging" the AI subnet and the transcoding mainnet at this point is still unclear.

@yondonfu, thank you for your response; it is helpful. Ultimately, I decided to proceed with the immutable standalone contract (i.e. https://github.com/livepeer/go-livepeer/pull/3004).

yondonfu:
@rickstaa If this is deployed and being used already suggest moving this out of draft so it can be properly reviewed + merged.

rickstaa:
> @rickstaa If this is deployed and being used already suggest moving this out of draft so it can be properly reviewed + merged.

Great idea! I wasn't sure if I should keep it as a draft until we merge with Mainnet but happy to move it out of draft 👍🏻.

yondonfu (approved):

PR #645, opened 2025-03-26
minter: Add inflation bounds to inflation adjustment algorithm

SidestreamSweatyPumpkin:

This PR extends the [original PR](https://github.com/livepeer/protocol/pull/644) (introduced inflation bounds for the Minter) with tests and other changes required for deploying the contract and merging the PR.

- Added unit tests for the `maxInflation` and `minInflation` constructor arguments, setters and the updated inflation logic
- Added functionality to seamlessly transfer state of the old Minter into the new one
  - Ensuring inflation doesn’t jump up or down due to the migration
  - Ensuring users receive their rewards for the round even during the update
- Added governor-level integration tests for the end-to-end Minter migration
- Updated relevant deployment scripts
- Applied small fixes:
  - Removed redundant check for the `minInflation` lower bound
  - Updated `deploy_contracts.ts` logic to match on-chain state
  - Updated `IMinter.sol` interface
  - Fixed typos in the comments / test names

We used the existing hardhat setup to extend the coverage for the newly added code:
- New tests for the new constructor arguments
  - `maxInflation` constructor argument can’t be > 100%
  - `minInflation` constructor argument can’t be > 100%
  - `minInflation` constructor argument can’t be > `maxInflation`
  - `minInflation` constructor argument can be same as `maxInflation` and can be `0`
- New tests for the new setter functions
  - `setMinInflation` function can only be called by the Controller owner
  - `setMinInflation` function fails if provided `minInflation` > 100%
  - `setMinInflation` function fails if provided `minInflation` > `maxInflation`
  - `setMinInflation` can set `minInflation` to 0, `maxInflation` or other valid value
  - `setMaxInflation` function can only be called by the Controller owner
  - `setMaxInflation` function fails if provided `maxInflation` > 100%
  - `setMaxInflation` function fails if provided `maxInflation` < `minInflation`
  - `setMaxInflation` can set `maxInflation` to 0, `minInflation` or other valid value
- New tests for the new function to migrate old minter state
  - `migrateOldMinterState` function can only be called by the Controller owner
  - `migrateOldMinterState` function can only be called when old minter is still registered in the controller
  - `migrateOldMinterState` function successfully transfers old minter variables to the current minter
  - New tests for the updated inflation logic
  - `inflation` is increased up to `maxInflation` IF `current < targetBondingRate`
  - `inflation` is increased up to `minInflation` IF `inflation < minInflation` (independent from the bonding rate)
  - `inflation` is decreased down to `minInflation` IF `current > targetBondingRate`
  - `inflation` is decreased down to `maxInflation` IF `inflation > maxInflation` (independent from the bonding rate)
  - `inflation` is maintained IF `current = targetBondingRate` AND within `minInflation` and `maxInflation`
  - `inflation` is maintained IF `minInflation = maxInflation`(independent from the bonding rate)
- New integration tests
  - “Complex update: migrate to new Minter” test is updated to mimic real-world update sequence:
    1. Call `migrateToNewMinter` on the previous Minter
    2. Call `migrateOldMinterState` on the new Minter
    3. Grant `MINTER_ROLE` to the new Minter
    4. Revoke `MINTER_ROLE` from the previous Minter
    5. Register new Minter in the Controller
  - `MinterUpgrade` that previously ensured that internal state is zeroed, changed to ensure state of the 3 “dynamic” variables is correctly migrated
  - `MinterUpgrade` that previously ensured that user received 0 rewards after the upgrade is changed to ensure received reward is greater than 0

We also took an opportunity to improve the CI by:
- Adding an additional `test:coverage:check` command that `exit`s with non-zero code, if coverage is not 100%
- Adding an additional step in the `test.yaml` workflow that executes the above command

To prepare for the new Minter deployment, we’ve also manually done the following checks:
- ℹ️ Checked current production Minter contract ([found in the docs](https://docs.livepeer.org/references/contract-addresses)) deployed at [0xc20DE37170B45774e6CD3d2304017fc962f27252](https://arbiscan.io/address/0xc20DE37170B45774e6CD3d2304017fc962f27252) on Feb-17-2022
- ✅ The verified code on Arbiscan matches[ f5facdf](https://github.com/livepeer/protocol/blob/f5facdf5eb98368449c0092fe201940dd73c3067/contracts/token/Minter.sol) commit
- ✅ Inspected changes to the contract since the last deployment: only the imported contracts were changed, but none of the changes are critical:
  - `@openzeppelin/contracts/token/ERC20/IERC20.sol` dependency was updated. Here is the [diff](https://www.diffchecker.com/T8ULLzCl/): only comments and argument names were changed, this has no effect on the actual bytecode
  - `@openzeppelin/contracts/utils/math/SafeMath.sol` dependency was updated. Here is the [diff](https://www.diffchecker.com/vtUdmF5Q/): only comments and formatting of the code were changed, this has no effect on the actual bytecode
  - `contracts/bonding/IBondingManager.sol` was updated. Here is the [diff](https://www.diffchecker.com/i4Fj3AMo/): 1 new event and 2 new functions were added, unused by the Minter contract

No, but it addresses the discussion started on [Livepeer Forum](https://forum.livepeer.org/t/inflation-focused-lip-discussion-thread/2753) and in the [original PR](https://github.com/livepeer/protocol/pull/644).

**Checklist:**

  ℹ️ We found no relevant documentation of the Minter contract inside this repository

**Next steps**
- Receive reviews from the core contributors
- Set actual `maxInflation` and `minInflation` values for the deployment (although not a blocker for this PR, since those values can be changed after the deployment during the actual governor update)


SidestreamCrunchyCarrot:
Suggested further actions:
- [x] Increase test coverage on the Minter contract to 100% (currently it's `98.28`)
- [ ] ~~Increase test coverage on other `treasury/*` contracts~~ [it seems to be excluded from coverage](https://github.com/livepeer/protocol/blob/e8b6243c48d9db33852310d2aefedd5b1c77b8b6/.solcover.js#L12-L13)
- [x] Enforce [coverage threshold](https://github.com/sc-forks/solidity-coverage/blob/master/docs/advanced.md#coverage-threshold-checks) to be 100% in the CI (as `delta` branch already have full coverage)

package.json
```diff
@@ -21,6 +21,7 @@
     "clean": "rm -rf cache artifacts typechain",
     "compile": "npx hardhat compile",
     "test:coverage": "npx hardhat coverage",
+    "test:coverage:check": "npx istanbul check-coverage ./coverage.json --statements 100 --branches 100 --functions 100 --lines 100",
```
victorges, line 24: Nice!

deploy/deploy_contracts.ts
```diff
@@ -372,6 +374,8 @@ const func: DeployFunction = async function(hre: HardhatRuntimeEnvironment) {
 
     await (await Token.grantRole(MINTER_ROLE, minter.address)).wait()
 
+    await (await Token.grantRole(DEFAULT_ADMIN_ROLE, governor.address)).wait()
```
victorges, line 377: Could you explain why is this required?

deploy/migrations.config.ts
```diff
@@ -161,7 +169,9 @@ const arbitrumMainnet = {
         // The inflation at the start of that round will be 218500 - 500 = 218000
         inflation: 218500,
         inflationChange: 500,
-        targetBondingRate: 500000000
+        targetBondingRate: 500000000,
+        maxInflation: 250000,
+        minInflation: 5000
```
victorges: Can you increment the comment above explaining these values?

contracts/token/Minter.sol
```diff
@@ -138,6 +183,21 @@ contract Minter is Manager, IMinter {
...
+     */
+    function migrateOldMinterState() external onlyControllerOwner {
+        IMinter oldMinter = IMinter(controller.getContract(keccak256("Minter")));
+        // Old Minter cannot be the current Minter
+        require(address(oldMinter) != address(this), "old Minter cannot be current Minter");
+
+        // Transfer state from old Minter
+        currentMintableTokens = oldMinter.currentMintableTokens();
+        currentMintedTokens = oldMinter.currentMintedTokens();
+        inflation = oldMinter.inflation();
```
victorges, line 198: Should we copy all fields including `inflationChange` and `targetBondingRate` to make sure we get a consistent state snaphot? Otherwise, we'd be relying on the transaction that constructed the contract to have initialized it with the same values as the current minter. 

We usually run the contract creation separately so that we have a fixed address for the governor transaction. In that case it probably makes sense to make the upgrade tx "self-contained", copying all state, not relying on args configured in other txs. WDYT?

test/integration/GovernorUpdate.js
```diff
@@ -132,75 +158,169 @@ describe("Governor update", () => {
...
+            }
+
+            // Run the migrate to new minter update
+            await governor.stage(update, "0")
+            await expect(governor.execute(update)).to.be.revertedWith(
+                "new Minter must be registered"
+            )
+        })
 
+        it("step 2 'transferStateData' fails: wrong target", async () => {
```
victorges: maybe better to use method name on test name
```suggestion
        it("step 2 'migrateOldMinterState' fails: wrong target", async () => {
```

test/integration/GovernorUpdate.js
```diff
@@ -132,75 +158,169 @@ describe("Governor update", () => {
...
+                    revokeRoleData,
+                    setInfoData
+                ],
                 nonce: 0
             }
 
-            // run the migrate to new minter update
+            // Run the migrate to new minter update
             await governor.stage(update, "0")
             await expect(governor.execute(update)).to.be.reverted
```
victorges: Can you check that the error is `Minter cannot be current Minter`?

test/unit/Minter.js
```diff
@@ -861,11 +1112,72 @@ describe("Minter", () => {
...
+                minter.address,
+                functionSig("setCurrentRewardTokens()")
+            )
+
+            const endInflation = await minter.inflation()
+            assert.equal(
+                startInflation.sub(endInflation).toNumber(),
+                await minter.inflationChange(),
+                "inflation rate did decrease correctly"
+            )
```
victorges: How does the `startInflation.sub(BigNumber.from(PERC_MULTIPLIER))` above compare to the current inflation/to the `inflationChange`? I was thinking we probably a want a test somewhat like this:
- inflation: 1000; inflation change 100
- max inflation updated to 500
- on the next `setCurrentRewardTokens` we want the inflation to always go to 900 and not 500 nor 1100

Is this already it? Or are we possibly setting the `maxInflation` to something like 950 and then checking if it updates to 900 and not 850?

contracts/token/Minter.sol
```diff
@@ -240,13 +300,20 @@ contract Minter is Manager, IMinter {
             currentBondingRate = MathUtils.percPoints(totalBonded, totalSupply);
         }
 
-        if (currentBondingRate < targetBondingRate) {
+        // Adjust inflation based on current bonding rate and target bonding rate, ensuring it stays within the floor and ceiling
+        if ((currentBondingRate < targetBondingRate && inflation < maxInflation) || inflation < minInflation) {
```
victorges: Interesting, so we want the inflation update to be gradual, even if the governors set a min/max inflation way lower than current inflation. I thought it could happen that we'd immediately set the inflation to the given min/max on the next round, but this is probably better for tokenomics stability and requires a strategy from governors.

deploy/deploy_contracts.ts
SidestreamSweatyPumpkin: In order for the new Minter address [to be able to mint new tokens](https://github.com/livepeer/protocol/blob/48e4f36745d44a176545e1c0fd644b5c166ffd6e/contracts/token/Minter.sol#L219) (and the old minter to loose this ability), we have to call `LPT.grantRole('MINTER_ROLE', newMinter)` and `LPT.revokeRole('MINTER_ROLE', currentMinter)` during the update ([see the role check](https://github.com/livepeer/protocol/blob/e8b6243c48d9db33852310d2aefedd5b1c77b8b6/contracts/token/LivepeerToken.sol#L28C58-L28C79)). To grant or revoke roles the Governor [have to have admin rights](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e50c24f5839db17f46991478384bfda14acfb830/contracts/access/AccessControl.sol#L145C79-L145C107) on the LPT. 

It turns out that the Governor already have said rights on-chain (see screenshot from [the arbiscan](https://arbiscan.io/address/0x289ba1701C2F088cf0faf8B3705246331cB8A839#readContract#F7) below). We could've just mocked this role in the particular integration test, but since this deploy script is used by the tests, we've decided to modify it instead, making all tests closer to the on-chain state. Let me know if it's not a desired approach.

Screenshot 2025-04-22 at 09 33 54

contracts/token/Minter.sol
SidestreamSweatyPumpkin: There are two reasons why we thought it would rather be more confusing:
- We wouldn't be able to copy all possible fields, since `maxInflation` and `minInflation` are not part of the currently deployed contract. In other words it would be confusing during the next upgrade that everything is copied except for those 2 values
- The overwrite of other arguments is redundant and can create additional confusion. For example, in case governance decides to also change some of the configurable parameters (e.g. `inflationChange`) during the Minter upgrade, it will result in 1) deploying the contract with the new or some values 2) calling `migrateOldMinterState` that will reset them to the old state 3) calling `set*` functions to set them to the new values

Based on this, we've decided to update only the minimum required state that is dynamically changing.

contracts/token/Minter.sol
SidestreamSweatyPumpkin: To be clear: this code is from the commit done in the [original PR](https://github.com/livepeer/protocol/pull/644). But we also think it makes sense to slowly reach the new bounds instead of hard-resetting the inflation.

deploy/migrations.config.ts
SidestreamSweatyPumpkin: I've added the `TODO` note above, as those are arbitrary values at the moment that will need to be updated once the community decides on the final approach to the inflation bounds [here](https://forum.livepeer.org/t/inflation-focused-lip-discussion-thread/2753).

Also, the comment above the block ([lines 165-169](https://github.com/livepeer/protocol/blob/48e4f36745d44a176545e1c0fd644b5c166ffd6e/deploy/migrations.config.ts#L165-L169)) currently talks about why `218500` inflation value is chosen here (due to L1->L2 migration happened back in 2022) and presents a formula to compute this number. Since we're not going to migrate Minter across layers and plan to use `migrateOldMinterState` function, this information is no longer relevant and can simply be removed rather than being updated. Please let me know if you have a different suggestion

test/unit/Minter.js
SidestreamSweatyPumpkin: > How does the `startInflation.sub(BigNumber.from(PERC_MULTIPLIER))` above compare to the current inflation/to the `inflationChange`?

Currently, the `inflationChange` is set to be 50 times smaller than `PERC_MULTIPLIER`:

https://github.com/livepeer/protocol/blob/48e4f36745d44a176545e1c0fd644b5c166ffd6e/test/unit/Minter.js#L28

So this test follows your logic:

- `inflation`: `26 * PERC_MULTIPLIER`, inflation change: `0.02 * PERC_MULTIPLIER`
- `max inflation set to`: `25 * PERC_MULTIPLIER`
- on the next `setCurrentRewardTokens` the inflation goes to `startInflation - INFLATION_CHANGE`

I've also improved tests to ensure bonding rate has no effect on those changes whether it's lower, higher or equal to the target bonding rate.

deploy/migrations.config.ts
SidestreamSweatyPumpkin: The values and the comment are updated to match [LIP-100](https://github.com/livepeer/LIPs/blob/master/LIPs/LIP-100.md) in 78176a0e5ccc85d939cd0953c53941711f770e22

deploy/deploy_contracts.ts
victorges: Sounds good!

contracts/token/Minter.sol
victorges: The problem is that the contracts are not created on the governance transaction, so someone would need to have run a separate transaction first to deploy the new `Minter` contract. This transaction would not be on `governor-scripts` repo so it would need to be reviewed separately, creating further complexity on the deployment flow (or at least that's how I remember we did it before).

So IMO an extra transaction to change any other parameters that the governance wants would be preferred, as it would also be very explicit (and not set on a "deploy contract" transaction that is not necessarily audited on the governance process).

Regarding not copying the `maxInflation` and `minInflation`, I think it could be covered by a local comment saying that the fields were added with the migration function, so they should be copied if a new version is made. WDYT?

contracts/token/IMinter.sol
```diff
@@ -29,9 +29,9 @@ interface IMinter {
 
     function inflation() external view returns (uint256);
 
-    function maxInflation() external view returns (uint256);
+    function inflationCeiling() external view returns (uint256);
 
-    function minInflation() external view returns (uint256);
+    function inflationFloor() external view returns (uint256);
```
victorges: I like the rename! Make sure there's no occurences of `[Mm]ax.+[Ii]nflation` on the project anymore 👀

contracts/token/Minter.sol
```diff
@@ -131,33 +131,33 @@ contract Minter is Manager, IMinter {
...
     }
 
     /**
-     * @notice Set minInflation. Only callable by Controller owner
-     * @param _minInflation New inflation floor as a percentage of total token supply
+     * @notice Set inflationFloor. Only callable by Controller owner
+     * @param _inflationFloor New inflation floor as a percentage of total token supply
      */
-    function setMinInflation(uint256 _minInflation) external onlyControllerOwner {
+    function setinflationFloor(uint256 _inflationFloor) external onlyControllerOwner {
```
victorges: Probably mis-replaced the name here:
```suggestion
    function setInflationFloor(uint256 _inflationFloor) external onlyControllerOwner {
```

Will need to rename in other places like tests etc

contracts/token/Minter.sol
```diff
@@ -131,33 +131,33 @@ contract Minter is Manager, IMinter {
     }
 
     /**
-     * @notice Set maxInflation. Only callable by Controller owner
-     * @param _maxInflation New inflation cap as a percentage of total token supply
+     * @notice Set inflationCeiling. Only callable by Controller owner
+     * @param _inflationCeiling New inflation cap as a percentage of total token supply
      */
-    function setMaxInflation(uint256 _maxInflation) external onlyControllerOwner {
+    function setinflationCeiling(uint256 _inflationCeiling) external onlyControllerOwner {
```
victorges: Same
```suggestion
    function setInflationCeiling(uint256 _inflationCeiling) external onlyControllerOwner {
```

test/unit/Minter.js
```diff
@@ -1063,177 +1076,207 @@ describe("Minter", () => {
...
+                inflationCeiling.toNumber(),
+                "inflation rate did not set to inflationCeiling"
             )
         })
 
-        it("inflation is increased IF inflation < minInflation (independent from bonding rate)", async () => {
-            const startInflation = await minter.inflation()
+        // Ensure the logic is unaffected by the target bonding rate
+        for (const targetBondingRate of [400, 500, 600]) {
+            it("inflation is increased IF inflation < inflationFloor (independent from bonding rate)", async () => {
```
victorges: **nit:** Even better than the parenthesis would be specifying the bonding rate on the test title
```suggestion
            it($"inflation is increased IF inflation < inflationFloor (targetBondingRate=${targetBondingRate})", async () => {
```

Same below

victorges (dismissed):
LGTM! The only required change is fixing the name of the `setinflationCeiling/Floor` functions

contracts/token/Minter.sol
SidestreamSweatyPumpkin: > creating further complexity on the deployment flow (or at least that's how I remember we did it before).

I'm not sure where the additional complexity is coming from, as the deployed contract and its state have to be reviewed anyway. Typically, the review of the contract should be part of the standard "governance payload review": if the payload mentions/interacts with a new contract, the code as well as the state have to be checked. In my opinion, having 2 places where a state variable is being modified is rather confusing than "explicit".

> So IMO an extra transaction to change any other parameters that the governance wants would be preferred, as it would also be very explicit (and not set on a "deploy contract" transaction that is not necessarily audited on the governance process).

Same problem here: do you then would also suggest to set inflation ceiling and floor in the governance payload to make it explicit? (although those values have already been set at deployment via the constructor arguments)

> Regarding not copying the `maxInflation` and `minInflation`, I think it could be covered by a local comment

If we still would go with updating `migrateOldMinterState` scope, what do you mean by the "local" comment? Some kind of TODO comment inside the function (e.g. "TODO: add inflationCeiling and inflationFloor once old miter has those values")?

PR #646, opened 2025-04-02
Ensure the treasury automatically resumes once balance < cap

stronk-dev:
Ensures the treasury automatically resumes instead of setting the cut to 0% once the cap is reached. This way, it will continue accumulating as the balance depletes.

Wondering if this can be a good solution instead of raising the cap. 
Note that this will cause an emitted `ParameterUpdate` event for the `treasuryRewardCutRate` after each round it flips below/above the cap.
This change does not affect the subgraph, as that is not updated to take the treasury into account at all at the moment.

- Removes logic in reward calls which modifies `nextRoundTreasuryRewardCutRate` when the cap is reached.
- Modify Round init to set the cut for _the current round only_ to 0 for as long as the cap is reached.

N/A

N/A

**Checklist:**


stronk-dev:
Tests need updating for this as well.

PR #648, opened 2025-09-12
Add reward caller (using transcoder-to-reward caller mapping direction)

SidestreamSweatyPumpkin:
### What does this pull request do? Explain your changes. (required)

This PR adds a new feature to the livepeer protocol which will allow transcoders to set a single address they trust to claim the rewards every on round without requiring the main wallet to stay unlocked.

### Specific updates (required)

- `contracts/bonding/BondingManager.sol` contract is modified to include
  - New `transcoderToRewardCaller` public mapping
  - New `setRewardCaller` function to set or unset a `rewardCaller`
  - New `rewardForTranscoder`, `rewardForTranscoderWithHint` functions to claim rewards using `rewardCaller` account
- `test/unit/BondingManager.js` test is modified to include new "reward delegation" section
- `yarn test:coverage:check` command is added to the CI to insure 100% test coverage

### How did you test each of these updates (required)

A new "reward delegation" section contains tests that ensure:
- A transcoder is able to to call the reward even if RewardCaller is set for them
- If a RewardCaller is set, the account is able to call `rewardForTranscoder`
- If the RewardCaller is unset, it no longer able to call `rewardForTranscoder`
- Checkpoint logs reward recipient, not the RewardCaller

### Does this pull request close any open issues?

No

### Checklist:

  - No relevant documentation found, the changes are backwards compatible


SidestreamSweatyPumpkin:
### Additional notes on the implementation

#### Note-01:
It is possible for a transcoder to set the RewardCaller as themself and call functions designated for RewardCallers. By doing so, the transcoder itself will be able to use the `rewardForTranscoder` and `rewardForTranscoderWithHint`. But other than being able to call both functions, there is no security risk identified. Note: we didn't include this edge case in the tests as this is not intended use-case.

#### Note-02:
It is possible for a TranscoderA to set the TranscoderB as RewardCaller. In this case, TranscoderB will be able to claim reward for themself by calling `reward()` and can claim reward for TranscoderA by calling `rewardForTranscoder(TranscoderA)`. We didn't identify any security risk associated with this edge case.

#### Note-03:
In this implementation it is possible that multiple different transcoders set the same address as their RewardCaller. It will allow the same RewardCaller to call `rewardForTranscoder(TranscoderA)`, `rewardForTranscoder(TranscoderB)` or even use multi-call functionality to save on gas.

#### Note-04:
It is not part of the current implementation, but it's possible to add a functionality to self-revoke RewardCaller. Currently, only the transcoder which set the RewardCaller can unset it by calling `setRewardCaller` with a different (or a zero) address.

contracts/bonding/BondingManager.sol
```diff
@@ -188,6 +191,16 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
         emit ParameterUpdate("numActiveTranscoders");
     }
 
+    /**
+     * @notice Set a reward caller for a transcoder
+     * @param _rewardCaller Address of the new reward caller
+     * @dev By providing address(0) the reward caller can be unset
+     */
+    function setRewardCaller(address _rewardCaller) external whenSystemNotPaused {
+        transcoderToRewardCaller[msg.sender] = _rewardCaller;
```
yondonfu, line 201: Just noting that in spite of the mapping name, anyone can call this function so the address key in the mapping may not correspond to a transcoder. I don't see an immediate problem with this, but wanted to a) confirm that this property has been considered and b) suggest some in-line documentation if this property has been deemed a non-issue.

contracts/bonding/BondingManager.sol
```diff
@@ -863,21 +885,53 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+
+    /**
+     * @notice Mint token rewards for an active transcoder and its delegators and update the transcoder pool using an optional list hint if needed
+     * @dev If the `_transcoder` is in the transcoder pool, the caller can provide an optional hint for its insertion position in the
+     * pool via the `_newPosPrev` and `_newPosNext` params. A linear search will be executed starting at the hint to find the correct position.
+     * In the best case, the hint is the correct position so no search is executed. See SortedDoublyLL.sol for details on list hints
+     * @dev Only callable by trusted rewardCaller
+     * @param _transcoder Address of the transcoder on behalf of which the reward is called
+     * @param _newPosPrev Address of previous transcoder in pool if the `_transcoder` is in the pool
+     * @param _newPosNext Address of previous transcoder in pool if the `_transcoder` is in the pool
```
rickstaa: @SidestreamSweatyPumpkin

```suggestion
     * @param _newPosNext Address of next transcoder in pool if the `_transcoder` is in the pool
```

contracts/bonding/BondingManager.sol
SidestreamSweatyPumpkin: @yondonfu
Thanks for checking that. To reply to your two specific points:
- (a) We originally considered adding `isRegisteredTranscoder` check to the setter. However, it doesn’t give any additional protection (a registered transcoder is easy to circumvent if you’re an attacker), but potentially makes fresh account setup harder (and therefore go-client implementation more complex). We identified no issues with keeping the setter open to anyone.
- (b) Good point, the mapping name could've been more abstract. But since it's already mentioned in the LIP, [I've added a `@dev` comment](https://github.com/livepeer/protocol/blob/99c00089b760d6980cb2e44fee3a2aad715657b0/contracts/bonding/BondingManager.sol#L106) to the mapping itself, indicating that the key can be a non-transcoder: “`@dev` Since the setter is callable by any address, a mapping key does not guarantee to be a registered or active transcoder”.

contracts/bonding/BondingManager.sol
SidestreamSweatyPumpkin: @rickstaa 
Thanks for catching this! It has been addressed in [d81d76b](https://github.com/livepeer/protocol/pull/648/commits/d81d76b8044358ef6242b7d8d99ea555881292d7)

yondonfu (approved):

package.json
```diff
@@ -21,6 +21,7 @@
     "clean": "rm -rf cache artifacts typechain",
     "compile": "npx hardhat compile",
     "test:coverage": "npx hardhat coverage",
+    "test:coverage:check": "npx istanbul check-coverage ./coverage.json --statements 100 --branches 100 --functions 100 --lines 100",
```
rickstaa, line 24: @SidestreamSweatyPumpkin https://www.npmjs.com/package/istanbul has ben deprecated should be replaced by https://github.com/istanbuljs/nyc.

contracts/bonding/BondingManager.sol
```diff
@@ -863,21 +886,53 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
...
+        address _newPosNext
+    ) private whenSystemNotPaused currentRoundInitialized autoCheckpoint(_transcoder) {
         uint256 currentRound = roundsManager().currentRound();
 
-        require(isActiveTranscoder(msg.sender), "caller must be an active transcoder");
+        require(isActiveTranscoder(_transcoder), "transcoder must be active");
         require(
-            transcoders[msg.sender].lastRewardRound != currentRound,
+            transcoders[_transcoder].lastRewardRound != currentRound,
             "caller has already called reward for the current round"
```
rickstaa, line 932: @SidestreamSweatyPumpkin this misnames the transcoder as the caller. Not a big problem just something I noticed.

package.json
SidestreamSweatyPumpkin: Thanks for the proposal. We can update it in the next maintenance PR. I'm afraid that if we do it here, we would need to do multiple changes to all kinds of tests, if the underlying logic of code coverage has been changed.

contracts/bonding/BondingManager.sol
SidestreamSweatyPumpkin: We didn't want to rename an error to not introduce a breaking change. But if you think it's safe, we can update it to say "transcoder has already called reward for the current round".

rickstaa:
**Non-blocking test suggestion:** the auth require is never tested with a wrong caller while one is set (mapping = A, sender = B). It's somewhat implicit in the existing unset-caller test given the current implementation, but a specific test would guard future changes — it pins the check to "only the authorized address" vs "anyone once set":

```js
it("should fail if caller is not the reward caller set by the transcoder", async () => {
    await bondingManager.connect(transcoder).setRewardCaller(nonTranscoder.address)
    await expect(
        bondingManager.connect(signers[2]).rewardForTranscoder(transcoder.address)
    ).to.be.revertedWith("caller must be a reward caller set by the transcoder")
})
```

test/unit/BondingManager.js
```diff
@@ -6158,6 +6158,130 @@ describe("BondingManager", () => {
...
+                    .setRewardCaller(ZERO_ADDRESS)
+                await expect(unsetRewardCallerTx)
+                    .to.emit(bondingManager, "RewardCallerSet")
+                    .withArgs(transcoder.address, ZERO_ADDRESS)
+
+                const rewardTx2 = bondingManager
+                    .connect(nonTranscoder)
+                    .rewardForTranscoder(transcoder.address)
+                await expect(rewardTx2).to.be.revertedWith(
+                    "caller must be a reward caller set by the transcoder"
```
rickstaa, line 6227: The auth require is never tested with a wrong caller while one is set (mapping = A, sender = B). It's somewhat implicit in the existing unset-caller test given the current implementation, but a specific test would guard future changes. However not a big blocker from my side.

rickstaa (approved):
@SidestreamSweatyPumpkin Implementation is correct, my personal security and code review found no exploits and no griefing angles (full notes [in Notion](https://app.notion.com/p/LIP-118-implementation-Rick-audit-3a0660222d08817d9617d9777c4f44d0)). Summary: setter is self-scoped and revocable; delegated reward is economically identical to a self-call; no new external call sites; storage layout append-safe; all BondingManager unit tests pass. 

Biggest finding was that we need to patch the explorer HistoryView in the explorer with the upgrade to ensure reward events still show up correctly for people using this new function. The change is not breaking the subgraph or go-livepeer. For go-livepeer as discussed before the changes are only addative.

Left some minor comments regarding wording and tests.

test/unit/BondingManager.js
SidestreamSweatyPumpkin: Thanks! We've run the suggested test and confirmed it passes locally. However, since the test coverage is already at 100.00%, and revert case is [already implicitly tested here](https://github.com/sidestream-tech/livepeer-protocol/blob/99c00089b760d6980cb2e44fee3a2aad715657b0/test/unit/BondingManager.js#L6223-L6228), we wouldn’t modify this PR further. However, we can add this test case in the maintenance PR when we update the test coverage package, as [proposed here](https://github.com/livepeer/protocol/pull/648#discussion_r3639711691).

package.json
SidestreamCrunchyCarrot: Resolving the comment since [Rick approved the PR](https://github.com/livepeer/protocol/pull/648#pullrequestreview-4766619082) (and as it's required by the repository rules before merge)

contracts/bonding/BondingManager.sol
SidestreamCrunchyCarrot: Resolving the comment since [Rick approved the PR](https://github.com/livepeer/protocol/pull/648#pullrequestreview-4766619082) (and as it's required by the repository rules before merge)

test/unit/BondingManager.js
SidestreamCrunchyCarrot: Resolving the comment since [Rick approved the PR](https://github.com/livepeer/protocol/pull/648#pullrequestreview-4766619082) (and as it's required by the repository rules before merge)

SidestreamCrunchyCarrot (approved):
Sanity checks:
- ✅ No changes unrelated to the PR description
- ✅ CI pass at the latest commit 99c0008
- ✅ Hardhat-based tests run locally on the latest block (`939 passing (6m)`)
- ✅ Test coverage is at 100%

contracts/bonding/BondingManager.sol
```diff
@@ -863,21 +886,53 @@ contract BondingManager is ManagerProxyTarget, IBondingManager {
      * @param _newPosPrev Address of previous transcoder in pool if the caller is in the pool
      * @param _newPosNext Address of next transcoder in pool if the caller is in the pool
      */
-    function rewardWithHint(address _newPosPrev, address _newPosNext)
-        public
-        whenSystemNotPaused
-        currentRoundInitialized
-        autoCheckpoint(msg.sender)
-    {
+    function rewardWithHint(address _newPosPrev, address _newPosNext) public {
```
dob, line 889: Why did we do away with the whenSystemNotPaused, currentRoundInitialized, autoCheckpoint modifiers on this function?

dob (commented):
I'm comfortable with the logic introduced in this change. Though I did leave one open question about the modifiers on the rewardWithHint function to address.

contracts/bonding/BondingManager.sol
SidestreamCrunchyCarrot: They didn't go away. All of those modifiers are being called in the underlying private function `_rewardWithHint` on line 926:

https://github.com/livepeer/protocol/blob/99c00089b760d6980cb2e44fee3a2aad715657b0/contracts/bonding/BondingManager.sol#L922-L926

This way, both `rewardWithHint` and `rewardForTranscoderWithHint` has the same modifiers.

rickstaa: To add to this, keeping it in the parent function may increase the risk of introducing side effects if `rewardWithHint` or any of the other external functions change.

PR #652, opened 2026-01-12
feat: extend auxData to 96 bytes

rickstaa:

Allow a 64–96 byte `auxData` field to support additional on-chain metadata. This was part of the protocol work to distinguish different job types on-chain (see this tech spec: [https://www.notion.so/On-Chain-Ticket-Job-Type-Distinction-2bf660222d08814d89b1fa7221ad3291](https://www.notion.so/On-Chain-Ticket-Job-Type-Distinction-2bf660222d08814d89b1fa7221ad3291)).

This proposal was originally drafted as a protocol change that would not materially increase gas costs or code complexity. However, given the new requirement to also [encode payment clearinghouse information on-chain](https://www.notion.so/livepeer/Payments-Clearinghouse-25e0a3485687817394d0c3442ffca622), please review this approach with @mehrdadmms, @yondonfu, and @j0sh to confirm whether this is still the preferred solution.

- Extended auxData field to accept 96 bytes.
- Updated tests.

Not fully tested, however made sure tests pass.

**Checklist:**

### DIFF
--- a/contracts/pm/mixins/MixinTicketProcessor.sol
+++ b/contracts/pm/mixins/MixinTicketProcessor.sol
@@ -74,7 +74,10 @@ abstract contract MixinTicketProcessor is MixinContractRegistry, MTicketProcesso
         pure
         returns (uint256 creationRound, bytes32 creationRoundBlockHash)
     {
-        require(_auxData.length == 64, "invalid length for ticket auxData: must be 64 bytes");
+        require(
+            _auxData.length >= 64 && _auxData.length <= 96,
+            "invalid length for ticket auxData: must be between 64 and 96 bytes"
+        );
 
         // _auxData format:
         // Bytes [0:31] = creationRound
--- a/test/unit/TicketBroker.js
+++ b/test/unit/TicketBroker.js
@@ -798,22 +798,79 @@ describe("TicketBroker", () => {
             ).to.be.revertedWith("ticket sender is null address")
         })
 
-        it("reverts if ticket auxData != 64 bytes", async () => {
-            const auxData = ethers.constants.HashZero
-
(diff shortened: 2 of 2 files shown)


j0sh:
Thanks for checking in on this @rickstaa 

Question: does the auxData need to a strict upper bound check? Is there an issue with allowing unbounded auxData through? That would allow us to add more data to tickets as needed, without on-chain changes.

The on-chain clearinghouse data is still TBD but I imagine it'd be somewhere in the neighborhood of 8-32 additional bytes, depending on the final encoding. We'd probably want to encode one or two fields, eg a clearinghouse "customer ID" and maybe an "user ID" of the customer's own customer.

We will probably need some way of signaling which types of metadata are encoded in the auxData, because ordinary gateways don't need to publish clearinghouse data that isn't relevant to them.

One simple way of signaling is to use a bit string indicating which metadata types are present. This can be a variable-length encoding, eg using the high bit as a continuation flag. So the overhead for that metadata signaling would be just 1 byte, to begin with. If we ever have more than 127 types of metadata, then we can set the high bit and add another byte.

The interpretation of that bit string (and the metadata bytes associated with each bit) can happen off-chain; I don't think this is anything the on-chain protocol needs to be aware of.

The actual metadata would naturally need to be encoded the same order they are signaled, and there would need to be consensus among clients (eg, a registry) to avoid new fields from stomping over one another.

So for example:


On-chain job type: `0b1`
Clearinghouse customer ID: `0b10`
Clearinghouse user ID: `0b100`

Encoding all these would be:
```
<Current 64-byte auxData>
0b00000111 ( 0x07 )
<32-bit job metadata>
<8-byte clearinghouse customer ID>
<8-byte clearinghouse user ID>
 ```

